PR #49, opened 2017-06-07
Adding basic theme command

0x726564646974:
Basic theme code.

(Note: pushed dev branch is missing ActiveScriptsUI.js so there may be some changes needed in there as well)


danielyxie:
Thanks! And sorry about my shitty css. I'm new to css as you could probably tell ;)

PR #62, opened 2017-06-16
Implement top

LTCNugget:
Implementation of the "top" command.

### DIFF
--- a/src/Terminal.js
+++ b/src/Terminal.js
@@ -954,37 +954,35 @@ var Terminal = {
                 }
 				break;
 			case "top":
-                if(commandArray.length != 1) {
-                  post("Incorrect usage of top command. Usage: top"); return;
-                }
+				if(commandArray.length != 1) {
+				  post("Incorrect usage of top command. Usage: top"); return;
+				}
 
-                post("Script                    RAM Usage                 Threads");
+				post("Script                    Threads         RAM Usage");
 
-                var currRunningScripts = Player.getCurrentServer().runningScripts;
-                //Iterate through active scripts on current server
-                for(var i = 0; i < currRunningScripts.length; i++) {
-                    //Iterate through scripts on current server
-                    for(var j = 0; j < Player.getCurrentServer().scripts.length; j++) {
-                        if(currRunningScripts[i] === Player.getCurrentServer().scripts[j].filename) { //If the script is active
-                            var script = Player.getCurrentServer().scripts[j];
+				var currRunningScripts = Player.getCurrentServer().runningScripts;
+				var currScripts = Player.getCurrentServer().scripts;
+				//Iterate through scripts on current server
+				for(var i = 0; i < currScripts.length; i++) {
+				    if(currRunningScripts.includes(currScripts[i].filename)) { //If the script is running
+					var script = currScripts[i];
 
-                            //Calculate name padding
-                            var numSpacesScript = 26 - script.filename.length; // 26 is the width of each column
-                            var spacesScript = Array(numSpacesScript+1).join(" ");
+					//Calculate name padding
+					var numSpacesScript = 26 - script.filename.length; //26 -> width of name column
+					var spacesScript = Array(numSpacesScript+1).join(" ");
 
-                            //Calculate and transform RAM usage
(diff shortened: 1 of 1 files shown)


src/Terminal.js
```diff
@@ -954,8 +954,38 @@ var Terminal = {
...
+                if(commandArray.length != 1) {
+                  post("Incorrect usage of top command. Usage: top"); return;
+                }
+                
+                post("Script                    RAM Usage                 Threads");
+                var currRunningScripts = Player.getCurrentServer().runningScripts;
+                for(i = 0; i < currRunningScripts.length; i++) {
+                    //Check if script is on current server
+                    for(j = 0; j < Player.getCurrentServer().scripts.length; j++) {
+                        if(currRunningScripts[i] === Player.getCurrentServer().scripts[j].filename) {
```
HDEHacks: Won't this check if the script is present on the current server, not whether it's running from the current server?

src/Terminal.js
```diff
@@ -954,8 +954,38 @@ var Terminal = {
...
+                            var script = Player.getCurrentServer().scripts[j];
+
+                            //Add script name
+                            var entry = script.filename;
```
HDEHacks: Instead of iteratively concatenating substrings, it is far more performant in JS to build up an array of substrings, and then to `.join()` them.

src/Terminal.js
```diff
@@ -954,8 +954,38 @@ var Terminal = {
...
+
+                            //Calculate padding and add RAM usage
+                            var numSpaces = 26 - entry.length;
```
CoryMcDonald: what is this magic number

src/Terminal.js
```diff
@@ -954,8 +954,38 @@ var Terminal = {
...
+                            var spaces = Array(numSpaces+1).join(" ");
+                            entry += spaces;
+                            var ramUsage = (script.ramUsage * script.threads * Math.pow(1.02, script.threads - 1)).toFixed(3);
```
HDEHacks: The `.toFixed(3)` doesn't appear anywhere else when printing ram (or any number) - I wouldn't do this here unless it's everywhere.

src/Terminal.js
LTCNugget: This is spacing for the columns. Each column is 26 characters wide, including text.

src/Terminal.js
LTCNugget: The first loop iterates through the current running scripts on the server, while the second iterates through the scripts property of the server, which has the ram usage for each script (That was the only way I could find to determine it). The if statement only passes if the script is running on the current server.

src/Terminal.js
LTCNugget: It is just there to round the RAM usage number to 3 decimal places. I wasn't sure if leaving it out would break the formatting because of how long the RAM usage number is.

src/Terminal.js
LTCNugget: Sorry about that, my JavaScript is a little rusty. I mostly just copied from `executeScanCommand` and changed values.

src/Terminal.js
HDEHacks: If `Player.getCurrentServer().scripts` show you the scripts running on the current machine, why do we have the outer loop? Cant you get both `ram` and `filename` from the inner loop for each script?

src/Terminal.js
HDEHacks: If I understand correctly, the 26 character width is just something you decided right? Why not put RAM last (and # threads second) and that way it can just be as long as it wants to be?

It's just weird to round it here when it's not rounded anywhere else.

src/Terminal.js
HDEHacks: Perhaps for each in `Player.getCurrentServer().scripts` check if it's in `currRunningScripts`?
e.g.
`currRunningScripts.includes(script)`

LTCNugget: `Player.getCurrentServer().scripts` is an array of Script objects that the server _has_ on it, not that it is running.

src/Terminal.js
LTCNugget: Agreed. Updated.

PR #94, opened 2017-07-25
Account for negative money

desci:
After taking a course that costs more than the total income, the money balance went negative. That forced the `character-overview-container` **div** to use scrolling bars.

Proposed solution is to use *19%* instead of *18%* for *width*.

lightscreen 1500961810


danielyxie:
I don't think this will truly address the issue because once you get even more money (billions/trillions, etc.) the scrollbar will still come up

Eventually I will add shorthand notation for numbers, like $13.54B, etc.

PR #105, opened 2017-08-01
Load Decimal.js values

danielyxie:


desci:
@danielyxie I'm not sure how you are trying to address this, but look at the functions [`decimalSeparator`](https://github.com/sparticle999/SpaceCompany/blob/28b869840681ceb0aeffe858e864bd27d6e7e4eb/utils.js#L36) and [`formatEveryThirdPower`](https://github.com/sparticle999/SpaceCompany/blob/28b869840681ceb0aeffe858e864bd27d6e7e4eb/utils.js#L42) in this file: https://github.com/sparticle999/SpaceCompany/blob/gh-pages/utils.js

That game stick to 6 characters for any amount, for example:

number | representation
--- | ---
one | 01.000
two hundred and thirty | 230.00
five thousand and four hundred and fifty five | 5,455k
twelve million | 12,00M
three hundred billion | 300,0B

And so on. In the example above I used commas for decimal separators.

danielyxie:
Yup, this was one of the things I was going to do next. I was recommended some other libraries but I'll take a look at those functions too, thanks!

PR #170, opened 2018-01-02
fileExists netscript function now works with txt files

devoidfury:
I was looking through the wiki and noticed that you can `read`/`write` to `*.txt` files, but `fileExists` always returned false on those. This should fix it, but please also test to make sure!


danielyxie:
Thanks!

PR #190, opened 2018-02-10
Fix some stuff with ls() and cat

wraithan:
Wanted to be able to clean up after a few scripts, but found I couldn't find the files programmatically and peeked in to see the `ls` netscript function was missing the chunk looking through `server.textFiles`

Also fixed some missing documentation. Will keep adding small stuff to this branch as I complete them.

Changed .txt files to be rendered using `<pre>` instead of `<p>` and removed the newline conversion to `<br>` since it wouldn't be needed. This way `\t` is retained and data which shouldn't be wrapped since newlines can be significant, doesn't get wrapped.

### DIFF
--- a/doc/source/netscriptfunctions.rst
+++ b/doc/source/netscriptfunctions.rst
@@ -326,9 +326,10 @@ scp
 ls
 ^^
 
-.. js:function:: ls(hostname/ip)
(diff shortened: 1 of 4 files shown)


danielyxie:
Thanks!

Issue #189, opened 2018-02-09
Missing files from repo


kopelli:
Trying to make sense of the source files in this repository and the bundled output, and it seems like you're missing some files from the repository. That may just be because you're hosting the game directly from the repository, but then why have the `src` and `util` directories still in it? Or from the lack of a license on the repo, you're trying to keep this closed source...but still taking pull requests from the community.

At a minimum a minimum, I would think `webpack.config` and `package.json` should be added. Honestly not sure how you've managed to keep working on the project this long without checking those files in.

kopelli:
Thanks @danielyxie for adding the license, webpack.config & package.json. Although I think that raises a lot more questions for me now than it answered. 

Originally I brought it up because I wanted to see if I could detangle the code and try it in a front-end framework other than jQuery (purely for a "see if I can do it/try out a specific UI framework" pet project). Which would require some linting, tests, and breaking down the inter-dependencies between files. But it looks like you have much of that covered based on the packages & scripts called out in those files, just not in the repo.

And given the license is essentially the Doom Source License rather than a SPDX/OSI recognized one I guess my pet project will remain purely my own 😢

danielyxie:
It looks like the package.json I uploaded doesn't actually work. I just tried to setup a new dev environment and not all dependencies were installed. I'll try to fix this when I get some time

The license I added was recommended by someone else. I'm open to changing it if you have any suggestions/preferences. The only requirement I have is that its noncommercial. Essentially something along the lines of "you can do whatever you want with the code, as long as you dont make any money off of it or from derivative works".

kopelli:
Although Creative Commons typically isn't as a source code license, it seems like you're looking for CC-BY-NC-SA-4.0 ([SPDX License](https://spdx.org/licenses/CC-BY-NC-SA-4.0.html) or [Human readable version](https://creativecommons.org/licenses/by-nc-sa/4.0/). It's more permissive and denies commercial use directly (and from derivatives). There are other lesser known licenses that may cover it, but most open source licenses cover attribution and intellectual property claims more than commercial use.

But I'm no lawyer -- and ultimately it is ***your*** code & project -- so it's your call.

danielyxie:
I was actually looking at some Creative Commons licenses at first but didn't use them because they explicitly recommend against using it for software in their FAQ. 

Which specific permissions are you looking for? It sounds like distribution is one, but is there anything else?

RabidGuy:
[...]

PR #204, opened 2018-03-21
Ability to run webpack with no errors

kopelli:
This is related to #189

Updated the values of package.json to actually reflect the repository, rather than webpack. Left in the Webpack dependencies & scripts on the off chance to utilize them in future improvements to the code.

Tried to get Webpack to output exactly the same `bundle.js` content, but it seems like Webpack 4 changed variable names and is seeing the dependency order differently. It *seems* to be effectively the same output, but after the PlayerObject code the differences get harder to line up and visually compare. Unsure how to fully test the changes.

---
This change is [Reviewable](https://reviewable.io/reviews/danielyxie/bitburner/204)


kopelli:
@danielyxie Let me know what you think about these changes. I tried to get the bundle.js output to match, but not sure the version you were building on. Hopefully this will at least get a consistent build for others.

danielyxie:
Just pushed an update to the game, so I'll take a look at this when I start working on it again

Thanks!

danielyxie:
Haven't been working on the game too much so not sure. I tested it initially though and it seemed to work great!

PR #210, opened 2018-05-05
Remove package-lock.json from the repo

jaguilar:
This file is meant to be auto-generated by each user. Its presence in the repo creates a merge minefield for collaborators.

May want to merge this before my other request.


kopelli:
You realize that package-lock.json is *supposed* to be checked in and shared. It even states why on the [documentation for it](https://docs.npmjs.com/files/package-lock.json):
> This file is intended to be committed into source repositories, and serves various purposes:
> - Describe a single representation of a dependency tree such that teammates, deployments, and continuous integration are guaranteed to install exactly the same dependencies.
> - Provide a facility for users to "time-travel" to previous states of node_modules without having to commit the directory itself.
> - To facilitate greater visibility of tree changes through readable source control diffs.
> - And optimize the installation process by allowing npm to skip repeated metadata resolutions for previously-installed packages.

jaguilar:
No, I did not realize that.

danielyxie:
Re-added in dev

PR #212, opened 2018-05-05
Incorporate Netscript JS into Bitburner

jaguilar:
This pull request depends on #209. I've split them into two separate requests, because many people prefer smaller pull requests. If you don't mind big pull requests, you can just review this.

This includes all the code needed to run user Javascript code in Bitburner. The problems we are trying to solve are:

- Netscript is slow.
- Netscript lacks many JS features.
- Netscript presents an ongoing maintenance burden, to the extent you want to bring it into parity with JS.

# Design Overview

This evaluation engine works by using dynamic import and blob URLs to simulate serving user scripts from a fake server. These features are supported on all major browsers except IE.

I'll create another pull request to add some documentation for the JS version of Netscript. The TL;DR is: 

- Write "await" before hack, grow, weaken, exec, and a few other APIs.
- Write "async function" before any function that contains await.
- Always await any function marked as async.
- Instead of using the global scope to execute your code, use `export async function main()`

Among some of the often requested Netscript features this enables implicitly:

- Object syntax
- Multi-level imports
- Spread syntax
- const and let
- True closure semantics

This also includes a more accurate way of determining script memory requirements.

# Testing

The demo is at

https://jaguilar.github.io/bitburner

You can import [this save file](https://github.com/danielyxie/bitburner/files/1976993/testSave.zip) to get you started with some JS scripts.

In terms of testing, I have been playing on this branch through BN1 and have not run in to much trouble. In terms of regressions:

- Syntax errors from dynamic import() don't come with line numbers. Not sure why. This makes typos harder to diagnose.
- It is possible to use setTimeout() to make code "escape" a script's termination. Such code cannot do much, since most Netscript functions check whether the worker script has terminated, but it is worth noting. They can still call any Netscript function that doesn't check termination.
- If the user writes an infinite loop, it will crash the tab.


jaguilar:
If you're curious about the types of things you can write in NetscriptJS, here's one of my non-trivial programs:

    const notificationFile = "hack-dist_notification.txt";

    // Other scripts can import this and use it to signal us to
    // reschedule our initial weaken/flexihack work.
    export function signal(s) {
        print("notifying hack-dist.script of a rescheduling condition");
        write(notificationFile, s, "w");
        scp(notificationFile, "home");
    }

    function consumeSignal() {
        if (fileExists(notificationFile)) {
            print("rescheduling reason: " + read(notificationFile));
            tprint("rescheduling reason: " + read(notificationFile));
            rm(notificationFile);
            return true;
        }
        return false;
    }

    async function awaitSignal() {
        while (!consumeSignal()) await sleep(10000);
    }

    function killHacks(h) {
        scriptKill("flexihack.js", h);
        scriptKill("weaken.js", h);
        scriptKill("watch-security.js", h);
        scriptKill("hack.js", h);
        scriptKill("watch-drain.js", h);
        if (h != "home") {
            scp(ls("home", "js"), "home", h);
        }
    }

    const minHomeRamAvailable = 100;  // Ballpark adjustment.

    function getScheduleableRam(h) {
        print("getSchedulableRam: " + h);
        let r = getServerRam(h);  // Should be nothing running on these servers (except home, where we reserve).
        if (h == "home") r[1] += minHomeRamAvailable;
        return Math.max(r[0] - r[1], 0);
    }

    export function getHosts(targets) {
        const all = read("spider_data.txt").split("\n").concat(getPurchasedServers()).concat(["home"]).map(h => ({
            address: h,
            ram: getScheduleableRam(h)
        })).sort((a, b) => b.ram - a.ram);
        let totalRam = 0;
        // Drop small servers. They're just noise.
        return all.filter(h => {
            if (h.ram < totalRam * 0.001) return false;
            totalRam += h.ram;
            return true;
        });
    }

    // Tag is an unused argument to differentiate different instances of the same script.
    export async function schedule(hosts, job) {
        print("scheduling: " + JSON.stringify(job));
        
        while (job.threads > 0 && hosts.length > 0) {
            print("on host: " + JSON.stringify(hosts[0]));
            const h = hosts[0];
            
            // Script should have been copied out to all candidate hosts already.
            const ramPerTask = getScriptRam(job.script, "home");
            
            let threads = Math.min(Math.floor(h.ram / ramPerTask), job.threads);
            if (threads > 0) {
                await scp(job.script, "home", h.address);
                await exec(job.script, h.address, threads, ...job.args);
            }
            job.threads -= threads;
            
            if (job.threads > 0) {
                // We couldn't schedule all our threads. We're done with this host.
                hosts.shift();  
            } else {
                // We scheduled everything. We're done with this job. Update available ram.
                h.ram -= threads * ramPerTask;
            }
        }
    }

    function desirability(address) {
        return getServerMaxMoney(address) * getServerGrowth(address) / getGrowTime(address);
    }

    // Starts hacking a target. Calls isSatisfied() when we have
    // as much RAM as we think we need to hack target forever.
    async function hackProcess(target) {
        if (getServerMaxMoney(target) < 1e6) {
            return;
        }
    
        // Put together a job list. These consist of as many hacks, grows and
        // weakens as we think we need to stably run the server.
        // (In the future, we might implement some sort of RAM reservation
        // system so that we can use a dynamic set of jobs rather than a static
        // one.)
        const jobs = [];
        
        tprint("jobs");
        
        // The first job is to weaken the server down to the minimum security.
        const nWeakens = Math.floor((getServerSecurityLevel(target) - getServerMinSecurityLevel(target)) / 0.05);
        const initialWeakenArgs = [target, Math.round(Math.random() * 10033000)];
        jobs.push({
            script: "weaken.js",
            threads: nWeakens,
            args: initialWeakenArgs,
        });
        
        const numBatch = 10;
        const weakPerBatch = 6;
        const hackPerBatch = weakPerBatch * 6;
        for (let j = 0; j < numBatch; ++j) {
            // j serves as an arbitrary tag to distinguish the batches.
            jobs.push({script: "weaken.js", threads: weakPerBatch, args: [target, j]});
            jobs.push({script: "flexihack.js", threads: hackPerBatch, args: [target, j]});
        }
        
        tprint("done jobs");
        
        async function waitUntilLowSec() {
            while (getServerSecurityLevel(target) > 3 + getServerMinSecurityLevel(target)) {
                await sleep(60000);
            }
        }
        
        tprint("schedulding jobs for " + target);
        
        while (jobs[jobs.length - 1].threads > 0) {
            const hosts = getHosts();
            for (let i = 0; i < jobs.length; ++i) {
                await schedule(hosts, jobs[i]);
            }
            if (jobs[jobs.length - 1].threads === 0) break;
            await sleep(20000);
        }
        
        tprint("done scheduling jobs for " + target);
        
        // We run this in the background because it is not part of our promise.
        // Nothing is "waiting" for this, but it still needs to be done.
        setTimeout(async function() {
            tprint("waiting for low sec signal " + target);
            await waitUntilLowSec();
            tprint("got low sec signal " + target);
            // Kill the initial weaken job before exitting. After that's done,
            // there's no more for this thread of control to work on. The jobs
            // will stay there unmolested until hack-dist restarts.
            getHosts().forEach(h => {
                kill("weaken.js", h.address, initialWeakenArgs);
            });    
        })
    }

    // Changes each promises to return, on success, an array of
    // [index in array, value].
    async function raceIndex(...promises) {
        return Promise.race(promises.map((p, i) => {
            return p.then(value => { return [i, value]; });
        }));
    }

    async function scheduleTargets(targets) {
        const signalPromise = awaitSignal();
        while (targets.length > 0) {
            tprint("inwhile");
        
            // When this resolves, it's scheduled as much ram as it needs, and
            // we can begin allocating more.
            const [i] = await raceIndex(hackProcess(targets.shift()), signalPromise);
            tprint("started hack");
            if (i === 1) return;  // Run a new scheduling loop.
            
            // Otherwise, we just finished working on this one job. We can start another.
        }
        
        // If we get to this point, we're hacking all targets. Now we only need to
        // wait for a signal.
        await signalPromise;
    }

    export async function main() {
        disableLog("scp");
        disableLog("sleep");
        
        consumeSignal();
        
        while (true) {
            consumeSignal();
            
            // The spider is responsible for finding targets.
            let targets = await read("spider_data.txt").split("\n");
            
            // Hack the most desirable servers. The ones that aren't in the top 15
            // are probably just noise.
            targets = targets.sort((a, b) => desirability(b) - desirability(a)).slice(0, 15);
            tprint("targets=" + targets);
            
            // Cancel all running weakens and flexihacks on hosts, and record
            // their available RAM.
            getHosts().forEach(h => { 
                scp(ls("home", ".js"), "home", h.address);
                killHacks(h.address); 
            });
            
            await sleep(6000);

            await scheduleTargets(targets);
        }
    }

jaguilar:
To fix the issue with infinite loops, we could probably add a query parameter like ?dontStartScripts, which would skip the loading of any scripts from the save when the page loads. In the long term, if users scripts ran in worker scripts, then this would not be an issue, but that would require a more significant effort to rearchitect the game.

danielyxie:
I merged this manually but Github was still reporting conflicts that I didn't want to deal with so I'm just closing this.

PR #218, opened 2018-05-06
no longer charge for while and for separatly

hydroflame:
every for can be changed into a while, every while can be changed into a for. Simpler to consider them the same.

tbh just close if you don't want that. I'm mostly doing it to understand how to dev against this project.


hydroflame:
I really don't mind, just waiting for a word from the author and I'll swap over.

PR #227, opened 2018-05-06
serverinit: numOpenPorts can be zero

jaguilar:
This caused a bug where anything with a falsy value got 5 set in this field.


hydroflame:
I think you have 2 PR in this branch (higher level runningFN and numportfix?)

jaguilar:
There's one in my dev branch, and one in this feature branch. I am a github baby, so I didn't realize how I was supposed to make PRs. So if this one gets merged after the other one, then I won't make this mistake again.

jaguilar:
Up to you whether you still want to accept this. I think !== undefined is better, but my JS understanding is limited, so it could be wrong.

src/Server.js
```diff
@@ -59,7 +59,7 @@ function Server(params={ip:createRandomIp(), hostname:""}) {
     this.serversOnNetwork        = [];
 
     //Port information, required for porthacking servers to get admin rights
-    this.numOpenPortsRequired = params.numOpenPortsRequired != null ? params.numOpenPortsRequired : 5;
+    this.numOpenPortsRequired = params.numOpenPortsRequired !== undefined ? params.numOpenPortsRequired : 5;
```
kopelli: From an idiomatic JavaScript perspective, I'd expect to read this as `this.numOpenPortsRequired = params.numOpenPortsRequired || 5;`. But having a private refactoring of a lot of this code, I know that there's too much that is explicitly looking for explicit undefined or null values rather than falsey values.

jaguilar: You can't do that, because numOpenPortsRequired can be zero.

Probably the easiest thing is to just have zero be default.

jaguilar:
Discussed offline, and probably this isnt' the right way to go.

PR #229, opened 2018-05-06
added getServerCost function in NS and doc, returns the cost of a ser…

hydroflame:
…ver given the amount of ram


kopelli:
One thing I noticed when trying to tease apart the circular/spiderweb of dependencies is all the stand-alone functions that could be in a separate file that gets imported independently from everything else. In this case, the `purchaseServerCost()` function could be pulled out on it's own.

hydroflame:
own function or own file ?

kopelli:
Their own file. It would increase the number of `import` lines at the top of a file, but it creates a diamond dependency graph rather than a circular graph.

There's *a lot* more that has be broken apart to fix all the circular dependencies thanks to the god objects like Engine, Terminal, Player and Server that have dependencies across the entire code base.

PR #231, opened 2018-05-06
removed log

hydroflame:
this log happens when new a player that doens't have the neuroflux aug loads the game. It doesn't look like it's an error because it's a normal situation.


danielyxie:
It happens if the Augmentation object DNE as well. It is not an issue when the game loads/initializes, but could indicate an error at other times. I updated the  log message internally to be more descriptive.

PR #248, opened 2018-05-17
Nsjs memoryleak

danielyxie:


src/Faction.js
```diff
@@ -1031,6 +1031,14 @@ function purchaseAugmentation(aug, fac, sing=false) {
         var txt = "You must first purchase or install " + aug.prereqs.join(",") + " before you can " +
                   "purchase this one.";
         if (sing) {return txt;} else {dialogBoxCreate(txt);}
+    } else if (Player.money.lt(aug.baseCost * fac.augmentationPriceMult)) {
```
kopelli: It seems like there was other things that got merged into this besides just working on memory leaks...

danielyxie: Yeah, this was to address an exploit I found while testing the new NSJS changes. I fixed it without really thinking about this

Bad practice I know :anguished: My manager at work just chewed me out about this a few months ago too

kopelli: We're all guilty of it at some point. 😃

Doesn't help that a massive change like this ripples across so many files already, so it's easy for something to get overlooked.

PR #254, opened 2018-05-27
make the stock market accound and api button green when bought

hydroflame:
that way it's obvious if you bought the accounts or not.

Oh and I'm literally colorblind so if someone wants to propose a color that actually looks good I'm all for it.

screen shot 2018-05-27 at 1 33 25 am


hydroflame:
screen shot 2018-05-27 at 1 33 25 am

PR #255, opened 2018-05-27
fix faction join exploit

hydroflame:
You could join all factions in game as long as you get invited to them before joining them.

With this fix it's no longer possible to join all city factions as they get removed when you join a pending invite.

before join
screen shot 2018-05-27 at 2 07 13 am

after joining sector-12 (who pretty much every other city hates apparently)
screen shot 2018-05-27 at 2 10 53 am


hydroflame:
screen shot 2018-05-27 at 1 50 47 am

PR #259, opened 2018-05-29
streamline crimes code

hydroflame:
All this PR does is aggregate all crime information in one place. It's not dismantling circular dependencies yet but it's a first step towards making that simpler. Also makes it way easier to add new crimes.

I tested manual crimes and singularity crimes.


danielyxie:
Will take a look at this and #260 after the next update

PR #261, opened 2018-05-30
[refactor] Pull in 'decimal.js' as an npm dependency

kopelli:
This change is [Reviewable](https://reviewable.io/reviews/danielyxie/bitburner/261)


package.json
```diff
@@ -13,6 +13,7 @@
     "async": "^2.1.2",
     "bluebird": "^3.5.1",
     "brace": "^0.11.1",
+    "decimal.js": "7.2.3",
```
kopelli: Note that it's locked to the version that was sitting in the `utils` folder.

PR #272, opened 2018-06-05
Fix faction rep

hydroflame:
ooooopppsssss

everything is NaN if this fix isn't in lol


stavvie34:
So I’m really high, and this was the best possible comment I could ever read. You have no clue ho silly this code has made me felt. God bless. Lol.

PR #273, opened 2018-06-05
Shrunk the UI of Hacknet Nodes to view more at a time on screen.

kopelli:
The UI of a single Hacknet Node now only takes up ~50% of the screen.
This allows two nodes to be displayed per "row" when the screen is wide
enough.

Also repositioned the buttons for the nodes so they are inline with the
information each updates. This visual correlation lets us reduce the
text that needs to be in each button.

Reduced the amount of DOM that needs to be continuously garbage
collected by updating specific text rather than throwing out entire HTML
elements.

---
This change is [Reviewable](https://reviewable.io/reviews/danielyxie/bitburner/273)

### DIFF
--- a/css/menupages.css
+++ b/css/menupages.css
@@ -288,30 +288,39 @@
 }
 
 #hacknet-nodes-text,
-#hacknet-nodes-money,
 #hacknet-nodes-container li {
-    width: 70%;
     margin: 10px;
     padding: 10px;
 }
 
-#hacknet-nodes-purchase-button {
-    display: inline-block;
+#hacknet-nodes-container li {
+    float: left;
+    overflow: hidden;
+    white-space: nowrap;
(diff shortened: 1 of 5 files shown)


hydroflame:
can you put a pic as well ?

kopelli:
Thanks @hydroflame for calling out screenshots. Not sure how I missed the text being hidden behind the player's info panel before.

_Current State of the game_
bitburner-hacknet-streamline-pre

_After this change_
bitburner-hacknet-streamline-post

hydroflame:
I haven't actually tested your changes, I didn't know there was some hidden text, I just thought it'd be easier for danielyxie to review if there was a pic 👍 looks good though

danielyxie:
TIL about "vw"

PR #275, opened 2018-06-07
Simpler home ram

hydroflame:
screen shot 2018-06-07 at 2 55 28 pm

buying ram was cheaper when doing it through singularity because a factor was different, now it's all reused code. Also the cost was changed to only calculate the first 5 digits of precision. This means the prompt can display a cleaner number like `$10.275t`, like upgrading CPU, instead of `$10,275,135,752,145`.

So neither a nerf nor a boost as it's not cheaper via manual clicking but more expensive via singularity

### DIFF
--- a/src/Player.js
+++ b/src/Player.js
@@ -413,8 +413,7 @@ PlayerObject.prototype.getUpgradeHomeRamCost = function() {
     //Calculate cost
     //Have cost increase by some percentage each time RAM has been upgraded
     const mult = Math.pow(1.58, numUpgrades);
-    var cost = currentRam * CONSTANTS.BaseCostFor1GBOfRamHome;
-    cost = cost * mult;
+    var cost = currentRam * CONSTANTS.BaseCostFor1GBOfRamHome * mult;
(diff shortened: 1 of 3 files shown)


src/Player.js
```diff
@@ -405,6 +405,20 @@ PlayerObject.prototype.getHomeComputer = function() {
...
 
+PlayerObject.prototype.getUpgradeHomeRamCost = function() {
+    //Calculate how many times ram has been upgraded (doubled)
+    const currentRam = Player.getHomeComputer().maxRam;
+    const numUpgrades = Math.log2(currentRam);
+
+    //Calculate cost
+    //Have cost increase by some percentage each time RAM has been upgraded
+    const mult = Math.pow(1.58, numUpgrades);
+    var cost = currentRam * CONSTANTS.BaseCostFor1GBOfRamHome;
```
kopelli: why not combine this line with the next?

hydroflame: I can, doesn't change much though

PR #283, opened 2018-06-12
[chore] Removing and ignoring test bundle output

kopelli:
Since test bundles aren't necessary for running the game in general, and
the content of it should be regenerated every time, it doesn't need to
be included as part of merge requests.

---
This change is [Reviewable](https://reviewable.io/reviews/danielyxie/bitburner/283)


danielyxie:
I went ahead and manually made the change in dev to avoid dealing with the conflicts

PR #287, opened 2018-06-12
Dev mode

hydroflame:
This PR adds a developer/cheat menu to the game. This is more and more needed to help setup our save files to the right state in order to test bug fixes/new features.

To enable dev mode just set `DevModeEnabled = true` in `devmode.js`, this file is `.gitignore`d so chances are it will never be pushed to prod.

Criticism is welcome. My html-fu is pretty bad though :\

screen shot 2018-06-12 at 5 16 52 pm
screen shot 2018-06-12 at 5 16 58 pm

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ Changelog.txt
 Netburner.txt
 /node_modules
 /dist/*.map
+src/devmode.js
--- a/css/menupages.css
+++ b/css/menupages.css
@@ -486,6 +486,21 @@
     width: 50%;
 }
 
+/* Dev menu */
+#dev-menu-container {
+    position: fixed;
+    padding-top: 10px;
+}
+
+#dev-menu-text {
+    width: 70%;
+    margin: 10px;
+}
+
+#dev-menu-container a {
+    width: 50%;
+}
+
 /* Location */
 #location-container {
 	position: fixed;
--- a/css/menupages.css
+++ b/css/menupages.css
@@ -501,6 +501,18 @@
     width: 50%;
 }
 
+.dev-text-input {
+    color: var(--my-font-color);
+    border: 1px solid white;
+    background-color:black;
+}
+
+.dev-dropdown-input {
+    color: var(--my-font-color);
+    border: 1px solid white;
+    background-color:black;
+}
+
 /* Location */
 #location-container {
 	position: fixed;
--- a/index.html
+++ b/index.html
@@ -488,20 +488,20 @@
 
         <p id='dev-menu-text'>Augmentation related: </p>
         <!-- gets populated with the list of all augments -->
-        <select id="dev-menu-aug-dropdown"></select>
+        <select id="dev-menu-aug-dropdown" class="dev-dropdown-input"></select>
         <a id="dev-add-aug" class="a-link-button tooltip">Queue Augmentation<span class="tooltiptext">May require save + reload</span></a>
 
-        <input id="dev-sf-n" type="number" placeholder="SourceFile-N"><input id="dev-sf-lvl" type="number" placeholder="SourceFile-Lvl"><a id="dev-add-source-file" class="a-link-button tooltip"> Add/Remove source file <span class="tooltiptext">If Lvl == 0 the sf will be removed, calling it with another level will replace your current source file. You CAN set a source file higher than it's maximum level.</span></a>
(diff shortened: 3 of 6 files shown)


kopelli:
I've been contemplating something similar. But I've been looking at a way for BDD tests to set up a given state rather than manually testing these things.

I was thinking something more along the lines of manipulating things from dev tools rather than building it out in the game itself. I've done it occasionally in my local changes by doing something like `window.Player = PlayerObject;` so that way it's globally exposed. Then I can manipulate the Player object from the console.

hydroflame:
Yeah I think there should be tests AND this menu. Because sometimes you just wanna test a new feature and it's not immediately obvious what will happen and so writing tests isn't the first thing you'd want to do.

the window.Player trick works but this is just more convinient.

Also I talked to danielyxie privately and he was ok with that kind of menu, he just wanted to decide when/how this would be enabled. I'm thinking either get argument `?dev`, `options > enable dev`, or setting a global variable when building.

kopelli:
My thought is that this would be a build-time variable as it's something that wouldn't normally be available in the released version of the game. Ideally we get to a point where things are more modular and Webpack can spit out a separate bundle of the HTML, CSS, and JS to include for this "play tester" mode.

The `options` checkbox puts it a little too font-and-center and makes it feel like it's actually part of the game rather than helping while developing the game.

The query parameter is a possibility, but feels like a hidden easter egg waiting to be found.

hydroflame:
I mean, the player can cheat anyway, we're just talking about how much effort. like he could use cheat engine to set infinite money or something. Ok granted dark souls doesn't offer a "infinite hp" mode just because it's possible to cheat. But I guess my point is let's not go overkill on trying to prevent cheats, those who want will do it anyway.

Heck you can just export your save file and modify it at will, it's just base64 json.

I'm with you that the checkbox and the query parameter aren't as safe as the "build" time variable though.

kopelli:
Very true. It is an open source game after all...

And you *could* attempt to modify the JSON in the save file manually, but with how huge that object is with scripts (and the actual literature files, and other large text objects that probably could/should be handled differently) it's quite cumbersome to manipulate without having some developer tools handy. Especially since each object is a string-encapsulated JSON representation with the way the `Generic_to/fromJSON` objects work.

Let me know if you want help figuring out a way to make it work with WebPack.

hydroflame:
My point with the json was that player can easily modify that to cheat.

hydroflame:
I've made the enabling a constant in a file called `devmode.js` which is gitignored from now on. There's probably a better way but I'm not familiar with the js environment. suggestions welcomed.

kopelli:
Having it as a separate file with just the export definitely works. Having it as a build flag (and likely as part of an NPM script) would probably be better since it would better reflect WebPack's nature of a Production vs. Development build.

But I'm pretty sure because you committed the file already, it won't be ignored as it's being tracked by Git. You'll need to delete it from Git (`git rm src/devmode.js`) and add it as a placeholder file (e.g. `src/devmode.template.js`) that can be copied to `src/devmode.js` for everybody that clones the repo. I *think* we could tap into the "preinstall" NPM script to copy the file whenever somebody runs `npm install`, but I'm not 100% sure if those scripts trigger when you're installing the package itself, or when installing its dependencies.

danielyxie:
Is creating separate webpack configs and specifying an environment variable a good way to do this? Something similar to what's here:

https://webpack.js.org/guides/production/
https://webpack.js.org/guides/production/#specify-the-environment

kopelli:
The problem with the WebPack documentation is a lot of it is still written for WebPack 3 style of plugins...

For WebPack 4, it's based on the mode passed in. So right now we're [forcing it to Development](https://github.com/danielyxie/bitburner/blob/master/webpack.config.js#L5), but the alternative way would be relying on [`--mode` flag](https://github.com/webpack/webpack/issues/6460#issuecomment-386947990). I believe we're defining all the optimizations that the mode would typically toggle, but it may change the output of the bundles.

danielyxie:
@kopelli @hydroflame 
Could you take a look at PR #297? I tried to make it so that "Dev Mode" is enabled using a build flag. I don't know if what I did is considered good/standard practice

hydroflame:
You need to remove the `import ... devmode.js` in `engine.js` also can you make PRs on PRs. else the git tree becomes fucked

kopelli:
Looks pretty good. Had one comment about the unnecessary `JSON.stringfy()` calls, but otherwise seems good

hydroflame:
Was copied over to #297

PR #292, opened 2018-06-13
fixed 282

### LINKED ISSUE #282: Exporting real estate to another city causes NaNs
save file: https://gist.github.com/hydroflame/f172aa78c2dd81991505fbc5ea5ef3d8
export "HomeOwnership" real estate to another city (default works) and notice the numbers go to `NaN`

hydroflame:
fix #282

### DIFF
--- a/src/CompanyManagement.js
+++ b/src/CompanyManagement.js
@@ -231,7 +231,8 @@ let MaterialSizes = {
     Chemicals:  0.05,
     Drugs:      0.02,
     Robots:     0.5,
-    "AICores": 0.1
+    AICores:    0.1,
+    RealEstate: 0,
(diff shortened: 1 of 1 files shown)


src/CompanyManagement.js
```diff
@@ -231,7 +231,8 @@ let MaterialSizes = {
     Chemicals:  0.05,
     Drugs:      0.02,
     Robots:     0.5,
-    "AICores": 0.1
+    AICores:    0.1,
+    RealEstate: 0,
```
hydroflame: Adding real estate fixed the undefined at line 1318, which goes `min(inf, x)`, which returns just `x`

PR #293, opened 2018-06-13
user can no longer select button text (meaning they can spam)

hydroflame:


kopelli:
Admittedly you could spam before since it still typically recognized the click event, but at least now it won't select text.

PR #295, opened 2018-06-13
window and document symbold now occur impossibly high ram cost in nsjs

hydroflame:


kopelli:
Wouldn't this mean the example's included in the [documentation](https://bitburner.readthedocs.io/en/latest/netscriptjs.html) become essentially invalid?

hydroflame:
Yeah but bigd wanted to avoid that anyway, but you're right I need to update documentation

danielyxie:
I did not want to prevent users from manipulating DOM, I just wanted to make it very expensive in terms of RAM (I'll be changing the numbers)

PR #297, opened 2018-06-13
Dev mode

danielyxie:


webpack.config.js
```diff
@@ -1,18 +1,21 @@
...
-      new webpack.ProvidePlugin({
-        // Automtically detect jQuery and $ as free var in modules
-        // and inject the jquery library
-        // This is required by many jquery plugins
-        jquery: "jquery",
-        jQuery: "jquery",
-        $: "jquery"
-      }),
+        new webpack.DefinePlugin({
+            'process.env.NODE_ENV': argv.mode === 'development' ? JSON.stringify('development') : JSON.stringify('production')
```
kopelli: You don't need the extra JSON.stringify() calls. All it does is make sure it's double-quoted.

PR #302, opened 2018-06-14
[WIP] Stat modifier type

hydroflame:
This patch introduces a new type called `StatModifier` which holds all sortof modifier to the player, be it max money, starting programs, access to corp, access to bladeburner, access to WSE.

So far I have all the source files converted to StatModifiers, meaning we can remove the cyclic import between SourceFile and Player.

Notice for example that sf8Modifier will grant WSE account and API access instead of somewhere else in the code.


kopelli:
Overall looking good. Now to actually apply them in the `applySourceFile` function...

src/StatModifier.js
```diff
@@ -0,0 +1,56 @@
+function StatModifier(params) {
+    //Hacking multipliers
```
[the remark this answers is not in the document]
hydroflame: oh that's smart yeah

hydroflame:
@kopelli if you noticed the applySourceFile function is no longer being called. The Player asks the source file object to generate the StatModifier and then applies it within itself. `applySourceFile` is actaully deprecated now

src/Player.js
```diff
@@ -2022,7 +2029,61 @@ PlayerObject.prototype.reapplyAllSourceFiles = function() {
             console.log("ERROR: Invalid source file number: " + this.sourceFiles[i].n);
             continue;
         }
-        applySourceFile(this.sourceFiles[i]);
+        const modifier = sourceFileObject.modifierFunc(this.sourceFiles[i].lvl);
+        this.applyStatModifier(modifier);
+        sourceFileObject.owned = true;
```
hydroflame: @kopelli  here is what I'm talking about

src/Prestige.js
```diff
@@ -57,17 +57,6 @@ function prestigeAugmentation() {
     AddToAllServers(homeComp);
     prestigeHomeComputer(homeComp);
 
-    if (augmentationExists(AugmentationNames.Neurolink) &&
```
hydroflame: notice that now the StatModifier takes care of programs

src/Prestige.js
```diff
@@ -311,8 +300,8 @@ function prestigeSourceFile() {
     }
 
     //BitNode 8: Ghost of Wall Street
-    if (Player.bitNodeN === 8) {Player.money = new Decimal(BitNode8StartingMoney);}
-    if (Player.bitNodeN === 8 || hasWallStreetSF) {
```
hydroflame: This check is no longer required because sf8 Modifier gives WSE+TIX

PR #303, opened 2018-06-14
Add Typescript and TSLint to the tool chain

kopelli:
Based on the Discord suggestion, introducing TypeScript (and TSLint) to the code base so we can start getting some better type checking across the code base.

Take a look at my message on [9c09e44](https://github.com/danielyxie/bitburner/commit/9c09e44db2b842b4966aecad6c4fe750fc1012a2) about the approach I took with TSLint. I'm sure there will a number of rules that we'll want to disable/relax as we start building out more, but I've always preferred to take the on-by-default approach.

Definitely open to feedback (@danielyxie & @hydroflame ) before it gets merged in.

---
This change is [Reviewable](https://reviewable.io/reviews/danielyxie/bitburner/303)


hydroflame:
That's sexy

danielyxie:
Looks good. I'm busy for the next couple of days but after that I'll merge and start playing around with it

PR #317, opened 2018-06-20
Add "getPurchasedServerCost" NS function

kopelli:
Relatively simple function, and users could manually incorporate the logic behind it into their script, but it's nice to have a way to expose the cast to purchase a server before actually attempting to purchase it.

Not very attached to the actual name, so open to suggestions.

---
This change is [Reviewable](https://reviewable.io/reviews/danielyxie/bitburner/317)


danielyxie:
Will merge this in the near future...I'm planning on adding "BitNodeMultipliers" for the cost of purchasing servers and the # of servers you can purchase. This way I can make both changes at the same time

hydroflame:
I mean you might as well merge it now so people can at least start using it, it's not like you're gonna break the API later

PR #319, opened 2018-06-21
[WIP] Contributor file

kopelli:
Adding a CONTRIBUTING file to repository. This can be where we document the expectations of what needs to be done for a good PR. Especially as more tools & tests are added to the repository.

Don't feel this is complete enough just yet, so would appreciate some feedback before it gets merged in.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,47 @@
+#Contributing to BitBurner
+
+##In General
+The game is made better because the community as a whole speaks up about
+ways to improve the game. Here's some of the ways you can make your voice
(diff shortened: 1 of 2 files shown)


danielyxie:
There are some specifics that I'd like to add but I think everything you have here is fine

kopelli:
Definitely agree that it will need to be fleshed out further. Figured at least getting a baseline in here would help a bit.

Also was a little leery about putting in a persistent, open connection to the Discord server. It felt like an open solicitation to spammers and bots. Not really sure of a good way to advertise it off the top of my head.

PR #321, opened 2018-06-21
moved the suppress buy aug confirmation to settings, will now persist

hydroflame:
The current confirm aug purchase setting doesn't persist through augments/bns, I moved it with the other suppress also prevent the "You purchase the xyz augment it will not take effect until you install it" message

### DIFF
--- a/src/Settings.js
+++ b/src/Settings.js
@@ -2,19 +2,20 @@ import {Engine} from "./engine.js";
 
 /* Settings.js */
 let Settings = {
-    CodeInstructionRunTime:         50,
-    MaxLogCapacity:                 50,
-    MaxPortCapacity:                50,
-    SuppressMessages:               false,
-    SuppressFactionInvites:         false,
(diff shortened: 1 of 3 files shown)


src/Settings.js
```diff
@@ -2,19 +2,20 @@ import {Engine} from "./engine.js";
...
+    AutosaveInterval:                    60,
+    DisableHotkeys:                      false,
+    ThemeHighlightColor:                 "#ffffff",
+    ThemeFontColor:                      "#66ff33",
+    ThemeBackgroundColor:                "#000000",
+    EditorTheme:                         "Monokai",
+    EditorKeybinding:                    "ace",
 }
 
 function loadSettings(saveString) {
```
kopelli: We should be setting the default value here so that way returning players will get the new property.

Something like: 
`Settings.SuppressBuyAugmentationConfirmation = Settings.SuppressBuyAugmentationConfirmation || false` to default to false.

PR #323, opened 2018-06-22
cap difficulty and fix negative security growth

hydroflame:


src/Server.js
```diff
@@ -93,18 +93,23 @@ Server.prototype.getScript = function(scriptName) {
     return null;
 }
 
-//Strengthens a server's security level (difficulty) by the specified amount
-Server.prototype.fortify = function(amt) {
-    this.hackDifficulty += amt;
+Server.prototype.capDifficulty = function() {
```
kopelli: This chould really be generalized to a "confineTo" type of function that's a utility function rather than another prototype function. Or in addition to this function at least. Something like

```
function confineTo(value, min, max) {
    return Math.min(Math.max(min, value), max);
}
...
this.hackDifficulty = confineTo(this.hackDifficulty, Math.max(this.minDifficulty, 1), 1000000);
```

src/Server.js
```diff
@@ -817,7 +822,8 @@ function processSingleServerGrowth(server, numCycles) {
     // if there was any growth at all, increase security
     if(oldMoneyAvailable !== server.moneyAvailable) {
         //Growing increases server security twice as much as hacking
-        const usedCycles = numCycleForGrowth(server, server.moneyAvailable / oldMoneyAvailable);
+        let usedCycles = numCycleForGrowth(server, server.moneyAvailable / oldMoneyAvailable);
```
kopelli: *technically* can do these still as one line and keep the `const`...

src/Server.js
hydroflame: Usually the function is called `Clamp(x, min, max)` but meh I'm just gonna leave it like this for now

src/Server.js
hydroflame: Yeah but it's easier to read in 2 lines

PR #324, opened 2018-06-22
first conversion

hydroflame:
My first attempt at converting something to ts


kopelli:
Overall looks good. I'm going to guess that there will be some TSLint violations in there (at least for the use of `any`, but otherwise great. Hopefully this will start helping us shake out functions that aren't even used anywhere too.

PR #325, opened 2018-06-23
Add Eslint to linting toolchain

kopelli:
As it currently stands, this is simply getting the tool in place with the NPM script available to run it. Many rules are disabled as there are violations across the code base. I didn't want to introduce too big of a ripple effect with this one, so I didn't want to start reformatting files just yet.

We can work towards turning one on at a time and fixing things up.

---
This change is [Reviewable](https://reviewable.io/reviews/danielyxie/bitburner/325)


src/BitNode.js
```diff
@@ -308,14 +308,14 @@ function initBitNodeMultipliers() {
...
-            for (let i = 0; i < Player.sourceFiles.length; i++) {
+            var sf12Lvl = 0;
+            for (var i = 0; i < Player.sourceFiles.length; i++) {
                 if (Player.sourceFiles[i].n === 12) {
                     sf12Lvl = Player.sourceFiles[i].lvl;
                 }
             }
-            const inc = Math.pow(1.01, sf12Lvl);
-            const dec = Math.pow(0.99, sf12Lvl);
+            var inc = Math.pow(1.01, sf12Lvl);
```
hydroflame: Isn't `var` global ?

kopelli: case statements don't have a scope. So in this case there's no difference.

PR #339, opened 2018-06-26
[WIP] Convert HelperFunctions to ts

hydroflame:


utils/HelperFunctions.ts
```diff
@@ -22,54 +13,61 @@ function clearObject(obj) {
...
+function removeElementById(id: string) {
+    const elem: HTMLElement | null = document.getElementById(id);
+    if(elem === null) return;
+    removeElement(elem as HTMLElement);
 }
 
-function removeElement(elem) {
-    if (elem == null || !(elem instanceof Element)) {return;}
+// Same as removeElementById except if takes a DOM element object rather than an ID
+function removeElement(elem: Element) {
```
kopelli: why the use of "Element" vs. "HTMLElement"? Usage seems to be mixed. I'm sure there is a hierarchy between them, just not sure which is the more general and if we really need it to be.

utils/HelperFunctions.ts
```diff
@@ -80,17 +78,57 @@ function removeChildrenFromElement(el) {
...
+    tabIndex?: any;
+    clickListener?: any;
+    inputListener?: any;
+    changeListener?: any;
+    onkeyup?: any;
+    onfocus?: any;
+}
+
+function createElement(type: string, params: createElementParams): any {
+    var el: any = document.createElement(type);
```
kopelli: does `Element` or `HTMLElement` not work?

utils/HelperFunctions.ts
```diff
@@ -165,19 +203,23 @@ function createPopup(id, elems) {
...
 
-    for (var i = 0; i < elems.length; ++i) {
+    for (let i = 0; i < elems.length; ++i) {
         content.appendChild(elems[i]);
     }
     container.appendChild(content);
-    document.getElementById("entire-game-container").appendChild(container);
+    const game = document.getElementById("entire-game-container");
+    if(game === null) {
+        throw new Error('unable to find entire-game-container, massive problem');
```
kopelli: 👍 We don't make use of true errors enough.

danielyxie:
Done in #352

PR #342, opened 2018-06-26
remove all .js from all appplicable imports

hydroflame:


src/Terminal.js
```diff
@@ -1054,7 +1054,7 @@ let Terminal = {
                     if (fn === "*" || fn === "*.script") {
                         for (var i = 0; i < s.scripts.length; ++i) {
                             var file = new Blob([s.scripts[i].code], {type:"text/plain"});
-                            zip.file(s.scripts[i].filename + ".js", file);
```
danielyxie: The two non-import changes in here should not be removed

PR #349, opened 2018-07-05
Fix donate invalid input

hydroflame:
basicaly if you wrote `asdf\b\b\b\b` you get `invalid donation amount`, now it just says `This donation will result in 0.000 reputation gain`

### DIFF
--- a/src/Faction.js
+++ b/src/Faction.js
@@ -306,13 +306,16 @@ function displayFactionContent(factionName) {
     var donateDivWrapper = createElement("div", {class:"faction-work-div-wrapper"});
     donateDiv.appendChild(donateDivWrapper);
     var donateRepGain = createElement("p", {
-        innerText:"This donation will result in 0 reputation gain"
+        innerText:"This donation will result in 0.000 reputation gain"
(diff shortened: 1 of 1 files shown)


src/Faction.js
```diff
@@ -306,13 +306,16 @@ function displayFactionContent(factionName) {
...
+        innerText:"This donation will result in 0.000 reputation gain"
     });
     var donateAmountInput = createElement("input", {
         placeholder:"Donation amount",
         inputListener:()=>{
-            var amt = parseFloat(donateAmountInput.value);
-            if (isNaN(amt) || amt < 0) {
+            let amt = 0;
+            if(donateAmountInput.value !== "") {
+                amt = parseFloat(donateAmountInput.value);
```
kopelli: I wonder if we really want something more along the lines of the example function under the ["A stricter parse function" section from MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseFloat#A_stricter_parse_function) to allow for entering amounts like "1M" or "1B"

PR #363, opened 2018-07-20
added bladeburner functions related to action levels

hydroflame:


src/NetscriptBladeburner.js
```diff
@@ -0,0 +1,27 @@
...
+function unknownBladeburnerExceptionMessage(functionName, err) {
+	return `Bladeburner.${functionName}() failed with exception: ` + err;
+}
+
+function checkBladeburnerAccess(workerScript, functionName) {
+	const accessDenied = `${functionName}() failed because you do not` +
+		" currently have access to the Bladeburner API. This is either" +
+		" because you are not currently employed at the Bladeburner division" +
+		" or because you do not have Source-File 7";
+	console.log(Player.sourceFiles);
```
kopelli: Probably shouldn't include the console logging if we can help it. Or at the very least it should be `console.debug`

danielyxie: I imagine this was accidentally left in. I'll remove it

PR #393, opened 2018-07-27
Add CranialSignalProcessor prereqs

Kline-:


hydroflame:
The problem with this one is that for example BitRunners offer G5 without offering G1-4.

Kline-:
The G3 is the only one available from multiple factions as is. So it's really a question of if the intent is to be able to skip factions to the most powerful aug or for them to be obtained while progressing through the factions in order.

hydroflame:
Right, I just wanted to lay it all out so D can make an informed decision. If he wants to prevent skipping Cranials I'm fine with that.

PR #399, opened 2018-07-29
Clarify augmentation cost increase

Kline-:
I'm currently in this BN and the description reads as if all augs would be 3x more, not just BB ones.


hydroflame:
I don't think that's true

PR #409, opened 2018-07-31
Nuke unused

hydroflame:
This file wasn't used, nuke it!


kopelli:
Can also `npm uninstall uuid` since it's not used anywhere else. One less package to install...

PR #441, opened 2018-08-29
Updated Interactive Tutorial and added a few helpful links to the hac…

danielyxie:
…kers-starting-handbook Literature file

### DIFF
--- a/src/InteractiveTutorial.js
+++ b/src/InteractiveTutorial.js
@@ -1,54 +1,76 @@
 import {Engine}                         from "./engine";
 import {Player}                         from "./Player";
-import {dialogBoxCreate}                from "../utils/DialogBox";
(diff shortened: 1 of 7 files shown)


src/InteractiveTutorial.js
```diff
@@ -1,55 +1,77 @@
 import {Engine}                         from "./engine";
 import {Player}                         from "./Player";
-import {dialogBoxCreate}                from "../utils/DialogBox";
+import {Settings}                       from "./Settings";
+import {Terminal}                       from "./Terminal";
```
kopelli: oy...more circular dependencies....

danielyxie: I've removed this (in a local WIP branch)  since it wasn't actually needed

PR #444, opened 2018-09-01
Numeral updates

Kline-:


src/Script.js
```diff
@@ -31,15 +31,14 @@ import {AllServers, processSingleServerGrowth}  from "./Server";
 import {Settings}                               from "./Settings";
 import {post}                                   from "./ui/postToTerminal";
 import {TextFile}                               from "./TextFile";
-
 import {parse, Node}                            from "../utils/acorn";
 import {Page, routing}                          from "./ui/navigationTracking";
+import numeral                                  from "numeral/min/numeral.min";
```
kopelli: I'd be interested in the difference that bringing in the minified version has vs. bringing in the normal version and letting Webpack minify the entire bundle.

Kline-:
Would need to sweep every file. I just looked up what the import was from
StockMarket.js and pulled that in.

PR #451, opened 2018-09-07
Server Randomization

kopelli:
Based on a conversation in the #noobsgame Discord channel related to the usefulness of running scripts on the non-purchased servers, this change separates the initial server data from how they are created. It also gives some flexibility of randomizing more of the server information, such as a range of RAM, or how many network layers away from `home` it is.

This change doesn't change any of the existing server information, only separates it out. I'll leave changing static values to ranges up to @danielyxie 😄.

### DIFF
--- a/src/data/servers.ts
+++ b/src/data/servers.ts
@@ -54,7 +54,7 @@ interface IServerMetadata {
      * This value is between 1 and 15.
      * If this is not populated, @specialName should be.
      */
-    networkLayer?: number;
+    networkLayer?: number | IMinMaxRange;
 
     /**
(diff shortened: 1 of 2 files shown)


src/data/servers.ts
```diff
@@ -0,0 +1,1457 @@
...
+}
+
+/**
+ * The metadata for building up the servers on the network.
+ */
+export const serverMetadata: IServerMetadata[] = [
+  {
+    hackDifficulty: 99,
+    hostname: "ecorp",
+    moneyAvailable: {
```
danielyxie: Is there any reason not to use 70e9/30e9 here?

PR #472, opened 2018-09-28
.mainmenu li a {cursor:pointer}

Blacksilver42:
<3


danielyxie:
This needs to be changed in the source file (/css/mainmenu.scss) and not the distribution files. I went ahead and implemented it in dev. You should see it in the next update

Blacksilver42:
I don't know what I'm doing lol

PR #488, opened 2018-10-16
Adding Ram Costs to Docs

CBJamo:
Fairly straightforward, just adds ram costs to the docs. This PR only has the basic functions, because I wasn't sure if this would be the best formatting or not. If the style/formatting looks good, I'll do the same for the other functions.


CBJamo:
Sounds good, do you know if spinx can pull directly from the Constants.js file? It would be obviously be better if these numbers only had to be maintained in one place. In either case, ping me after you've made those changes and I'll do the other functions.

PR #549, opened 2019-03-12
Update getStamina.rst

hydroflame:
The doc for bladeburner getStamina no longer proposes to use `getStaminaPercentage` in the example.


danielyxie:
What's your reasoning for this change?

hydroflame:
The function is literally called getStamina. In game it's called getStamina. It's just the example in the doc that's wrong

hydroflame:
I'm an idiot who can't read doc

PR #554, opened 2019-03-12
Update Bladeburner.js

hydroflame:
https://github.com/danielyxie/bitburner/blob/master/src/Bladeburner.js#L1420

It's technically true that it increases your stamina gain but that's only because stam gain is a percentage of max stamina. Hyperbolic chamber directly increases stamina not stamina per second.

### DIFF
--- a/src/Bladeburner.js
+++ b/src/Bladeburner.js
@@ -3953,7 +3953,7 @@ function initBladeburner() {
     GeneralActions[actionName] = new Action({
         name: actionName,
         desc: "Enter cryogenic stasis using the Bladeburner division's hi-tech Regeneration Chamber. " +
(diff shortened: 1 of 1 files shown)


danielyxie:
Shouldn't it be just "stamina" then? i.e.

"slightly increase your stamina"

instead of

"slightly increase your max stamina"

hydroflame:
Well the description for "Training" says "increases your max stamina." so I mirrored it to hyperbolic chamber.

danielyxie:
Training actually does increase your max stamina. Hyperbolic Chamber increases just your stamina. 

I've updated the description to just be 

"This will slowly heal your wounds and slightly increase your stamina."

PR #555, opened 2019-03-12
added getBlackOpRank

hydroflame:
Added `getBlackOpRank` function to bladeburner API and related documentation.


danielyxie:
Both `AceNetscriptMode` and `CodeMirrorNetscriptMode` need to be updated with function names, for future reference (I've already done it locally for CodeMirror)

PR #570, opened 2019-03-23
Added dropdown class to select tags

koriar:
As requested by hydroflame, I added the dropdown class to the dropdown menus.
... I don't know anything about React though, so I left the code in the Corporation UI alone.


koriar:
Updated dropdown menus in Sleeves:

image

PR #576, opened 2019-03-23
infiltration now scale exp gained by some factor to prevent some abuse

hydroflame:
exp^0.7

```
                     1k       10k      100k     1m       10m
before: {1, 10,      1000,    10000,   100000,  1000000, 10000000}
after:  {1, 5.01187, 125.893, 630.957, 3162.28, 15848.9, 79432.8}
```

the best gym gives 13/s, you could get million and millions before. I'm also thinking maybe there should be a max security before the army get's involved and kills you?


koriar:
But hydroflame, there IS a limit on the amount of exp you can get!

```
InfiltrationInstance.prototype.gainStrengthExp = function(amt) {
    if (isNaN(amt)) {return;}
    this.strExpGained       += amt;
}
```

:D

hydroflame:
1e308?

PR #623, opened 2019-05-25
Update script.module immediately

Ornedan:
Fixes the problem where starting multiple copies of a script when that script's module has not been loaded yet leads to every instance of the script having a duplicate module eval'd for it.

Test case:

bug-controller.ns
```js
    const WORKER_COUNT = ns.args[0];
    
    for (let n = 0; n < WORKER_COUNT; n++) {
        await ns.exec("bug-worker.ns", "home", 1, n);
    }
}
```

bug-worker.ns
```js
    await ns.sleep(200);
    debugger;
}
```

### DIFF
--- a/src/NetscriptJSEvaluator.js
+++ b/src/NetscriptJSEvaluator.js
@@ -26,9 +26,9 @@ export async function executeJSScript(scripts = [], workerScript) {
         // load fully dynamic content. So we hide the import from webpack
         // by placing it inside an eval call.
         urlStack = _getScriptUrls(script, scripts, []);
-        script.module = await eval('import(urlStack[urlStack.length - 1])');
+        script.module = new Promise(resolve => resolve(eval('import(urlStack[urlStack.length - 1])')));
     }
-    loadedModule = script.module;
+    loadedModule = await script.module;
 
     let ns      = workerScript.env.vars;
 


Ornedan:
The `module` field is only read in NetscriptJSEvaluator.js, in that one function. All other places that touch the field overwrite the old module, if any, with an empty string. The effect of this patch is to ensure that if a particular version of a NS script is loaded by the game, the game loads only one copy of the script's module graph.
The old code already attempted to do this, but did not consider the case where multiple instances of the script are being started with a cold cache.

Type? Before it was `script: string | Module`, now it is `script: string | Promise<Module>`. Though, unfortunately, the browser `Module` type is not visible at compile-time so that can't actually be expressed in the code.

Mess up? No, I tested the change locally and all it does is improve speed and stability. No more freezing the tab when starting many copies of the same script without the script already cached.

Clarify semantics? By changing the type in Script.ts from `any` to `string | Promise<any>` and saying in the documentation that the `any` is actually `Module`, maybe?

Ornedan:
Try the test case scripts.

Test 1
=====
- reload Bitburner tab
- `run bug-controller.ns 1`
- wait a few seconds until `ps` shows no scripts running
- `run bug-controller.ns 10000`

Test 2
====
- reload Bitburner tab
- `run bug-controller.ns 10000`

Without my patch, there will be a major difference in how fast the 10k scripts start up between the tests. Test 2 is also likely hit the more severe problem where Bitburner tries to import far too many copies of the script and goes unresponsive under the CPU load.

jaguilar:
Do we need the variable at all, then, if it's not accessed anywhere? Should we instead just rename it to "hasUpToDateModule" and make it a boolean?

Ornedan:
Yes. `Script.module` is the cached module, which is pretty important for performance.

Ornedan:
Also, both read and write are "access". The field is read in that one function, but written to in several places. The point of which writes is to clear away a cache that just became stale due to other things that code is doing.

PR #626, opened 2019-06-02
Close dialog boxes on pressing escape

jaguilar:
This is consistent with the way other popups are handled (e.g. logfiles). And it's just nicer this way.

I tested that all the other methods of closing these boxes still work.


jaguilar:
Please ignore the branch name. I had a notion of doing a thing but it proved more complicated than I wanted.

PR #629, opened 2019-06-02
recompile ns2 scripts if dependencies change

### LINKED ISSUE #628: NS2 modules should be recompiled whenever any of their dependencies change
Suppose I have module A like

    export function do(ns) { ns.tprint("foo"); }

and module B like
    
    import {do} from "A.js";
    export async function main(ns) { do(); }

then I run B.js once, before changing module A to print "bar" instead. If I run B.js again, "foo" will still be printed. This is because when running B, we only check if B has changed, not if any of its imports have. If B has not changed, we use the module we originally compiled for B, including the old versions of module A.

We could fix this any number of ways. One simple approach would be to store a global sequence number, which is incremented and assigned to each worker script when it is compiled. We would also store with each script a listing of its dependencies. When we go to run the script, we check whether any dependency has changed by comparing the stored module's sequence number to the sequence number of all of its dependencies. If any of the dependencies currently have a higher sequence number, that means that the current script should be recompiled. Or, instead of a sequence number, we could just timestamp all the scripts, which would have much the same effect.

jaguilar:
Fixes #628. Previously, if you changed a dependency A of script B, script B would use the old version of A until it was recompiled. Now, script B will use the new version of A the next time it runs.

I tested this manually, and also imported my real game for a little while and ran into no issues.

### DIFF
--- a/src/NetscriptFunctions.js
+++ b/src/NetscriptFunctions.js
@@ -1139,7 +1139,7 @@ function NetscriptFunctions(workerScript) {
                     var oldScript = destServer.scripts[i];
                     oldScript.code = sourceScript.code;
                     oldScript.ramUsage = sourceScript.ramUsage;
-                    oldScript.module = "";
+                    oldScript.module.markUpdated();
                     return true;
                 }
             }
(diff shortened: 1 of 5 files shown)


Ornedan:
I'd recommend using a logical clock for `updateTimestamp` instead of real time clock. The system clock's value is not guaranteed to be monotonically increasing - it can go backwards if the system clock had run ahead and then syncs with a time server. Also browsers specifically fuzz their reported clock values a bit in order to mitigate the Spectre vulnerability.

While this is an unlikely problem, if anyone ever gets hit by it it's going to make for a bizarre and hard-to-reproduce bug report.

jaguilar:
Seems like a reasonable change to suggest. I'll do it.

jaguilar:
Done, PTAL

PR #649, opened 2019-07-22
Upgrade terminal parser

helldragger:
Testing the grammar itself needs ANTLR4 to be installed on the PC, the file TerminalInput.txt provides a rather complete sample of test with every edgecase I could think of.
see here : https://github.com/antlr/antlr4/blob/master/doc/javascript-target.md

The new parser allows:
- **automatic argument type detection**, manipulating those means creating a new custom listener (see the custom listener CommandListener as an exemple), including numbers, short flags, long flags, words, simple quoted strings, double quoted strings and key value pairs.
- **automatic string misescaping detection** is also now possible, but not yet implemented, see the custom listener system, where being to act when entering and exiting a node of the abstract syntax tree allows to keep track of such errors.
- it also allows **multi line scripting**, with or without semi colon.

It has been implemented and tested in lieu of the previous .split(";") which did not take into account strings.

Testing the grammar can be done using either the grun tool shipped with antlr, or using the antlr parse tree extension for visual code.


helldragger:
ah wait a file is missing, and it's not requesting the right branch, brb

PR #652, opened 2019-07-31
Filesystem rework

helldragger:
- refactorized every terminal command and the programs into executables as part of the servers core library found in /Server/lib, added unit tests for all these executables (approx 300),
 - Added a new filesystem based on memfs to allow each server to have a real file system. 
 - Added relevant tools for it (ls, tree, bash filtering, cp, mkdir, ...), plus a versioning system when overwriting files.
 - Adapted some netscripts functions (ls, scp...) to be used a wrappers around the aformentioned executables, hence covering them in unit tests.
 - Added more in game documentation and reformated the help data format.
 - Upgraded some systems such as the script search to use a map, resulting in a O(1) lookup instead of searching in a list every time.

Finally, at the current state, I playtested the game for a few hours, no bug, errors or crash occured, seems to be working fine and all tests pass.

N.B.: Still, having a way to test every ns functions in game to stress test the game could be a great thing, but I have no idea how to test this in a browser automatically.


danielyxie:
Did a manual merge as recommended by Github. Annoyingly it didn't seem to update this PR. All the changes seem to be there though (I'm assuming because it was done by a different user?). All the changes seem to be there though, so closing this out. This is an awesome change, thanks a bunch! Looking forward to testing it out

PR #658, opened 2019-12-14
Fix ls

helldragger:
fixed ls behaviour to work like the default ubuntu ls behaviour.

Tests have been modified to reflect this, ls is working, all tests pass, and download has been updated to deal with the new ls output format.


helldragger:
also, I saw you tried to modify the output of the ls to include color, please don't do this in the ls directly, check the way the ls output is dealt with in the download.ts, the output is supposed to be used by various things, ls just outputs pure strings, please add colors out of it by passing a custom out function where necessary

PR #667, opened 2019-12-16
Origin/fix nano not creating when in subdirectory

helldragger:
-  also fixes aliases output management, 
-  run not parsing script mono arguments correctly, 
-  add the touch command to create files on the fly without editing them
-  add tests to nano and touch

### DIFF
--- a/src/Terminal.js
+++ b/src/Terminal.js
@@ -884,9 +884,15 @@ let Terminal = {
                 };
 
                 executable(server, Terminal, postForLsCommand, postError, commandArray.splice(1));
-            } else {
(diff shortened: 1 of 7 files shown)


src/Terminal.js
```diff
@@ -844,23 +844,12 @@ let Terminal = {
 
         /****************** END INTERACTIVE TUTORIAL ******************/
 
-        this.executeCommandHelper(command);
-    },
-
-    // TODO Rename and refactor this. Temporary fix to make Interactive Tutorial work with new filesystem
-    executeCommandHelper: function(command) {
```
danielyxie: I think removing this breaks the interactive tutorial

PR #674, opened 2020-05-04
harden JS against prototype tampering

Farbdose:
Prevent exploits like https://gitlab.com/Farbdose/bitburner/-/blob/master/ev4l.js#L10


Farbdose:
> just please don't freeze Number.prototype.toLocaleString if you want to expand on this on the future, unless that option would be configurable

Todo: carefully add `Number` to the list

hydroflame:
So this may seem weird but this will not be fixed. Instead the game is going to embrace exploits and award the player for finding them.

PR #818, opened 2021-03-20
v0.50.0

hydroflame:
v0.50.0 - 2021-03-20 Intelligence (hydroflame)
    -------
    Intelligence
    * int exp gain and effect has been reworked. It is now much more easy to
      acquire and far more powerful. The goal here is to feel like players have
      another tool in their arsenal.

    Factions
    * Hacking factions no longer have hacking level requirements since their associated servers do.

    Misc.
    * Sleeve styling.
    * number formatting
    * remove wiki button in Hacking Missions.
    * Fix NaN displayed when very very large numbers are reached.


hydroflame:
So int gain has been reviewed and boosted. I estimated how much of each feature that gives int experience would be used over the course of 6 bitnodes and aimed to make the player reach level 80-100 int.


All formulas that use int have been reviewed so that they are more powerful. Most of them now use percentage bonus based on int.

Players who just acquired int will get a bonus of around 1.5% to many actions.
Player who played the game normally will bump this bonus to around 6-7%
Players who specifically farm int exp can reach ~13-14%.

PR #833, opened 2021-04-03
Fix typo in example documentation

Pimvgd:
processes variable went unused and ps was used which wouldn't have worked


Pimvgd:
wrong base branch

PR #834, opened 2021-04-03
Fix typo in example documentation for ps

Pimvgd:
processes variable went unused and ps was used which wouldn't have worked


hydroflame:
I don't know how these merge conflicts arised but I have copied your changes. Thank you!

PR #915, opened 2021-05-02
Noodles

hydroflame:


src/NetscriptFunctions.js
```diff
@@ -1619,7 +1619,7 @@ function NetscriptFunctions(workerScript) {
         },
         getBitNodeMultipliers: function() {
             updateDynamicRam("getBitNodeMultipliers", getRamCost("getBitNodeMultipliers"));
-            if (SourceFileFlags[5] <= 0) {
+            if (Player.bitNodeN !== 5 && SourceFileFlags[5] <= 0) {
```
Pimvgd, line 1622: Seems weird that the check is flipped compared to the one in getServer (line 1630) but okay

src/Terminal.jsx
```diff
@@ -1406,25 +1410,34 @@ let Terminal = {
                     if (commandArray.length < 2) {
                         postError("Incorrect number of arguments. Usage: tail [script] [arg1] [arg2]...");
                     } else {
-                        const scriptName = Terminal.getFilepath(commandArray[1]);
-                        if (!isScriptFilename(scriptName)) {
-                            postError("tail can only be called on .script files (filename must end with .script)");
-                            return;
-                        }
+                        if(typeof commandArray[1] === 'string') {
```
Pimvgd, line 1413: else-if chain can be merged (body of else consists of nothing but if statements

Pimvgd: Alternatively, put a return and remove the else since all other error messages seem to follow that format. Easier to follow that way.

Pimvgd:
Reviewed that the code which has been written is sensible. Not reviewed whether there's missing code. I feel like the code changes are the kind of code changes that would be made with regards to the commit messages.

PR #934, opened 2021-05-04
[SPOILERS] Stanek's gift

hydroflame:


src/Augmentation/data/AugmentationNames.ts
```diff
@@ -107,6 +107,13 @@ export const AugmentationNames: IMap<string> = {
     BladeArmorOmnibeam:                 "BLADE-51b Tesla Armor: Omnibeam Upgrade",
     BladeArmorIPU:                      "BLADE-51b Tesla Armor: IPU Upgrade",
     BladesSimulacrum:                   "The Blade's Simulacrum",
+    StaneksGift:                        "Stanek's Gift",
+    StaneksGiftI:                       "Stanek's Gift - II",
+    StaneksGiftII:                      "Stanek's Gift - III",
+    StaneksGiftAscensionPort:           "Stanek's Gift - Ascension",
+    StaneksGiftNeckExpansion:           "Stanek's Gift - Neck expansion",
+    StaneksGiftVertabraeExpansion:      "Stanek's Gift - Vertabrae expansion",
+    StaneksGiftCranialExpansion:        "Stanek's Gift - Cranial expansion",
```
ModdedGamers, line 116: You should probably make the names more consistant here. My suggestions:

1. Change the internal names to be numbers instead of roman numerals(`StaneksGift1`, `StaneksGift2`, etc)
2. Change the numbers to be more consistant with Neurotrainer(Lines 59 - 61)I suggest changing the names to Stanek's Gift I, Stanek's Gift II, and Stanek's Gift III
3. Change the expansion naming to be more consistent with the Blade Armor naming(using colon-space instead of space-hyphen-space).
I understand if you don't like these changes, they're just suggestions.

PR #953, opened 2021-05-10
Add basic docker support for development

MartinFournier:
Assuming docker engine is installed, it is now possible to run the dev
server using ```docker-compose up --build``` or run the production version using
```docker build -t bitburner . && docker run -it -p 8000:80 bitburner```.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,14 @@
+node_modules/
+
+.git
+.gitattributes
+.gitignore
+.editorconfig
+
+.dockerignore
+Dockerfile
+docker-compose.yml
+
+*.md
+Quotes.txt
+netscript_tests/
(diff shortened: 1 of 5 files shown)


Daniferrito:
Looks good to me!
You added all the things that I would nitpick about docker, like a comprehensive .dockerignore, multistage build, using alpine for prod...
It even has a working dev container with auto reloading, which is not something that I had seen before inside docker, but that I want now for all my projects.

The only thing that I would recomment would be to use docker buildkit's --mount=type=cache, to speed up subsequent prod compilations but it seems it has not reached stable yet.

Maybe a nice adittion would be to have a prod compilation that saves the output to the corresponding folders, instead of generating an image. That way the deployment is the same but the compilation environment can be the same across any machine.
But probably it would be better to implement this through GithubActions.

PR #954, opened 2021-05-11
Bladeburner fixes

Daniferrito:
Added time overflow for the Bladeburner tick update (when finishing actions and when calculating random events)
Added autofocus for the input in the popups for the "Set team size"


hydroflame:
This is dope.

PR #957, opened 2021-05-12
Brachiblade upgrade name typo

Daniferrito:


hydroflame:
This is actually a very complicated fix because it'll break anyone who has this aug name.

Daniferrito:
True, I did not think of that.
Then i would suggest simply to close this PR. The fix is not that relevant, and a system to upgrade savegames might be added later that allows for this change.

PR #959, opened 2021-05-13
Add setting to suppress Bladeburner prompts

MartinFournier:
Used to suppress the message that is shown when your Bladeburner action
is cancelled when busy with something else. Useful once you have BN-7 and have a watcher script restarting the action as needed.

firefox_XHH79zAnkg

Let me know if there's something else that needs to be added to make save files compatible.

### DIFF
--- a/src/Bladeburner.jsx
+++ b/src/Bladeburner.jsx
@@ -24,6 +24,7 @@ import {
     convertTimeMsToTimeElapsedString,
 } from "../utils/StringHelperFunctions";
 
+import { Settings } from "./Settings/Settings";
 import { ConsoleHelpText } from "./Bladeburner/data/Help";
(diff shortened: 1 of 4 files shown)


hydroflame:
hmm. The settings can't be visible before bladeburners are available.

MartinFournier:
Would adding display:'none' to the surrounding fieldset when `!Player.canAccessBladeburner()` work? It could lead to the flag being set to true while in BN-8 and not being able to unset it though. Also there would be spoilers in the markup if someone inspects.

If that's alright, I committed that modification.

hydroflame:
Yeah that works.

hydroflame:
You set the field on game load. It should be re-evaluated every time the menu comes up.

hydroflame:
I'm gonna merge and fix it.

PR #984, opened 2021-06-02
Modified numeralFormat.ts to prevent NaNt from appearing

Tesseract1234567890:


hydroflame:
Ok well this code is repeated like 7 times. Let's make a method instead.

src/ui/numeralFormat.ts
```diff
@@ -136,14 +140,11 @@ class NumeralFormatter {
     }
 
     formatShares(n: number): string {
-        if (n < 1000) {
```
hydroflame, line 139: Removing this is incorrect. Shares are always whole number.

Tesseract1234567890: I fixed that bad boy

PR #997, opened 2021-06-06
Allowed Sleeves UI to update on workout change via API

Tesseract1234567890:


hydroflame:
That's not it. 
If I run the script 
```ns.sleeve.setToGymWorkout(0, 'Powerhouse Gym', 'def');```
The ui will display
Screen Shot 2021-06-06 at 2 22 48 PM

Tesseract1234567890:
It should be fixed now

PR #998, opened 2021-06-06
Comp. Rep. Focus Fix

Tesseract1234567890:
This was brought to my attention by kwazygloo


hydroflame:
Uh, applying for a new job shouldn't reset your work status.

Tesseract1234567890:
That should only do it if the player gets the new job.

hydroflame:
Right, but getting a new job isn't the issue. The issue is that the rep is transfered.

Tesseract1234567890:
This should fix the reputation being transferred. Additionally, it will cancel the job if the player is accepted into a new job.

PR #1000, opened 2021-06-07

Kwazygloo:


hydroflame:
This is different than what was already there.

Kwazygloo:
I thought I would try to keep the "original intent" of the descriptions but remove the literal repeat of descriptions for iteration 2 and 3 and provide  more detail and a sense of progression to the descriptions of each iteration.

Issue #1005, opened 2021-06-08
Company Reputation Bug


Kwazygloo:
Working for a company and switching  to another without cancelling will transfer the reputation queued to the new company.

PR #1006, opened 2021-06-08
Comp. Rep. Focus Fix

Tesseract1234567890:
Resolves issue #1005


Tesseract1234567890:
I think I got it this time

PR #1017, opened 2021-06-18

Kwazygloo:
Unique augment proposed for Aevum, since it is the only city that does not have a unique augment. The augment is themed around "luck". All of the modifiers it provides are 7.77% or 77.7% and the flavor text makes use of Aevum's casino. It would also be a neat gimmick to have it provide DeepscanV1.exe and AutoLink.exe upon a reset.

### DIFF
--- a/src/Augmentation/AugmentationHelpers.jsx
+++ b/src/Augmentation/AugmentationHelpers.jsx
@@ -1681,20 +1681,20 @@ function initAugmentations() {
              "uses advanced mathematical algorithims to rapidly identify and compute statistical " +
              "outcomes of nearly every situation. The use of this implant is outlawed for use in " +
              "Aevum's casino.<br><br>" +
-             "Increases charisma by 7%.<br>" +
-             "Increases charisma experience gain by 7%.<br>" +
-             "Increases money gained from working by 7%.<br>" +
-             "Increases reputation gain from factions and companies by 7%.<br>" +
-             "Increases crime success rate by 7%.<br>" +
-             "Increases money gained from committing crimes by 7%.<br>" +
+             "Increases charisma by 7.77%.<br>" +
+             "Increases charisma experience gain by 7.77%.<br>" +
+             "Increases money gained from working by 77.7%.<br>" +
+             "Increases reputation gain from factions and companies by 7.77%.<br>" +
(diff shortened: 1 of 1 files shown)


src/Augmentation/AugmentationHelpers.jsx
```diff
@@ -1675,6 +1675,32 @@ function initAugmentations() {
 	// Aevum
     // TODO Later Something that lets you learn advanced math...this increases int
     // and profits as a trader/from trading
+    const PCMatrix = new Augmentation({
+        name:AugmentationNames.PCMatrix, repCost:1e4, moneyCost:1e7,
```
danielyxie, line 1679: You need to add the `PCMatrix` property to the `AugmentationNames` (`AugmentationNames.ts`) object, which as far as I can tell you haven't done here.

src/Augmentation/AugmentationHelpers.jsx
```diff
@@ -1675,6 +1675,32 @@ function initAugmentations() {
 	// Aevum
     // TODO Later Something that lets you learn advanced math...this increases int
```
danielyxie, line 1676: We can remove this comment now

src/Augmentation/AugmentationHelpers.jsx
```diff
@@ -1675,6 +1675,32 @@ function initAugmentations() {
...
+             "uses advanced mathematical algorithims to rapidly identify and compute statistical " +
+             "outcomes of nearly every situation. The use of this implant is outlawed for use in " +
+             "Aevum's casino.<br><br>" +
+             "Increases charisma by 7.77%.<br>" +
+             "Increases charisma experience gain by 7.77%.<br>" +
+             "Increases money gained from working by 77.7%.<br>" +
+             "Increases reputation gain from factions and companies by 7.77%.<br>" +
+             "Increases crime success rate by 7.77%.<br>" +
+             "Increases money gained from committing crimes by 7.77%.<br>" +
+             "Provides DeepscanV1.exe and AutoLink.exe after a reset.",
```
danielyxie, line 1690: This effect hasn't been implemented yet

danielyxie (changes requested):

PR #1079, opened 2021-08-21
Instakill player when automating infiltration

threehams:
Make it clear that infiltration is not meant to be automated by killing the player instantly if they send in untrusted events.
(This would have saved me some time today, plus I think it's fun.)

I'd like to have a custom message when this happens, but I might need some guidance on the best way to do that.

### DIFF
--- a/src/Infiltration/ui/KeyHandler.tsx
+++ b/src/Infiltration/ui/KeyHandler.tsx
@@ -2,6 +2,7 @@ import React, { useEffect } from 'react';
 
 interface IProps {
     onKeyDown: (event: React.KeyboardEvent<HTMLElement>) => void;
+    onFailure: (options?: { automated: boolean }) => void;
 }
 
 export function KeyHandler(props: IProps): React.ReactElement {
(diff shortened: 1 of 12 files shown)


src/Infiltration/ui/KeyHandler.tsx
```diff
@@ -2,14 +2,20 @@ import React, { useEffect } from 'react';
...
 
 export function KeyHandler(props: IProps): React.ReactElement {
     let elem: any;
     useEffect(() => elem.focus());
 
     function onKeyDown(event: React.KeyboardEvent<HTMLElement>): void {
-        if(!event.isTrusted) return;
+        console.log("isTrusted?", event.isTrusted)
+        if(!event.isTrusted) {
+            console.log("untrusted event!")
```
sschmidTU, line 15: console.log should be removed eventually

sschmidTU:
This is a neat balancing mechanic, though the question is if we should really punish people for automating the UI, after they've already gone through the effort of doing that. People who automate UI minigames probably know they're breaking the balance. It's a design question whether to allow it or not. We might also break some people's code who are automating the whole game.

PR #1122, opened 2021-09-02
Support optional chaining in static RAM calculator.

threehams:
Upgrade acorn and set it to use latest ecmaVersion.

`acorn` changelog: https://github.com/acornjs/acorn/blob/master/acorn/CHANGELOG.md

The breaking changes in 7.0.0 and 8.0.0 don't affect this project, as nothing is following dynamic imports.


sschmidTU:
I remember breaking RAM calculation with optional chaining (`?.`) a while ago and being stuck on why for hours :)

PR #1167, opened 2021-09-05
Add browser tests for tutorial, NetScript 2

threehams:
This adds browser tests for:
- tutorial from start to finish
- basic create/run for a NetScript 2.0 script.
- static vs. dynamic RAM check

This uses [Cypress](https://www.cypress.io/) which is extremely fast and does automatic waiting.
Additional selectors are added with [cypress-testing-library](https://testing-library.com/docs/cypress-testing-library/intro/). This adds find by text (great!) and ensures elements are accessible with assistive technologies - like buttons are really buttons.

To run:
```
npm run cy:dev
```
To run against a production server:
```
npm run build
npm run cy:test
```

Click "Run 2 integration specs" on the upper right:
Screenshot - 9_5_2021 , 12_25_13 AM


.browserslistrc
```diff
@@ -0,0 +1,3 @@
+last 4 versions
```
threehams, line 1: Both `browserslist` and `.browserslistrc` are supported, but the first one crashes Cypress 🤷

PR #1185, opened 2021-09-07
Jest cypress types

threehams:
Separate src, test, and cypress TS types. Cypress and Jest both have `describe`, `it`, and `expect`, but they're slightly different. By using multiple `tsconfig.json` files , called TS [project references](https://www.typescriptlang.org/docs/handbook/project-references.html), it's possible to scope global types to only one part of a project.

This also removes GitHub Actions which was an experiment not really ready to merge.

This reverts the previous fix for tests.


jest.config.js
```diff
@@ -4,6 +4,6 @@ module.exports = {
   transform: {
     "^.+\\.(js|jsx|ts|tsx)$": "babel-jest",
   },
-  // testMatch: ["**/?(*.)+(test).[jt]s?(x)"],
+  testMatch: ["**/?(*.)+(test).[jt]s?(x)"],
```
threehams, line 7: This avoids jest picking up Cypress `.spec.ts` files

PR #1256, opened 2021-09-15
Fix import not working

hydroflame:


sschmidTU:
I don't really get how the PR fixes imports. Is it just the new build with changes that were already in? Or is it just the .scss change? Otherwise it's mostly removing 2 files, one of which was initializing labels (i guess it wasn't used or something).

hydroflame:
`<input type="file" id="import-game-file-selector" name="file"/>` was missing

PR #1352, opened 2021-09-22
Work Focus Penalty Reduction Augmentation

Nolshine:


src/Augmentation/Augmentation.tsx
```diff
@@ -54,7 +54,7 @@ interface IConstructorParams {
   programs?: string[];
 }
 
-function generateStatsDescription(mults: IMap<number>, programs?: string[], startingMoney?: number): JSX.Element {
+function generateStatsDescription(mults: IMap<number>, programs?: string[], startingMoney?: number, focusBonus?: number): JSX.Element {
```
hydroflame, line 57: Revert, use stats instead. Like RedPill or HiveMind

src/Augmentation/Augmentation.tsx
```diff
@@ -390,6 +390,10 @@ export class Augmentation {
   // The Player/Person classes
   mults: IMap<number> = {};
 
+  // Focus penalty reduction given by this Augmentation.
+  // Isn't really a multiplier, so it's out here on its own.
+  focusBonus?: number;
```
hydroflame, line 395: Unused

Nolshine: oops

hydroflame:
Added independelty, credits will still be given

PR #1386, opened 2021-09-26
allow inputting rep amt wanted in faction donation

Nolshine:
e.g. instead of inputting 5m in cash you can input that you want 10 rep and it'll set the donation and show the cost.


hydroflame:
Ok so that's going to sound weird but I'm going to say no to that. The reason is kindof weird but it's because the conversion, although not very difficult, is something the player can write scripts for. If I remove that task it's one less thing for the player to do.

PR #1534, opened 2021-10-17
Issue 1522

Tyasuh:
Sleeves commit scripted crimes fix.

### DIFF
--- a/src/Crime/CrimeHelpers.ts
+++ b/src/Crime/CrimeHelpers.ts
@@ -30,7 +30,7 @@ export function determineCrimeSuccess(p: IPlayer, type: string): boolean {
   }
 }
 
-export function findCrime(roughName: string): Crime | null {
+export function findCrime(roughName: string): Crime | null {roughName = roughName.toLowerCase();
   if (roughName.includes("shoplift")) {
     return Crimes.Shoplift;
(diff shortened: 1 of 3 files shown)


src/Crime/CrimeHelpers.ts
```diff
@@ -30,7 +30,7 @@ export function determineCrimeSuccess(p: IPlayer, type: string): boolean {
   }
 }
 
-export function findCrime(roughName: string): Crime | null {
+export function findCrime(roughName: string): Crime | null {roughName = roughName.toLowerCase();
```
hydroflame, line 33: Should be on a separate line.

src/NetscriptFunctions/Bladeburner.ts
```diff
@@ -326,7 +326,7 @@ export function NetscriptBladeburner(
       checkBladeburnerCity("getCityEstimatedCommunities", cityName);
       const bladeburner = player.bladeburner;
       if (bladeburner === null) throw new Error("Should not be called without Bladeburner");
-      return bladeburner.cities[cityName].commsEst;
+      return bladeburner.cities[cityName].comms;
```
hydroflame, line 329: The function should be renamed to `getCityCommunities` because it's no longer an estimate, all the strings should be changed as well.

The documentation for that function needs to be updated as well.

PR #1763, opened 2021-11-22
Return the price paid/recieved on short/sellShort

alainbryden:


hydroflame:
This is correct and will be merged but I need to put it in a numbered version.

PR #1797, opened 2021-11-27
Bad copy pasta from isFocused -> setFocus

alainbryden:


alainbryden:
This PR is obsolete if you decide to delete the setFocus function instead and just add a new bool to workForComapny / workForFaction. (I think isFocused might still be useful in its own right though)

PR #1874, opened 2021-12-13
Reimplement setFocus

theit8514:
Work functions do not take focus from the user.

setFocus readded and updated to send the user to the Work page if focusing or to the Terminal if not focusing.

### DIFF
--- a/src/NetscriptFunctions/Singularity.ts
+++ b/src/NetscriptFunctions/Singularity.ts
@@ -829,6 +829,8 @@ export function NetscriptSingularity(
         return false;
       }
 
+      const wasWorking = player.isWorking;
+      const wasFocused = player.focus;
       if (player.isWorking) {
         const txt = player.singularityStopWork();
         workerScript.log("workForCompany", () => txt);
@@ -839,7 +841,11 @@ export function NetscriptSingularity(
       } else {
         player.startWork(companyName);
       }
-      player.stopFocusing();
+      
+      if (!wasWorking || (wasWorking && !wasFocused))
+        player.stopFocusing();
+      else if (wasWorking && wasFocused)
+        player.startFocusing();
       workerScript.log(
         "workForCompany",
         () => `Began working at '${player.companyName}' as a '${companyPositionName}'`,
@@ -979,6 +985,8 @@ export function NetscriptSingularity(
         return false;
       }
 
+      const wasWorking = player.isWorking;
+      const wasFocused = player.focus;
       if (player.isWorking) {
         const txt = player.singularityStopWork();
         workerScript.log("workForFaction", () => txt);
@@ -1077,7 +1085,10 @@ export function NetscriptSingularity(
             return false;
           }
           player.startFactionHackWork(fac);
-          player.stopFocusing();
+          if (!wasWorking || (wasWorking && !wasFocused))
+            player.stopFocusing();
+          else if (wasWorking && wasFocused)
+            player.startFocusing();
           workerScript.log("workForFaction", () => `Started carrying out hacking contracts for '${fac.name}'`);
           return true;
         case "field":
@@ -1088,7 +1099,10 @@ export function NetscriptSingularity(
             return false;
           }
           player.startFactionFieldWork(fac);
-          player.stopFocusing();
+          if (!wasWorking || (wasWorking && !wasFocused))
+            player.stopFocusing();
+          else if (wasWorking && wasFocused)
+            player.startFocusing();
           workerScript.log("workForFaction", () => `Started carrying out field missions for '${fac.name}'`);
           return true;
         case "security":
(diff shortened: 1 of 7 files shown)


theit8514:
Rebased from dev without extra commit

src/PersonObjects/Player/PlayerObjectGeneralMethods.tsx
```diff
@@ -842,15 +838,13 @@ export function startFactionWork(this: IPlayer, router: IRouter, faction: Factio
   this.workRepGainRate *= BitNodeMultipliers.FactionWorkRepGain;
 
   this.isWorking = true;
-  this.focus = true;
   this.workType = CONSTANTS.WorkTypeFaction;
   this.currentWorkFactionName = faction.name;
 
   this.timeNeededToCompleteWork = CONSTANTS.MillisecondsPer20Hours;
-  router.toWork();
```
alainbryden, line 850: Hmm, I could see this going wonky if the user is already focused / working say for a faction, and then a script changes them to working for a company.

I think in all these cases you need to check if the player is focused and still route to the corresponding work page if they were.

src/NetscriptFunctions/Singularity.ts
```diff
@@ -622,6 +622,31 @@ export function NetscriptSingularity(
...
+      if (!(player.workType == CONSTANTS.WorkTypeFaction || player.workType == CONSTANTS.WorkTypeCompany || player.workType == CONSTANTS.WorkTypeCompanyPartTime)) {
+        throw helper.makeRuntimeErrorMsg("setFocus", "Cannot change focus for current job");
+      }
+      if (!player.focus && focus) {
+        player.startFocusing();
+        Router.toWork();
+        return true;
+      } else if (player.focus && !focus) {
+        player.stopFocusing();
+        Router.toTerminal();
```
alainbryden, line 645: I was questioning doing any routing at all on `setFocus(false)`, but I see now that we will *only* route to terminal if we were currently focused (e.g. on a work screen) so this makes sense.

src/NetscriptFunctions/Singularity.ts
```diff
@@ -1060,7 +1096,11 @@ export function NetscriptSingularity(
             workerScript.log("workForFaction", () => `Faction '${fac.name}' do not need help with hacking contracts.`);
             return false;
           }
-          player.startFactionHackWork(Router, fac);
+          player.startFactionHackWork(fac);
+          if (!wasWorking || (wasWorking && !wasFocused))
+            player.stopFocusing();
+          else if (wasWorking && wasFocused)
+            player.startFocusing();
```
alainbryden, line 1103: I think this is a backwards-compatiblity concern, users with "set work" scripts out there will expect it to focus the first time they set some work (which is required to gains full XP unless you have that aug). It looks like this code will *only* start focusing on the new work if they were previously working (and focused).

So I think you have to change this condition to just `if(wasFocused)` - which should be true by default (right?) when the game is loaded up and the player has never had a script explicitly call `setFocus(false)`?

Even then, we might get complaints that players relied on their scripts "putting them back into focused work" after some time if they popped by their computer, manually unfocused to check in on something, and left it unfocused. They would now need to edit their scripts to put them back in focus like they used to.

I'm not sure there's a backwards compatible way to do this unless we go with @hydroflame 's proposal to can the `setFocus` function entirely and instead add a new optional boolean `focus` parameter to each of the 3 work functions.

alainbryden (changes requested):

src/PersonObjects/Player/PlayerObjectGeneralMethods.tsx
theit8514: The company and faction work pages are the same. When switching between the two like:
```js
/** @param {NS} ns **/
    ns.workForFaction("Tetrads", "security");
    ns.setFocus(true);
    await ns.sleep(5000);
    ns.workForCompany();
    await ns.sleep(10000);
    ns.stopAction();
}
```
In the background what happens is that the workForCompany function detects you were working and calls `singularityStopWork` which completes the current working operation, and then starts a new working operation. If you were on the Work page, the text updates automatically to be the Company work description.

src/NetscriptFunctions/Singularity.ts
theit8514: Unfortunately, if the `workForFaction` or `workForCompany` command takes the focus by default, there is no storage of where the user was when focus was taken. If we then call setFocus(false) we have to send the user to some other page than the Work page (which becomes funky and invalid when isWorking false). If the Work screen was not a page but an overlay of the user's focus, this would work much better since we could just close the overlay, but that's beyond my limited React skills.

I originally proposed a focus parameter to the two work functions, so if that's the route that hydroflame wants to take I can refactor this PR.

PR #1895, opened 2021-12-15
Editing git docs

TheCoderJT:
Hopefully, everything works out smoothly.


markdown/bitburner.ns.hackanalyze.md
```diff
@@ -34,8 +34,8 @@ Returns the percentage of the specified server’s money you will steal with a s
 
 
 ```ts
-//For example, assume the following returns 1:
+//For example, assume the following returns 0.01:
 hackAnalyze("foodnstuff");
-//This means that if hack the foodnstuff server, then you will steal 1% of its total money. If you hack using N threads, then you will steal N% of its total money.
+//This means that if hack the foodnstuff server, then you will steal 1% of its total money. If you hack using N threads, then you will steal N*0.01% of its total money.
```
alainbryden, line 39: Same here as above

package-lock.json
```diff
@@ -5,6 +5,7 @@
   "requires": true,
   "packages": {
     "": {
+      "name": "bitburner",
```
alainbryden, line 8: I've been avoiding checking in changes to this file. Maybe it should be gitignored?

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -3707,6 +3707,20 @@ interface Stanek {
 /**
  * Collection of all functions passed to scripts
  * @public
+ * @remarks
+ * <b>Basic ns1 usage example:</b>
+ * ```ts
+ * property.methodName;
```
alainbryden, line 3713: Maybe we should give examples of both methods that are part of a sub-property, and methods that are not - and use real examples to avoid confusion.

e.g.

<b>Basic ns1 usage example:</b>
```ts
// Basic ns functions can be used directly
getHackingLevel();
// Some related functions are gathered within a common namespace
stock.getSymbols()
```

<b>Basic ns2 usage example:</b>
```ts
  // Basic ns functions can be accessed on the ns object
  ns.getHackingLevel();
  // Some related functions are gathered under a sub-property of the ns object
  ns.stock.getSymbols()
}
```

dist/bitburner.d.ts
```diff
@@ -2321,9 +2335,9 @@ export declare interface NS extends Singularity {
      *
      * @example
      * ```ts
-     * //For example, assume the following returns 1:
+     * //For example, assume the following returns 0.01:
      * hackAnalyze("foodnstuff");
-     * //This means that if hack the foodnstuff server, then you will steal 1% of its total money. If you hack using N threads, then you will steal N% of its total money.
+     * //This means that if hack the foodnstuff server, then you will steal 1% of its total money. If you hack using N threads, then you will steal N*0.01% of its total money.
```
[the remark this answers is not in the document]
TheCoderJT: I never changed these files manually by using npm run doc think you automatically changed files.

src/ScriptEditor/NetscriptDefinitions.d.ts
TheCoderJT: Yeah, this would be idea on the actual specific method pages and showing how to use said method but on the parent ns page thing my original woudl suffice. But, if the dev asks me to do more will

markdown/bitburner.ns.hackanalyze.md
TheCoderJT: I never changed these files manually by using npm run doc think you automatically changed files.

package-lock.json
TheCoderJT: I never changed these files manually by using npm run doc think you automatically changed files.

src/ScriptEditor/NetscriptDefinitions.d.ts
TheCoderJT: Umm, i might make some changed based on your feedback. But, i didn't really want to be too decriptive on the main ns page here https://github.com/danielyxie/bitburner/blob/dev/markdown/bitburner.ns.md I wanted to save your version for pages like property pages https://github.com/danielyxie/bitburner/blob/dev/markdown/bitburner.bladeburner.md and method pages https://github.com/danielyxie/bitburner/blob/dev/markdown/bitburner.bladeburner.getactionautolevel.md

TheCoderJT: I don't know what I should do if i do it my way its going to take ages if I do it your way and be more descriptive i might be able to get away with not adding example on the property pages or method pages

TheCoderJT: I will make slight changes using some of your suggestions, I still don't know if i should specific actually function names and properties in the examples. Anyways will update this pull request and see what you all think of the changes.

TheCoderJT (commented):
Once it's merged, this is what the main ns page will look like screenshot my screenshot removed the code background color just imagine thats there aswell sorry about that.

Issue #1922, opened 2021-12-15
cp doesnt work


hydroflame:
[home ~/]> cp scan-servers.js version2/list-servers.js
scp failed

PR #1932, opened 2021-12-15
Update cp and mv terminal commands #1922

theit8514:
Add support for relative file paths.
Fix bug with mv deleting files when file path is the same.
Correct cp error messages.

### DIFF
--- a/src/Terminal/commands/cp.ts
+++ b/src/Terminal/commands/cp.ts
@@ -3,6 +3,7 @@ import { IRouter } from "../../ui/Router";
 import { IPlayer } from "../../PersonObjects/IPlayer";
 import { BaseServer } from "../../Server/BaseServer";
 import { isScriptFilename } from "../../Script/isScriptFilename";
+import { getDestinationFilepath, areFilesEqual } from "../DirectoryHelpers";
 
 export function cp(
   terminal: ITerminal,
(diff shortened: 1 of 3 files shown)


alainbryden:
Good to paste a link so that git registers the linked issue: https://github.com/danielyxie/bitburner/issues/1922

src/Terminal/commands/cp.ts
```diff
@@ -50,7 +65,7 @@ export function cp(
 
       const tRes = server.writeToTextFile(dst, txtFile.text);
       if (!tRes.success) {
-        terminal.error("scp failed");
+        terminal.error("cp failed");
```
alainbryden, line 68: Hmm... this is a tricky situation, I think this code is shared by both CP and SCP commands:
image

So if you change it to one, it'll look wrong when someone used the other alias. Perhaps you should just change all errors to say "copy failed" and then it will make sense whether they used "cp" or "scp"?

alainbryden (approved):

src/Terminal/commands/cp.ts
theit8514: I believe scp is it's own command. https://github.com/danielyxie/bitburner/blob/dev/src/Terminal/commands/scp.ts

alainbryden: My bad, I could have sworn they shared the same implementation under the hood!

PR #1936, opened 2021-12-16
Issue1920

TheCoderJT:
👍

### DIFF
--- a/src/Terminal/commands/runProgram.ts
+++ b/src/Terminal/commands/runProgram.ts
@@ -20,7 +20,8 @@ export function runProgram(
   const programName = args[0] + "";
 
   if (!player.hasProgram(programName)) {
-    terminal.error("No such (exe, script, js, or ns) file on home computer (Only programs that exist on your home computer can be run)");
+    terminal.error(`No such (exe, script, js, or ns) file! (Only programs that exist on your home computer or scripts on ${player.getCurrentServer().hostname
+    } can be run)`);
     return;
   }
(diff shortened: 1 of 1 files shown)


src/Terminal/commands/runProgram.ts
```diff
@@ -20,7 +20,8 @@ export function runProgram(
   const programName = args[0] + "";
 
   if (!player.hasProgram(programName)) {
-    terminal.error("No such executable on home computer (Only programs that exist on your home computer can be run)");
+    terminal.error(`No such (exe, script, js, or ns) file! (Only programs that exist on your home computer or scripts on ${player.getCurrentServer().hostname
+    } can be run)`);
```
alainbryden, line 24: I think users can "run" coding contracts too right? So we'd need to add `cct` to the list.

Also, most will probably think of ".script / js / ns" files as "programs" so you might want to clarify the wording to "Only .exe files on your home computer, or scripts on ... can be run"

TheCoderJT: I disagree with "most will probably think of ".script / js / ns" files as "programs" the main issue hydoroflame stated "run asd.scrip without the t thinks it's a .exe, the message should be more generic." so i'm going to keep the generic specificity here. thanks for your input though

Hedrauta (approved):
i'ts fine, but i do understand alainbrydens comment.... they .js/.ns/.script - files are executables too 😉

Issue #1906, opened 2021-12-15
Terminal weaken should only print to 2 digit


hydroflame:

TheCoderJT:
https://github.com/danielyxie/bitburner/pull/1938

PR #1938, opened 2021-12-16
#1906

TheCoderJT:
😁 ✅

### DIFF
--- a/src/Terminal/Terminal.ts
+++ b/src/Terminal/Terminal.ts
@@ -254,9 +254,9 @@ export class Terminal implements ITerminal {
     const expGain = calculateHackingExpGain(server, player);
     server.weaken(CONSTANTS.ServerWeakenAmount);
     this.print(
-      `'${server.hostname}' security level weakened to ${server.hackDifficulty}. Gained ${numeralWrapper.formatExp(
-        expGain,
-      )} hacking exp.`,
+      `'${server.hostname}' security level weakened to ${server.hackDifficulty.toFixed(
+        3,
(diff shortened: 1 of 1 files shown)


src/Terminal/Terminal.ts
```diff
@@ -254,9 +254,9 @@ export class Terminal implements ITerminal {
     const expGain = calculateHackingExpGain(server, player);
     server.weaken(CONSTANTS.ServerWeakenAmount);
     this.print(
-      `'${server.hostname}' security level weakened to ${server.hackDifficulty}. Gained ${numeralWrapper.formatExp(
-        expGain,
-      )} hacking exp.`,
+      `'${server.hostname}' security level weakened to ${server.hackDifficulty.toFixed(
+        3,
+      )} and Gained ${numeralWrapper.formatExp(expGain)} hacking exp.`,
```
alainbryden, line 259:
```suggestion
    this.print(`'${server.hostname}' security level weakened to ${server.hackDifficulty.toFixed(3)} ` +
      `and Gained ${numeralWrapper.formatExp(expGain)} hacking exp.`,
```
I don't know why it was ever formatted the way it was, but perhaps something like this would be more liked all round?

TheCoderJT: Umm, the comma and the + are personal preference to be honest, they do the same thing really.

TheCoderJT: Thanks for the suggestion. But, I'm going to keep it how I committed it.

PR #1959, opened 2021-12-16
Update gettingstartedguideforbeginnerprogrammers.rst

cblte:
Updated the scan-analyze 2 output with current values. The server n00dles only has 4GB ram available, so I took him of the list of usable servers to run the script on.

### DIFF
--- a/doc/source/guidesandtips/gettingstartedguideforbeginnerprogrammers.rst
+++ b/doc/source/guidesandtips/gettingstartedguideforbeginnerprogrammers.rst
@@ -196,59 +196,63 @@ Here's what mine showed at the time I made this::
     [home ~]> scan-analyze 2
     ~~~~~~~~~~ Beginning scan-analyze ~~~~~~~~~~
 
-    >n00dles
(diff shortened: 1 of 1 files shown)


doc/source/guidesandtips/gettingstartedguideforbeginnerprogrammers.rst
```diff
@@ -196,59 +196,63 @@ Here's what mine showed at the time I made this::
...
-
-    ---->CSEC
-    ------Root Access: NO, Required hacking skill: 54
-    ------Number of open ports required to NUKE: 1
-    ------RAM: 8
+    --RAM: 32.00GB
 
 Take note of the following servers:
 
-* |n00dles|
```
MageKing17, line 251: The references to `n00dles` from here on down shouldn't be removed, they should be replaced with `foodnstuff`.

cblte: This is also an option which can be considered. Good point.

PR #1962, opened 2021-12-16
Attempt to fix processes that stay up after closing app

MartinFournier:
These calls seem redundant but since I added them I did not see the process stay up after closing the app (Windows 10)...


hydroflame:
That won't work with "reload n kill all"

MartinFournier:
I think this way works for the hard reload.
Feel free to discard the pull request if it's too hacky.

hydroflame:
I just pushed something that MIGHT work. I didn't know about process.exit, that might be the key

Please leave yours open in case my attempt does not work.

PR #1984, opened 2021-12-17
feat: Add vim mode to script editor

billyvg:
This adds an option to turn on vim mode using the [`monaco-vim`](https://github.com/brijeshb42/monaco-vim) library

https://user-images.githubusercontent.com/79684/146585397-a4cd559b-fb4b-45e0-b538-dc5c05f1a4c9.mov

.

PR #1986, opened 2021-12-17
feat: add `vim` terminal command

billyvg:
This requires https://github.com/danielyxie/bitburner/pull/1984, but adds a `vim` term command.


hydroflame:
Same problem here

billyvg:
OK, cleaned up some stuff

PR #1993, opened 2021-12-18
Fix problems finding root files in cat and ns.read

Nick-Colclasure:
Cat ends up translating the path it receives from relative to absolute twice, which I fix by changing the filename to an absolute path before it's passed to getTextFile with a leading "/" so that it doesn't interpret the filename as being relative.

Read I fixed by causing getTextFile to remove the leading "/" from files that are in the root directory, since that is required to translate their name into the native "filesystem"s format.

The cat changes can be tested via the following:
`
nano test_root_file.txt
test root file text
ctrl-b
nano test_directory/test_dir_file.txt

ctrl-b
cd test_directory
cat test_root_file.txt
`
And the ns.read() change via the attached file:
[ns_read_bugfix_test.zip](https://github.com/danielyxie/bitburner/files/7738748/ns_read_bugfix_test.zip)

### DIFF
--- a/src/Terminal/DirectoryHelpers.ts
+++ b/src/Terminal/DirectoryHelpers.ts
@@ -185,7 +185,7 @@ export function getDestinationFilepath(destination: string, source: string, cwd:
     return destination;
   } else {
     // Append the filename to the directory provided.
-    let t_path = removeTrailingSlash(dstDir);
+    const t_path = removeTrailingSlash(dstDir);
     const fileName = getFileName(source);
     return t_path + "/" + fileName;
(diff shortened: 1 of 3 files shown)


hydroflame:
That causes a new problem.

`nano dir/test.txt`
write whatever, save
`cd dir`
`cat test.txt`
ERROR

Nick-Colclasure:
Alright, this should fix that issue. I tested cat of a file in a subdir from root, from the subdir, and a cat of a file with the same name (and different contents) in root from both root and the subdir.

PR #1996, opened 2021-12-18
Re-add --no-scripts launch option for electron

MartinFournier:
It must have been overwritten during the merge with the previous fix.

If possible, it may be nice to check in the app's settings if there's a way to get a popup on launch with the option to run with this flag. Something like this:

steam_yTWIrxbhLm

Where the second option would have the launch option already set for you.


hydroflame:
Oh I didn't even notice, but I did all that this morning.

PR #2003, opened 2021-12-18
Add easily switchable predefined themes

MartinFournier:
Includes the default theme and a Monokai-ish variant

We just have to add new entries to the Themes.ts file, we could pick the most popular community ones.

kgrl2dSvrv

### DIFF
--- a/src/ui/React/ThemeEditorModal.tsx
+++ b/src/ui/React/ThemeEditorModal.tsx
@@ -7,9 +7,11 @@ import Paper from "@mui/material/Paper";
 import TextField from "@mui/material/TextField";
 import IconButton from "@mui/material/IconButton";
 import ReplyIcon from "@mui/icons-material/Reply";
+import PaletteSharpIcon from "@mui/icons-material/PaletteSharp";
 import { Color, ColorPicker } from "material-ui-color";
 import { ThemeEvents } from "./Theme";
 import { Settings, defaultSettings } from "../../Settings/Settings";
+import { ITheme, getPredefinedThemes } from "../../Settings/Themes";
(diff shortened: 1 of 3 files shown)


src/ui/React/ThemeEditorModal.tsx
```diff
@@ -96,250 +105,262 @@ export function ThemeEditorModal(props: IProps): React.ReactElement {
...
+          sx={{ mb: 1 }}
+          multiline
+          fullWidth
+          maxRows={3}
+          label={"import / export theme"}
+          value={JSON.stringify(customTheme)}
+          onChange={onThemeChange}
+        />
+          <>
+            <Typography sx={{ my: 1 }}>Copy the string above if you want to backup or share your theme with others.</Typography>
```
millennIumAMbiguity, line 359: Text is too hard to read, would suggest the following as a replacement:
*Backup your theme or share it with others by copying the string above.*

src/ui/React/ThemeEditorModal.tsx
```diff
@@ -96,250 +105,262 @@ export function ThemeEditorModal(props: IProps): React.ReactElement {
...
+            <Typography sx={{ my: 1 }}>Use the buttons below to replace the current theme with a pre-built template</Typography>
```
millennIumAMbiguity, line 360: Missing dot. Additionally, it's a bit hard to read. see following suggestion:
*Replace the current theme with a pre-built template using the buttons below.*

millennIumAMbiguity (changes requested):
The texts are a bit long and hard to read, see comments.

Issue #2042, opened 2021-12-19
duplicate .txt


Impetus2708:
creating a .txt and using ls in the terminal shows the created txt both in green and yellow. Using rm on the .txt removes the green one, the yellow one stays. You can still autocomplete with that txt after using rm, but it will state No such file exists.

Rez855:
Fixed in #2044

Rez855:
fixed #2044

PR #2044, opened 2021-12-19
Script Editor Refactor

Rez855:
Refactored Script Editor
When closing a tab it will now ask you if you want to save any changes made
Added back ctrl + b shortcut to save and close editor
Fixed a bug where creating a text file would create 2 of them
Added new command to remove broken txt files, "rm [TextFileName] -tas"
Scripts being edited will not lose data if the game were to crash
When closing a tab it will open the closest tab instead of the first tab
Fixed css styling for PromptEvent
Script Editor will no longer open a temporary file when no other files are open


Rez855:
fixed #2042

PR #2049, opened 2021-12-19
Issue #2048 fix

Kejikus:
Issue #2048 fix.
Not tested, based on fix in devtools in running game.

Although, I would suggest applying background-primary color to \<html\> element directly.
Or to the \<body\>, but remove its padding, because there are not-so-nice gaps, as shown below:
Снимок экрана 2021-12-20 в 00 16 45


Kejikus:
#2048

PR #2075, opened 2021-12-20
feat: Add pre-commit hook to lint and format 

billyvg:
This adds `husky` and `lint-staged` in order to run lint+fixing and formatting when making a commit. This should keep the code a bit tidier.


hydroflame:
Ok so because I don't know how this tech works (and it's not critical) I'm just going to close this.

PR #2076, opened 2021-12-20
meta: Format all markdown files with prettier

billyvg:
Alternatively, you could just mark these files as ignored w/ prettier.


hydroflame:
Well those are autogenerated so I don't really care if they're prettified or not.

Issue #2027, opened 2021-12-19
Various documentation updates


hydroflame:
getHackTime(), getWeakenTime(), getGrowTime() 
Influenced by: currentSecurityLevel of host, Hacking Level of player, ...?

hackAnalyze(), growthAnalyze()
Influenced by: currentSecurityLevel of host, ...?

grow(), hack(), weaken()
The percentage is calculated when the method finishes executing. security level is checked once at the start (to determine how long the operation takes) and once at the end (to determine its effect) (from MAGEKING)

hexnaught:
> grow(), hack(), weaken()
> The percentage is calculated when the method finishes executing. security level is checked once at the start (to determine how > long the operation takes) and once at the end (to determine its effect) (from MAGEKING)

Is it worth also clarifying in the docs/method dockblocks that by default, these commands use all available script threads as per `-t <number>`. (_I quickly check `ns.hack()` in game and didn't see this included_).

As per discord:

> hydroflame  — 16/12/2021:
>   No one needs to set the optional argument of hack btw
>   Specially not this early in the game
>   hack(hostname) will use all possible threads the script was started with.

phyzical:
anything else still missing?

PR #2104, opened 2021-12-22
Documentation updates

mikomyazaki:
Addresses #2027 

I think the `grow()` documentation is already clear enough about when things are calculated?


hydroflame:
Those files are autogenerated by NetscriptDefinitions.d.ts

Those changes cannot be accepted as they will be overwritten.

PR #2105, opened 2021-12-22
Update bitburner.ns.getweakentime.md

spdkils:
Should be self explanatory...

Function is for weaken, however documentation listed grow.


hydroflame:
Cannot accept those changes as `NetscriptDefinitions.d.ts` autogenerated that file, please move those changes to NetscritpDefinitions.d.ts

Issue #1944, opened 2021-12-16
ls -l


hydroflame:
https://steamcommunity.com/app/1812820/discussions/2/5545618081292766987/

mikkerlo:
Maybe ```ls -1``` then? It does exactly what they want in Linux and doesn't contain useless info such as creation date, e.t.c

phyzical:
doing this in two parts first pr is up to handle the --grep logic change in ls.

PR #2139, opened 2021-12-22
Implements 'ls -l' for single-column output

mikomyazaki:
Addresses #1944 

ls_list_output


hydroflame:
uhhh, not quite, the codebase already has a flag parser that we can use to simplify a lot of this code.

mikomyazaki:
Converted to use the flags parser.

There is another command `ps()` which uses the `| grep pattern` format for its arguments, do you want me to convert it to `--grep pattern` for consistency, in this PR? Or do it separately.

hydroflame:
I want it converted to `--grep` because it's easier for the game to handle. Even though it does not make "real world" sense.

hydroflame:
Ok so this isn't really using the flags library as intended, `-l` doesn't work. So I'm going to close this.

PR #2144, opened 2021-12-23
Adding support for ctrl + arrow key navigation around open files in script-editor

PhilipArmstead:
This listens for the keyup event instead of keypress or keydown because, I believe, Monaco is swallowing those events when pressing buttons that don't enter printable characters

### DIFF
--- a/src/ScriptEditor/ui/ScriptEditorRoot.tsx
+++ b/src/ScriptEditor/ui/ScriptEditorRoot.tsx
@@ -219,6 +219,31 @@ export function Root(props: IProps): React.ReactElement {
     };
   }, [options, editorRef, editor, vimEditor]);
 
+  useEffect(() => {
+    function changeActiveTab(event: KeyboardEvent): void {
+      if (Settings.DisableHotkeys) return;
+
+      if ((event.ctrlKey || event.metaKey) && ['ArrowLeft', 'ArrowRight'].includes(event.key)) {
+        event.preventDefault();
+        event.stopPropagation();
+
+        const currentTabIndex = getCurrentTabIndex()
+        let nextTab
+
+        if (event.key == 'ArrowLeft') {
+          nextTab = currentTabIndex < 1 ? openScripts.length - 1 : currentTabIndex - 1
+        } else {
+          nextTab = currentTabIndex >= openScripts.length - 1? 0 : currentTabIndex + 1
+        }
+
+        onTabClick(nextTab)
+      }
+    }
+
+    document.addEventListener("keyup", changeActiveTab);
+    return () => document.removeEventListener("keyup", changeActiveTab);
+  });
+
   // Generates a new model for the script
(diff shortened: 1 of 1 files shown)


aaronransley:
@PhilipArmstead I use ctrl+arrows to jump to word + special character stops. Would it be possible to use alt+arrows for tab switching instead? It seems those are currently unbound by Monaco.

Hedrauta (approved):
👍  ... it does even save the line and row i was in...

Hedrauta:
> @PhilipArmstead I use ctrl+arrows to jump to word + special character stops. Would it be possible to use alt+arrows for tab switching instead? It seems those are currently unbound by Monaco.

oh.... 🤔  never using this feature, bc i do program my scripts in vscode 😆

Hedrauta (changes requested):
🤔 rethinking the approval and maybe just add Alt as additional Button... so Ctrl + ArrowLeft/Right can still be used to jump in the line 😉

PhilipArmstead:
I've added support for the Alt key too but that's not going to work very well in browsers (it will navigate forwards/backwards through history if available; at least Firefox does.)

src/ScriptEditor/ui/ScriptEditorRoot.tsx
```diff
@@ -223,7 +223,7 @@ export function Root(props: IProps): React.ReactElement {
     function changeActiveTab(event: KeyboardEvent): void {
       if (Settings.DisableHotkeys) return;
 
-      if ((event.ctrlKey || event.metaKey) && ['ArrowLeft', 'ArrowRight'].includes(event.key)) {
+      if ((event.ctrlKey || event.metaKey || event.altKey) && ['ArrowLeft', 'ArrowRight'].includes(event.key)) {
```
Hedrauta, line 226: i meant  (Ctrl/Cmd) + Alt + arrowkey
```suggestion
      if (((event.ctrlKey || event.metaKey) && event.altKey) && ['ArrowLeft', 'ArrowRight'].includes(event.key)) {
```
ps: tested.... works great ;)

Hedrauta:
How it looks:

gif (6)

Hedrauta (approved):

apixdae:
It may be too late to add this to the discussion, but the standard keyboard shortcut for switching tabs in a web browser, VSCode, etc is `Ctrl+Tab` to move forward and `Ctrl+Shift+Tab` to move backward (although VSCode handles it slightly differently, moving between tabs based on which were most recently visited, as opposed to position on the tab bar). It would be useful to be able to use the common shortcuts to switch tabs as opposed to new ones like `Ctrl+Alt+Left/Right`.

Hedrauta:
That may work for the steam version


But there are still people on the domain  one, which would interfere with the Browser's shortcut

hydroflame:
That's kindof the problem. Whatever shortcut we use will cause issues for one OS or another :\

PR #2169, opened 2021-12-24
Feature/add grep option to ps

maxtimum:
added the option to pass in a regular expression to match running process names against in ps

| use case | example |
|-|-|
|show all running processes (unchanged)|image|
|show all processes staring with `a`|image|
|show all processes starting with `a[^n] \| b`|image|

|error case| example |
|-|-|
|missing `grep` |image|
|missing `\|`|image|
|missing both|image|

### DIFF
--- a/doc/source/basicgameplay/terminal.rst
+++ b/doc/source/basicgameplay/terminal.rst
@@ -432,10 +432,10 @@ empty file will be created.
 ps
 ^^
 
-    $ ps [-g pattern]
+    $ ps [-g, --grep pattern]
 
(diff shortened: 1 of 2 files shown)


hydroflame:
Ok, I'll accept this if you use flags instead of pipe, since this game does not support pipes.

maxtimum:
> Ok, I'll accept this if you use flags instead of pipe, since this game does not support pipes.

Sure, I modeled this off of the | grep option for ls so I figured it was a flavor thing. I'll change to -g <pattern>

PR #2172, opened 2021-12-24
Cache the blobs generated by scripts

theit8514:
When running a script, generate a hash of the code used by that script. If that hash is in our blob cache, we reuse the blob URL. If not, a new blob is generated and stored in the cache.

Some notes:
* This still has the issue that scripts are recompiled on soft reset, but the hash of the script should not change, so the blobs and ScriptUrls will still be the same.
* We are no longer calling URL.revokeObjectURL on URLs since they could be in use by another script. This will leave blob URLs leaked on occasion, but that leaking should be at a minimum with the cache. As far as I can tell, URL.revokeObjectURL doesn't remove the URL from the DevTools anyways, so those URLs are in memory somewhere.
  * Example, this only generates 4 blobs total (1 for the main script, 1 for helpers.js, and one each for the flip/flop):
   ```js
    import {
        getNsDataThroughFile
    } from './helpers.js'
    
    /** @param {NS} ns */
    async function getCurrentFactionFavour(ns, factionName) {
        return await getNsDataThroughFile(ns, `ns.getFactionFavor('${factionName}')`, '/Temp/faction-favor.txt');
    }
    
    /** @param {NS} ns **/
        while (true) {
            let currentFavor = await getCurrentFactionFavour(ns, "Bachman & Associates");
            ns.tprint(currentFavor);
            currentFavor = await getCurrentFactionFavour(ns, "Sector-12");
            ns.tprint(currentFavor);
            await ns.sleep(5000);
        }
    }
    ```
* Scripts running across multiple servers or multiple filenames with the same hash will have the same blobs, so this necessitated a change with how we lookup blob urls to script name mappings.
* A 3rd party library is imported for sha256 hashing since the browser based hashing method is async and we needed synchronous calculation of the hash in several places.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -34,6 +34,7 @@
         "file-saver": "^1.3.8",
         "fs": "^0.0.1-security",
         "jquery": "^3.5.0",
+        "js-sha256": "^0.9.0",
         "jszip": "^3.7.0",
         "material-ui-color": "^1.2.0",
         "monaco-editor": "^0.27.0",
@@ -13698,6 +13699,11 @@
       "version": "3.6.0",
       "resolved": "https://registry.npmjs.org/jquery/-/jquery-3.6.0.tgz",
       "integrity": "sha512-JVzAR/AjBvVt2BmYhxRCSYysDsPcssdmTFnzyLEts9qNwmjmu4JTAMYubEfwVOSwpQ1I1sKKFcxhZCI2buerfw=="
+    },
+    "node_modules/js-sha256": {
+      "version": "0.9.0",
+      "resolved": "https://registry.npmjs.org/js-sha256/-/js-sha256-0.9.0.tgz",
+      "integrity": "sha512-sga3MHh9sgQN2+pJ9VYZ+1LPwXOxuBJBA5nrR5/ofPfuiJBE2hnjsaN8se8JznOmGLN2p49Pe5U/ttafcs/apA=="
     },
     "node_modules/js-tokens": {
       "version": "4.0.0",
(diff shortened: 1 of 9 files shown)


smolgumball:
@theit8514 I am running into a stale dependency issue where if an imported script is updated, the parent doesn't pick up on the changes. Internally, the `_.hash` property of the `Script` object is updating, but whatever mechanism is used for caching the blobs is causing the old blob to be reused rather than the newly regenerated one.

Really not sure how best to package my findings for you to reproduce, but I can try if needed, just let me know!

paulcdejean:
@hydroflame can this maybe be reverted while a better implementation is worked on? This has been causing headaches to basically anyone who tries to play without using a rollup process...

https://github.com/bitburner-official/bitburner-vscode/issues/10
https://github.com/danielyxie/bitburner/issues/2413
https://github.com/danielyxie/bitburner/issues/2336
https://github.com/danielyxie/bitburner/issues/2408
https://github.com/danielyxie/bitburner/issues/2443

I've been needing to do this workaround, modifying the bundle file every time there's an update:

> It seems like you can disable to import cache check by editing the main bundle ( \Bitburner\resources\app\main.bundle.js) and replacing
`l||(l=_getScriptUrls(s,t,a))`

> with
`(l=_getScriptUrls(s,t,a))`

paulcdejean:
Basically I think the performance gains here benefit a small number of players while the buggy behavior caused by the incomplete implementation is affecting a lot of people.

tvanderpol:
I'd at least really appreciate a toggle - I'd be very happy to eat a performance impact if that meant I was always running the code I thought I was running.

Domanoc:
I would even go for an option to have a netscript command to clear the cache of a specific script/blob then we could manually clear chained imports while scripting. Without having to go in roundabout ways of updating or reloading the game.

PR #2183, opened 2021-12-25
Get rid of console spam in NetscriptFunctions.ts

MageKing17:
This was filling up the console and making it hard to debug anything other than `exec` calls.


Hedrauta (approved):
🤣  after 400k entries, it's kind of  getting annoying ...

ErzengelLichtes:
I would suggest changing it to a `workerScript.log` or something like that so that `ns.disableLog("exec")` could be used for people who don't want the log spam, but is still there for people who need it.

PR #2184, opened 2021-12-25
Suggestion: Change Ram-Formatting to IEC 60027-2

Hedrauta:
Just an idea <.<
pic

### DIFF
--- a/src/ui/numeralFormat.ts
+++ b/src/ui/numeralFormat.ts
@@ -110,11 +110,11 @@ class NumeralFormatter {
   }
 
   formatRAM(n: number): string {
-    if (n < 1e3) return this.format(n, "0.00") + "GB";
-    if (n < 1e6) return this.format(n / 1e3, "0.00") + "TB";
-    if (n < 1e9) return this.format(n / 1e6, "0.00") + "PB";
-    if (n < 1e12) return this.format(n / 1e9, "0.00") + "EB";
-    return this.format(n, "0.00") + "GB";
+    if (n < Math.pow(2,10)) return this.format(n, "0.00") + " GiB";
+    if (n < Math.pow(2,20)) return this.format(n / Math.pow(2,10), "0.00") + " TiB";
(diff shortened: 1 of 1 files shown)


MageKing17:
RAM can't be measured in GiB, because then functions would cost fractional bytes (0.05 GiB is 53687091.2 bytes).

Hedrauta:
🤔  what do you mean? it's still fine.... just the suffix will change
pic

Hedrauta:
oh... !
no... everything starts at GiB... just the counting up to TiB, PiB & EiB will change

Hedrauta:
or... to be more clear: there is no "MiB" or "KiB" used in the game 😉 otherwise, i would have respect that one ;)

MageKing17:
> 🤔 what do you mean? it's still fine.... just the suffix will change

The point is that the suffix wouldn't make *sense*: 1.60 GiB is 1.6 * 1024 * 1024 * 1024 bytes, or 1717986918.4 bytes, and ".4 bytes" is a nonsensical value (it would be 3.2 bits). Whereas 1.60 GB is 1.6 * 1000 * 1000 * 1000 bytes, and one billion six hundred million bytes is a perfectly valid value.

Hedrauta:
> The point is that the suffix wouldn't make _sense_: 1.60 GiB is 1.6 * 1024 * 1024 * 1024 bytes, or 1717986918.4 bytes, and ".4 bytes" is a nonsensical value (it would be 3.2 bits). Whereas 1.60 GB is 1.6 * 1000 * 1000 * 1000 bytes, and one billion six hundred million bytes is a perfectly valid value.




> or... to be more clear: there is no "MiB" or "KiB" used in the game 😉 otherwise, i would have respect that one ;)

it's just for the eyes..... or you want get shown 1.04PB servers ? 
🤔 ... maybe only change the formatting AFTER the GB, so it would be GB, TiB, PiB and EiB .... then, everything would be fine "in your opinion"... just the math above GB would change....

PR #2217, opened 2021-12-28
Refactor NetscriptPort class and add to intellisense

ErzengelLichtes:
Added documentation for the class returned by getPortHandle into the intellisense. This required refactoring the IPort interface to be an actual interface of methods instead of an object of functions.
Refactored NetscriptPort and IPort to be a normal class and interface, using Javascript's Hash-name members to provide a true private field instead of the local-shared "private" that was used by the NetscriptPort function.


hydroflame:
Ok, I understand why you did this. However it USED to be like this and if you write Port like that it will not work properly in NS1. 

It might be possible to leave the implementation like that and just add the documentation though.

ErzengelLichtes:
getPortHandle says in its documentation that it doesn't work in NS1, though? 
This was already in the jsdoc:
`WARNING: Port Handles only work in NetscriptJS (Netscript 2.0). They will not work in Netscript 1.0`
I assumed because of that remark that it wouldn't need to work in NS1. I tested the NS methods (writePort, etc) that they work in NS1.
If it is necessary that it works in NS1, I can create a wrapper class that's returned by getPortHandlerNS1 or something like that.

ErzengelLichtes:
I added a system for Netscript 1 to have prefixed methods to tell it to use that method instead of a Netscript 2 method. It looks for the prefix "ns1_" and, if it finds it, it uses that method (after stripping the prefix from the name) instead of the one without the prefix (which is considered to be Netscript 2 specific).

There is also a function in NetscriptFunctions that creates a wrapper method that will call the Netscript 2 method, take its return value and turn it into an object Netscript 1 can understand, and return that, then takes that method and puts it into the NS object as a prefixed method.

This way all the features of Netscript 2 will be available in development while still being able to provide the return value to Netscript 1.

hydroflame:
Included changes to documentatiaon separately.

PR #2221, opened 2021-12-29
Won't let you take a course when your balance is 0 or less. Will cancel your course if you have 0 or less.

### LINKED ISSUE #2205: Courses let you go into negative balance
Started a course, left it running overnight apparently.
Came back to find myself in 59m debt
Also seems you can start courses while being in negative balance.

image

K3v1nNZ:
Starting a course with a balance equal to $0 or less will make a dialog box saying "You don't have enough money to take a course!".
If you are in the middle of a course and reach $0 or less, It will cancel the course.
This should fix #2205


hydroflame:
No this would make it impossible to Create Program with negative money.

PR #2226, opened 2021-12-29
don't return to terminal after saving a script (fixes #2223)

### LINKED ISSUE #2223: Is it possible to save scripts without closing editor ?
Hi!
First, thanks for your game, I'm really enjoying it!

There is just a small issue for me, but maybe it's me who haven't found the magic setting. 
I would like to be able to save without closing the editor every time, is it possible ?

I'm a developer myself, and hitting CTRL+S is just a reflex, but every time it closes the script editor. I could be used to that, but the worst part is that is disable possibles undo.
For example, I deleted a portion of my code by inadvertance, CTRL+S and run the script, when the error popup made my mistake obvious, I reopened the script, but I wasn't able to undo to retrieve my missing piece of code.

jasonhaxstuff:
if the editor didn't close every time i saved i would feel more compelled to write long complicated scripts (fixes #2223)

### DIFF
--- a/src/ScriptEditor/ui/ScriptEditorRoot.tsx
+++ b/src/ScriptEditor/ui/ScriptEditorRoot.tsx
@@ -431,7 +431,6 @@ export function Root(props: IProps): React.ReactElement {
             server.scripts,
           );
           if (Settings.SaveGameOnFileSave) saveObject.saveGame();
-          props.router.toTerminal();
           return;
         }
       }


ErzengelLichtes:
Would it be possible to make it so ctrl+s leaves you on the editor and ctrl+b sends you back to the terminal?

hydroflame:
Already in dev

PR #2235, opened 2021-12-30
Update bitburner.ns.hackanalyze.md

simon-bertlin:


theit8514 (changes requested):
Please modify the src/ScriptEditor/NetscriptDefinitions.d.ts file instead of the markdown files directly.

PR #2243, opened 2021-12-31
Add Function Call signatures to ns.kill so typescript is able to call ns.kill correctly

TheMas3212:
previously this would fail to compile due to not providing a host argument
`
const pid: number = 42;
ns.kill(pid)
`


TheMas3212:
Added extra call signatures for getScriptIncome and getScriptExpGain so that typescript will work with the no argument behaviour of them

Issue #2297, opened 2022-01-03
Cannot print import-like string for current file


DyspC:
It looks like something tries to replace the import text recursively and crashes

```
RUNTIME ERROR
recurseCrash.js@home

Maximum call stack size exceeded
stack:
RangeError: Maximum call stack size exceeded
at String.replace ()
at _getScriptUrls (file:///D:/SteamLibrary/steamapps/common/Bitburner/resources/app/main.bundle.js:19:83552)
at file:///D:/SteamLibrary/steamapps/common/Bitburner/resources/app/main.bundle.js:19:83907
at String.replace ()
at _getScriptUrls (file:///D:/SteamLibrary/steamapps/common/Bitburner/resources/app/main.bundle.js:19:83552)
at file:///D:/SteamLibrary/steamapps/common/Bitburner/resources/app/main.bundle.js:19:83907
at String.replace ()
at _getScriptUrls (file:///D:/SteamLibrary/steamapps/common/Bitburner/resources/app/main.bundle.js:19:83552)
at file:///D:/SteamLibrary/steamapps/common/Bitburner/resources/app/main.bundle.js:19:83907
at String.replace ()
```

- recurseCrash.js
```js
/** @param {NS} ns **/
export async function main(ns) {
    ns.tprintf('import * as lib from "./recurseCrash.js"')
}
```

theit8514:
This is a known limitation of the import system. It also parses comments so beware of that. The permanent fix is to replace the regex replace with the acorn AST parser.

As a workaround you can split the string inside the word 'from' e.g. `ns.tprintf('import * as lib fr' + 'om "./recurseCrash.js"')`.

PR #2310, opened 2022-01-03
Replace regex import with acorn AST parser

theit8514:
I separated this change out from the potential save breaking change of fixing relative directories. This change will still import files from the same location, but will resolve issues with trying to put imports in comments and quoted strings.

This will resolve https://github.com/danielyxie/bitburner/issues/2297

Tested with three styles of import:
```js
import * as helpers from '/helpers.js'
// import * as self from '/test-import1.js'

/** @param {NS} ns **/
	ns.tprint("import * as self from '/test-import1.js'");
	await helpers.runCommand(ns, 'ns.alert("test")');
}
```
```js
import { runCommand } from '/helpers.js'
// import { main as selfMain } from '/test-import2.js'

/** @param {NS} ns **/
	ns.tprint("import { main as selfMain } from '/test-import2.js'");
	await runCommand(ns, 'ns.alert("test")');
}
```
```js
import { runCommand as newCommand } from '/helpers.js'
// import { main as selfMain } from '/test-import3.js'

/** @param {NS} ns **/
	ns.tprint("import { main as selfMain } from '/test-import3.js'");
	await newCommand(ns, 'ns.alert("test")');
}
```

### DIFF
--- a/src/NetscriptJSEvaluator.ts
+++ b/src/NetscriptJSEvaluator.ts
@@ -1,3 +1,10 @@
+/**
+ * Uses the acorn.js library to parse a script's code into an AST and
+ * recursively walk through that AST to replace import urls with blobs
+ */
+import * as walk from "acorn-walk";
+import { parse } from "acorn";
+
 import { makeRuntimeRejectMsg } from "./NetscriptEvaluator";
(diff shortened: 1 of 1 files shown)


theit8514:
Rebased and resolved issue with './' prefix on imports.

Issue #2276, opened 2022-01-01
ns.purchaseServer() doesn't properly validate RAM amount


apricotorange:
```
export async function main(ns) {
  ns.purchaseServer("darkweb", 2-Math.pow(2, 33));
}
```

This purchases a server with negative RAM.  It costs $0.  As far as I can tell, you can't do anything interesting with that server; it doesn't have enough RAM to run any scripts.

PR #2317, opened 2022-01-03
Validate RAM amount in ns.purchaseServer()

nickofolas:
Validates that the RAM amount specified is a positive integer for `ns.purchaseServer()` and `ns.getPurchasedServerCost()`.

Addresses  #2276


hydroflame:
That's actually not the problem and the issue you linked has already been fixed. However your changes are still beneficial so I'm merging them.

Issue #2067, opened 2021-12-20
Log windows can be lost on resize


Icehawk78:
If you open a log window at full width, and place it in the right half of the screen, then resize the game window to be half your screen width, the log window is inaccessible until you resize the game again.

Proposed solution: Active Scripts -> Log should close the open logging window if it's currently open, rather than just trying to open it, so that if you click that twice, it re-opens it in the current window width.

juliandie:
The top corner can disappear in the top and can't be reached again. Therefore it can't be closed.
The log window borders shall be clamped to the window size.

Or the log-window might be attached to the related editor-tab (like the inspect feature in browsers) and a drop-down to select a running instance.

phyzical:
Can confirm this is fixed and closable
![Animation2](https://user-images.githubusercontent.com/5182053/152819491-bc772bd9-b8f9-4e39-bbc4-7bedc7678ac5.gif)

PR #2319, opened 2022-01-04
Restrict log window to visible window

nickofolas:
Sets the bounds of log windows so that they can't be lost outside the visible window space.

Addresses #2067


hydroflame:
Unfortunately I have to  reject because it prevent ANY part of the window from going outside, whihc is going to make even more people angry.

We need a solution that lets part of it outside :\

PR #2334, opened 2022-01-04
Fix #2330 typo

### LINKED ISSUE #2330: Typo in ns.getWeakenTime() description
image

Should propably say "weaken" instead of "grow".

woody-lam-cwl:
Fix #2330 by modifying `bitburner.d.ts`


TheMas3212:
the file `bitburner.d.ts` is actually generated automatically by the build process, what you actually want to edit is `src/ScriptEditor/NetscriptDefinitions.d.ts`

PR #2342, opened 2022-01-05
Fix #2275 accepts NaN donation bug

### LINKED ISSUE #2275: ns.donateToFaction() accepts donation of NaN
```
  ns.donateToFaction("Daedalus", NaN);
}
```

This sets your reputation with that faction to NaN; this displays as zero, but it becomes impossible to gain reputation with that faction.  I didn't look deeply at all the possible effects.

woody-lam-cwl:
Fix #2275 by adding `isNaN` check


woody-lam-cwl:
Ran workflow on branch individually and all tests passed - should be ready to merge once dev is fixed

PR #2345, opened 2022-01-05
Fix #2278

### LINKED ISSUE #2278: Tooltip for "cache level" on hacknet servers is wrong for for 5x purchase.
The tooltip shown hovering over the purchase button always shows the result of a 1x purchase, e.g. "+64.000 hashes", instead of the actual increase.

woody-lam-cwl:
Fix #2278 modifying `HacknetServerElem.tsx`


woody-lam-cwl:
Ran workflow on branch individually and all tests passed - should be ready to merge once dev is fixed

PR #2364, opened 2022-01-05
Add 'host' config for electron api server

MartinFournier:
Only tweakable through the config, this can allow external
access to the server (ie - if running through a VM or WSL)

### DIFF
--- a/electron/api-server.js
+++ b/electron/api-server.js
@@ -69,12 +69,13 @@ function enable() {
   }
 
   const port = config.get('port', 9990);
-  log.log(`Starting http server on port ${port}`);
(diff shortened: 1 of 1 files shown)


anstosa:
Hi @MartinFournier how you use this? I'm running in WSL so I'm pretty sure I just have to change the host value to `0.0.0.0`. I see you've made it configurable but where do I set this flag?

Thanks!

anstosa:
I think I found the right config (`AppData/Roaming/bitburner/config.json`). The VSCode extension is still trying to access `127.0.0.1` (I think it has to be `localhost`?). Are you using that to sync or something else?

PR #2385, opened 2022-01-06
Fix script arguments to current property name

pakoito:
2022-01-06 03_02_49-Bitburner — Mozilla Firefox


pakoito:
Huh, `ns.args` exists. Nevermind.

PR #2398, opened 2022-01-06
Remove NS.write() method mentioning ports

Kebap:
With writePort() now a thing, this is no longer needed.


hydroflame:
Need to change NetscriptDefinitions.d.ts as that file is autogenerated

PR #2412, opened 2022-01-06
format bulletin points in createProgram.md

Tsunder:


hydroflame:
Won't work, need to change NetscriptDefinitions.d.ts

PR #2439, opened 2022-01-07
Remove duplicate entry from documentation

tthbalazs:
Noticed that `getHacknetMultipliers()` had a duplicate entry. This PR removes it.


hydroflame:
That's an autogenerated file

PR #2445, opened 2022-01-07
Add singularity RAM cost info to typings

MartinFournier:
Added tag %SINGULARITY_MULTIPLER% in the typing files, replaced by a value before being added to Monaco.

```sh
   * @remarks
   * RAM cost: 2 GB * %SINGULARY_MULTIPLIER%
   *
   * This function will automatically set you to start taking a course at a university.
```

firefox_FBMNtiY7Vr


Narnach:
This looks useful, thanks!

hydroflame:
Unfortunately I have to reject because markdown doc doesn't handle it.

I think it's better to just byte the bullet and write 16/4/1 manually

PR #2454, opened 2022-01-08
Apply bitnode multiplier before capping at 1 in calculatePercentMoneyHacked in hacking.ts

zeddrak:
It is entirely possible to hack more than the BN multiplier, which means this function is reporting the effects of the multiplier incorrectly. By moving the BN multiplier to before the caps are applied, the functionality is corrected.

(ie, hack % * BN multiplier, THEN cap to 1; not cap to 1, then BN multiply the 1)

TESTED: I'm currently running in BN 5 ( * BitNodeMultipliers.ScriptHackMoney = 0.15), and sent a 4b thread hack against several fully grown servers (Avail Money = 100% of Max Money). Each went down to 0% (not 85% as the current implementation would suggest) after the hack hit.

I've also tested and confirmed the proposed change's functionality in my current game through my custom formulas, but didn't do a full build or anything.

### DIFF
--- a/src/Hacking.ts
+++ b/src/Hacking.ts
@@ -49,7 +49,7 @@ export function calculatePercentMoneyHacked(server: Server, player: IPlayer): nu
 
   const difficultyMult = (100 - server.hackDifficulty) / 100;
   const skillMult = (player.hacking - (server.requiredHackingSkill - 1)) / player.hacking;
-  const percentMoneyHacked = (difficultyMult * skillMult * player.hacking_money_mult) / balanceFactor;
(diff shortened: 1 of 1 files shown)


hydroflame:
I **think** that's correct but I hope it doesn't break anyones script.

zeddrak:
It actually fixed many of my scripts, since hackAnalyze started returning the correct values. :)
(for example, if 1 thread would've hacked 120% in BN1, it'd report it as 100% in BN 5, but was actually only hacking 18%)

hydroflame:
Might fix yours and break others :)

zeddrak:
Yeah, if they were somehow manually compensating, the correction might make it overcompensate. :)

zeddrak:
Still it should make it work better for new players, and old players can probably figure it out... ;)

PR #2458, opened 2022-01-08
added that this function works in Bitnode 5

zeddrak:


hydroflame:
wont' work because that file is autogen but I did merge the change locally

PR #2460, opened 2022-01-08
Use wasted hashes on "Sell for Money"

### LINKED ISSUE #1893: extra hash should convert to money

MartinFournier:
Converts a portion of wasted hashes into money. 
Currently using the minimum value between the wasted hashes and 1% of the hash capacity, so that scripting the action would still be preferable.

Since the current cost of the action is 4 hashes, it won't trigger before 512 hash capacity.

Fixes #1893


hydroflame:
Great, but I'm going to change the 1% to 100%

PR #2461, opened 2022-01-08
Fix invalid sleeve's augmentation data (null ref)

MartinFournier:
I had an invalid sleeve augmentation name in my save which caused the
sleeve page to crash.

firefox_0e1CEhAM0F


hydroflame:
Would rather fix it like this
https://github.com/danielyxie/bitburner/issues/2471

PR #2482, opened 2022-01-08
Update hackingalgorithms.rst

KenJohansson:
Removed "Growth can be made more efficient by dividing it into many processes, instead of one script with a high thread count. Four grow scripts with 20 threads will outperform one grow script with 80 threads." since this has been proven not to be true.

Also brings this up to date with the information present on the full documenation.

### DIFF
--- a/doc/source/advancedgameplay/hackingalgorithms.rst
+++ b/doc/source/advancedgameplay/hackingalgorithms.rst
@@ -74,8 +74,6 @@ Meaning if we have space for 100 threads across the entire network 7 threads wil
 
 When utilizing this strategy, monitor the amount of money and security on the target server, if the money is not hovering around maximum and the security around the minimum, the ratios should be tweaked until that is the case.
 
-Growth can be made more efficient by dividing it into many processes, instead of one script with a high thread count. Four grow scripts with 20 threads will outperform one grow script with 80 threads.
-
 Utilizing `sleep()` or `asleep()` to ensure that your scripts do not all start at the same time can decrease the chance of issues associated with overhacking occurring. Both functions have a ram cost of zero.
 
 Batch algorithms (HGW, HWGW, or Cycles)


TorleifHensvold:
Does this include the possibility where "weaken" scripts cause their effect between the four "grow" scripts, thus lowering the security between each effect? I would also enjoy seeing the proof you speak of, if you've got a quick link to it. Would be nice to read.

heap-underflow:
I don't see how it could be proven not to be true.  https://github.com/danielyxie/bitburner/blob/master/src/Server/ServerHelpers.ts

Consider a server that starts at $100, ignoring all the various multipliers, there's clearly a difference between a single 100% growth (=$200) and five 20% growths compounded (=~$248.8)

KenJohansson:
Not my calculations, nor really my conversations. But discussions on the Discord for the past few days concluded that the math doesn't check out. A few snippets of that (=== between responses):

joesguns has $3.73m out of $5.00m (74.63%)
It will take 368 threads to grow to full and would end with a security of 10.59 (currently 9.119999999999992)

If we divide that in to 4 groups of 92...
The server will end up with $4.92m (98.44% of maximum) and would end with a security of 10.59
===
the security effect is something I had not considered, but the recursive function f(n) = (f(n-1)+1)growPct does grow marginally faster than the alternative
===
it does not
assume the growth multiplier is 1.008
assume starting money is 100
(100 + 1) * 1.008, repeated 80 times, gives you (approximately) 301.51465616
whereas (100 + 80) * 1.008 ** 80 gives you approximately 340.49839871
===
$100000, 5 threads, 1.0008 growth

5 threads = $100405.66
5 processes = $100405.65
it's super marginal
but it is worse
===
If you want to read the full discussion it's here: https://discord.com/channels/415207508303544321/415207508303544323/928809830834139176
And the final conclusion is here:
https://discord.com/channels/415207508303544321/415207508303544323/928815075014377602

So the documentation has already been updated given that information. (Commit 3769081) but RTD hasn't been updated due to it being missed.

heap-underflow:
Just did some testing and I was wrong -- apologies.  Both large and small servers seem to end up at virtually the same spot whether it's 1 grow operation of `x` threads or `n` grow operations of `x/n` threads.  

There are likely a few more places the docs need to be cleaned up to make this clear, IIRC around the "grow operation results are based on the current money available" parts (which is still clearly true, but made confusing by the per-thread flat adjustment that causes the results to line up) but I can't argue w these changes.

PR #2505, opened 2022-01-09
Had text from create program still present

zeddrak:
What does it return if the attempt failed? 0? If you know, maybe update with failed return value as well.


hydroflame:
need to update NetscriptDefinitions.d.ts

zeddrak:
Ok, will do that.

Issue #2538, opened 2022-01-10
`Browser` - Game Crash with `Error: Oh no! This is an error.` in `DevMenu`


MartinFournier:
# `Browser` - Game Crash with `Error: Oh no! This is an error.` in `DevMenu`

## How did this happen?

**Please fill this information with details if any.**

---

### Environment

* **Error**: `Error: Oh no! This is an error.`
* **Runtime**: `v1.3.0 (63a8679f) - Browser in "DevMenu"`
* **UserAgent**: `Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:95.0) Gecko/20100101 Firefox/95.0`
* **Features**: `lang=en-US cookiesEnabled=true doNotTrack=1 indexedDb=true`
* **Source**: `http://localhost:8000/main.bundle.js line 5575 > eval line 1 > eval`

### Stack Trace

```
General@webpack-internal:///919:21:83
DevMenuRoot@webpack-internal:///918:48:58
div
withEmotionCache/<@webpack-internal:///154:52:73
Box@webpack-internal:///1417:31:84
div
withEmotionCache/<@webpack-internal:///154:52:73
Box@webpack-internal:///1417:31:84
SnackbarProvider@webpack-internal:///673:737:24
SnackbarProvider@webpack-internal:///181:26:19
ErrorBoundaryBase@webpack-internal:///887:273:5
WithStyles@webpack-internal:///1435:70:30
GameRoot@webpack-internal:///98:241:18
div
withEmotionCache/<@webpack-internal:///154:52:73
Box@webpack-internal:///1417:31:84
LoadingScreen@webpack-internal:///650:39:19
InnerThemeProvider@webpack-internal:///1403:20:82
ThemeProvider@webpack-internal:///1404:48:7
ThemeProvider@webpack-internal:///1403:42:7
StyledEngineProvider@webpack-internal:///758:24:7
TTheme@webpack-internal:///210:332:16
```

### Save
```
*Add your save string here if possible (if not too big)*
```

PR #2540, opened 2022-01-10
Add an uncaught errors handler

MartinFournier:
- Handles uncaught errors and shows a page with details about the error & environment.
- Includes a link to open a new GitHub issue using the information from the page.
- Also adds a button to throw a new Error in the general dev menu.
- Refactored the Delete Game Button into its own component.

Current template looks like the content of this issue #2538

---

The examples in these screenshot are using `throw new Error("Oh no! This is an error.");`

firefox_Z1ewLNDmxg

firefox_G1BpZqFsIV

firefox_v9SYQcm6dV

firefox_qaNwbUDFUM

### DIFF
--- a/src/DevMenu/ui/General.tsx
+++ b/src/DevMenu/ui/General.tsx
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { useEffect, useState } from "react";
 
 import Accordion from "@mui/material/Accordion";
 import AccordionSummary from "@mui/material/AccordionSummary";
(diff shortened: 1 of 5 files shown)


MartinFournier:
So the current Windows external link launch does not work when having a query string on an URL. Also, explorer will truncate the data when urls are over 256ish characters. 

I added a workaround, but as far as I tested it does cause the process issue, again. It's only applied on url that have a query string, but I could revert it (8aee5ff) if you prefer.

smolgumball:
Potential overlap with #2551?

Issue #2513, opened 2022-01-09
Fix Documentation About Infiltrations


Playegage:
You should update the documentation on Infiltrations to tell that the number of fails you are allowed is proportional to how much health you have with between 7-max_hp being lost per fail depending on the difficulty of the infiltration.

Issue #2491, opened 2022-01-09
softReset requirements don't match


mike-feldmeier:
The [documentation](https://github.com/danielyxie/bitburner/blob/dev/markdown/bitburner.singularity.softreset.md) lists the RAM Cost at 5 GB, but inserting the function into a script increases it by 80 GB.

Issue #2444, opened 2022-01-07
n00dles might not be 16GB RAM


fddemora:
![n00dles](https://user-images.githubusercontent.com/58346374/148608020-65218e13-958e-4a4d-adec-913f7471125f.PNG)
https://bitburner.readthedocs.io/en/latest/guidesandtips/gettingstartedguideforbeginnerprogrammers.html

I would disagree here since n00dles is 4.00GB RAM. 
I'd make a pull request but I'm not sure where to find the file.

heap-underflow:
The rst source is here: [doc/source/guidesandtips/gettingstartedguideforbeginnerprogrammers.rst](https://github.com/danielyxie/bitburner/blob/master/doc/source/guidesandtips/gettingstartedguideforbeginnerprogrammers.rst)

Issue #2424, opened 2022-01-07
PurchasedServerSoftcap Bitnode Multiplier not documented


macdjord:
The `PurchasedServerSoftcap` multiplier is not documented, nor is its effect - making larger servers cost exponentially more expensive per gigabyte - mentioned in the descriptions of the bitnodes which it affects (which is most of them).

PR #2541, opened 2022-01-10
Documentation updates

Feodoric:
Bunch of documentation updates from the issues list:

#2444, #2513, #2403, #2281, #2269, #2201, #2219, #2424

Docs will need to be regenerated. Issues #2233 and #2491 were fixed in the definitions file by previous PRs but the docs haven't been regenerated since then.


daanflore:
Your changes to the netport documentation remove some possibilities.

The underlying type is any so allowing you to put anything in it I use it for objects.
But your changes to the doc only allow stringd and numbers.

The implementation where I based my changes to the documentation on:
https://github.com/danielyxie/bitburner/blob/dev/src/NetscriptPort.ts

daanflore:
Just took a look at the other code to write to port and see there are differences here to:

https://github.com/danielyxie/bitburner/blob/dev/src/NetscriptFunctions.ts
If you look at tryWritePort you are allowed to store anything in it here there is no check on the datatype.
But at the same time if you use writePort you are limited to the datatypes you added according to the code.

PR #2542, opened 2022-01-10
NetscriptDefinition.d.ts correcting based on code

daanflore:
I added the IPort interface to the definitions and also changed the return value of getPortHandle to IPort.

I based my changes on the following files:
definition:
https://github.com/danielyxie/bitburner/blob/dev/src/NetscriptPort.ts

code logic:
https://github.com/danielyxie/bitburner/blob/75283504fd70bbb943400056a43ba41aa68683e7/src/NetscriptFunctions.ts#L1906
https://github.com/danielyxie/bitburner/blob/75283504fd70bbb943400056a43ba41aa68683e7/src/NetscriptFunctions.ts#L445

Do not know if it should also be changed in de docs?
But they are auto generated so I do not know where it should be changed.

https://github.com/danielyxie/bitburner/blob/dev/markdown/bitburner.ns.getporthandle.md

would also need a new file to contain the correct return type of a port

### DIFF
--- a/src/NetscriptFunctions.ts
+++ b/src/NetscriptFunctions.ts
@@ -1901,7 +1901,7 @@ export function NetscriptFunctions(workerScript: WorkerScript): NS {
       const iport = helper.getValidPort("clearPort", port);
       return iport.clear();
     },
-    getPortHandle: function (port: any): any {
+    getPortHandle: function (port: any): IPort {
       updateDynamicRam("getPortHandle", getRamCost(Player, "getPortHandle"));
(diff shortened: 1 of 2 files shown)


daanflore:
I also changed the return value of ns.getporthandle to IPort and had no problems running the game

daanflore:
Should the IPort defenition be moved to NetscriptDefinitions.d.ts and removed from NetscriptPort.ts and work with an import from NetscriptDefinitions.d.ts ?

daanflore:
Seems like there is some overlap with: #2541 about changing to return value but different type is used

PR #2543, opened 2022-01-10
Make augment price multiplier prettier (was tested by Pigalot)

zeddrak:
Just an attempt to make the augment price multiplier look nicer.
I haven't set up my dev environment yet, but I didn't want to lose my place, so not tested - sorry.
(Will test shortly and update the message if all looks good).

tested by Pigalot (discord reference for screenshots, etc.: https://discord.com/channels/415207508303544321/923707505698283560/930230679705378816)


hydroflame:
ok well you can't use the formatExp function to format something that is not exp

PR #2549, opened 2022-01-11
clarify error

markusariliu:
The error message for running multiple instances of the same script with the same args should be updated to clarify that the args is the issue not just the multiple instances


Snarling (approved):
Looks like a good change.

PR #2565, opened 2022-01-11
Wipe running scripts when entering BitNode screen

MartinFournier:


hydroflame:
That's not how this should work actually. Any NS2 script still running could still run in the background

Issue #2562, opened 2022-01-11
Analyze should report backdoor installed


deathly809:
I cannot remember which servers I installed a backdoor on.  Because of this I just keep running backdoor on every server.  It would make since that analyze should report if a backdoor is installed.  Of course, backdoor could return an error if it is already installed.

deathly809:
Closing, PR merged!

PR #2566, opened 2022-01-11
Analyze now outputs if a server has a backdoor installed or not

deathly809:
Currently you have no way to determine if a server has a backdoor installed.  This adds a new output to analyze so that it reports that information.

This is in response to #2562


deathly809:
I have no idea if there is anything else I need to do, I saw that there are these mangled main.bundle.js and main.bundle.js.map files but not sure what to do with those.

deathly809:
Output from home

image

Output from n00dles

image

heap-underflow:
@deathly809 Just a heads up `ns.getServer(<hostname>).backdoorInstalled` will work for determining that programmatically.

deathly809:
@heap-underflow  Thanks!  I have tried using scripts to propagate to other nodes buy I keep having issues where I cannot tell if I have visited a node before.  So now I just do it by hand lol.

PR #2568, opened 2022-01-11
Update bitburner.singularity.installbackdoor.md

tywunon:
Returns: Promise<void>
True if the installation was successful?


hydroflame:
Need to change NetscriptDefinitions.d.ts

PR #2574, opened 2022-01-12
Fix stale faction data after work

### LINKED ISSUE #2569: Faction reputation not added until leaving faction screen
I've had been doing hacking contracts for Daedalus, after it said I've gained enough reputation, I clicked on 'Stop faction work' but the faction reputation doesn't get added until I click 'back' again to get to the factions overview, when I click Daedalus again, the reputation get's rewarded.

Imho, faction reputation should be updated as soon as I stop working.

MartinFournier:
Fixes #2569

Change the toFaction router action to work with a faction?.name instead of the full object.

Before that, we were keeping the previous page data and refreshing the page with that stale copy.

### DIFF
--- a/src/Bladeburner/ui/Stats.tsx
+++ b/src/Bladeburner/ui/Stats.tsx
@@ -40,7 +40,7 @@ export function Stats(props: IProps): React.ReactElement {
       joinFaction(faction);
     }
 
-    props.router.toFaction(faction);
+    props.router.toFaction(faction?.name);
   }
 
   return (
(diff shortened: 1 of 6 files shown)


nickofolas:
Please forgive me if I'm wrong here, but wouldn't mimicking the behavior of `toJob` and `toScriptEditor` be a more efficient fix?

I have a working fix which changes the `toFaction` method as such:
```ts
    toFaction: (faction?: Faction) => {
      if (faction) setFaction(Factions[faction.name]);
      setCurrentPage(Page.Faction, faction);
    },
```

If I'm missing something, please let me know! :)

MartinFournier:
Yep that's way better

Issue #2577, opened 2022-01-12
Quitting an action has no effect on player.focus


Feodoric:
Unsure if this is intended or not. Cancelling an action either through the UI or with singularity `ns.stopAction()` leaves the `player.focus` flag untouched. This results in scenarios where `ns.isBusy()` returns false, but `ns.isFocused()` returns true.

deathly809:
Not sure where UI call is located, but If you call stopAction it just returns you to the terminal, I see elsewhere that there is usually a stopFocusing before the return.

```ts
    stopAction: function (): any {
      helper.updateDynamicRam("stopAction", getRamCost(player, "stopAction"));
      helper.checkSingularityAccess("stopAction");
      if (player.isWorking) {
        if (player.focus) {
          Router.toTerminal();
        }
        const txt = player.singularityStopWork();
        workerScript.log("stopAction", () => txt);
        return true;
      }
      return false;
    },
```

deathly809:
For the cancel button I see that the finishWork does not unset focus, for cancel or finishing after 8 hours.  All the others do, I can do a PR to fix both.

phyzical:
i have 100% seen this behavior when i was trying to implemented optional focus to commitCrime, i just assumed it was by design tbh. most places kind of ignore it and override it anyway i.e in the singularity functions.

Hoekstraa:
Still happens on dev as of today;
![issue shown](https://user-images.githubusercontent.com/19653491/163249245-a44e64c8-229d-4d71-9452-f8d083e1f2ea.png)

PR #2580, opened 2022-01-12
Ensure that we leave focus when we call stopAction or cancel working

deathly809:
When a job is cancelled, we did not correctly reset focus.  In addition, when stopAction was called we did not reset focus there as well.

#2577 - This is in response to this issue.


src/NetscriptFunctions/Singularity.ts
```diff
@@ -738,6 +738,7 @@ export function NetscriptSingularity(
       helper.checkSingularityAccess("stopAction");
       if (player.isWorking) {
```
deathly809, line 739: I suspect because we are working something else might need to be done.  @hydroflame thoughts? I see there is a singularityStopWork function.

PR #2588, opened 2022-01-12
API: FIX #2391 atExit can now run netscript functions.

### LINKED ISSUE #2391: `atExit` does not execute properly if script is killed during `sleep`
When killing a script that was awaiting on `ns.sleep`, the `atExit` function cannot be called properly. Am I doing something wrong, or is this a bug?

Steps to reproduce:
1. Run the following script:
    ```Javascript
    /** @param {NS} ns **/
        ns.atExit(() => ns.tprint(ns.ls("home")));
    
        await ns.sleep(1000 * 60);  // Kill process during the sleep
    }
    ```
2. Kill the script during its sleep

Expected behaviour:
See the content of home printed in the CLI

Actual behaviour:
Nothing is printed, error is raised:
image

timotheeandres:
This PR aims to fix the issue presented in #2391, namely that `atExit` currently cannot work properly if its argument contains calls to ns.

### DIFF
--- a/src/NetscriptWorker.ts
+++ b/src/NetscriptWorker.ts
@@ -88,7 +88,7 @@ function startNetscript2Script(player: IPlayer, workerScript: WorkerScript): Pro
         "Concurrent calls to Netscript functions are not allowed! " +
         "Did you forget to await hack(), grow(), or some other " +
         "promise-returning function? (Currently running: %s tried to run: %s)";
-      if (runningFn) {
+      if (runningFn && !(runningFn === "sleep" && workerScript.exiting)) {
         workerScript.errorMessage = makeRuntimeRejectMsg(workerScript, sprintf(msg, runningFn, propName));
         throw workerScript;
       }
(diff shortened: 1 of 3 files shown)


Ornedan:
```js
	ns.atExit(() => {
		ns.tprint("atExit");
	});
	await ns.weaken("n00dles");
}
```
Once the changes are working as intended intended, this script should print "atExit" to terminal when killed.

src/NetscriptWorker.ts
```diff
@@ -88,11 +88,16 @@ function startNetscript2Script(player: IPlayer, workerScript: WorkerScript): Pro
         "Concurrent calls to Netscript functions are not allowed! " +
         "Did you forget to await hack(), grow(), or some other " +
         "promise-returning function? (Currently running: %s tried to run: %s)";
-      if (runningFn) {
+      if (runningFn && !(runningFn === "sleep" && workerScript.exiting)) {
```
jrchamp, line 91: Why ignore `sleep`? If we're 'exiting', shouldn't whatever previously running task already be canceled?

timotheeandres: I changed it (haven't worked on this PR in days now), but the idea was that if we are currently running `sleep` and we're exiting, then calling another function is allowed

PR #2594, opened 2022-01-12
Make the augments multiplier look nicer

zeddrak:
Sorry about the last one, thought that was format Exponent not Experience.
I used formatMultiplier this time, which seems more appropriate. :)
Might still need to go to a larger number type format, but that might be better addressed by modifying the formatMultiplier, so I'll leave it at just this for now. :)


zeddrak:
Alternatively, we can replace numeralWrapper.formatMultiplier(mult) with numeralWrapper.formatReallyBigNumber(mult,3)

PR #2603, opened 2022-01-13
Bound log boxes to parent div to prevent them from being dragged outside window

SlyCedix:


SlyCedix:
Closing to make sure log boxes stay inside the window when the window changes sizes

PR #2608, opened 2022-01-13
Update bitburner.ns.scan.md

shiggityshaggs:
an string -> a string


SagePtr:
See https://github.com/danielyxie/bitburner/blob/dev/CONTRIBUTING.md#as-a-documentor

PR #2619, opened 2022-01-13
MISC: Add an event log to persist certain game events

MartinFournier:
## Add an event log to persist certain game events

Allows errors & noteworthy events to be logged so that they are not lost
if missed by the player when AFK or when clicking away by mistake.

- Add an Event Log page listing the last events, auto-refreshed
- Add setting to select maximum event log capacity
- Refactor Options Sliders into its reusable component (OptionSlider)
- Some events may have an action to link to, like achievements
- Events can be tied to a unique key to grab dynamic data when loading the list
- Add a dev page to create dummy events, spam create & clear events
- Adds a flag to indicate if events have been seen with a counter in the
sidebar
- Log some events when certain actions are done or errors are thrown
  - When a player receives an achievement
  - When a player augments
  - When a player completes a BitNode
  - When a player gets promoted
  - When a gang member dies
  - When the player get errors trying to run bad scripts

### Testing

* Tested saving & triggering events around the game. 
* Tested with dummy events (dev menu).
* Tested on browser & electron.
* Tested with saves that had no events.
* Tested with bigger & smaller max log capacity.
* Tested extra metadata displayed for certain event types (like achievements)

### Screenshots

firefox_AjxP9YC84O

firefox_6GXa1f1zXD

firefox_AgfjrY5YOn

firefox_BQWgFlngQu

firefox_tEILIK4woV

### DIFF
--- a/src/Augmentation/AugmentationHelpers.tsx
+++ b/src/Augmentation/AugmentationHelpers.tsx
@@ -22,6 +22,7 @@ import {
   initUnstableCircadianModulator,
 } from "./AugmentationCreator";
 import { Router } from "../ui/GameRoot";
+import { EventLog, LogCategories, LogTypes } from "../EventLog/EventLog";
 
 export function AddToAugmentations(aug: Augmentation): void {
   const name = aug.name;
@@ -166,6 +167,7 @@ function installAugmentations(force?: boolean): boolean {
       break;
     }
   }
+  const nbAugmentations = Player.queuedAugmentations.length;
   for (let i = 0; i < Player.queuedAugmentations.length; ++i) {
     const ownedAug = Player.queuedAugmentations[i];
     const aug = Augmentations[ownedAug.name];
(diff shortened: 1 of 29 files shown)


MartinFournier:
When you got some time @hydroflame can you let me know what you think of this feature? Any changes you'd like me to make or just drop it altogether? No rush.

attrib:
Just stumbled across this, really like this idea. What I would like to see here in addition is a NetScript function, so a user could add events from their scripts to this, similar to toast/alert

phyzical:
Love the idea, few weeks ago would have said its needed 100% but now the tab for previous scripts logs makes things a little bit easier to debug.

But i think this is would be an amazing feature especially for newer players imo

phyzical:
@MartinFournier changes look sound to me. one suggestion, i wonder if we cant maybe move the event calls into the log helpers? i.e `console.log/error` , `dialogBoxCreate` ect.

That way it will be easier to enforce going forward.

maybe it takes the event object as an option param that just defaults to an info generic style event? or maybe its required to just make it explicit to future implementations (though having a default would make it easier for anywhere that may have been missed) maybe moving to required in a future change

Granted i don't think its that simple as we call logs and dialogs in pieces of code. but if it was just `dialogBoxCreate` that would probably be safe.

phyzical (approved):

MartinFournier:
> @MartinFournier changes look sound to me. one suggestion, i wonder if we cant maybe move the event calls into the log helpers? i.e `console.log/error` , `dialogBoxCreate` ect.
> 
> That way it will be easier to enforce going forward.

You're right, currently it'd be hard to enforce a generic style. I'd leave this for a future improvement though as the current implementation takes in more data than just the log helpers.

phyzical (approved):

hydroflame:
PR made before mega fmt

PR #2627, opened 2022-01-14
ns.corporation.getDivision() should get data based on type

endofzero:
ns.corporation.getDivision() should get data based on the immutable value of 'type' as 'name' is defined during creation of the division.

This allows you to check via program to see if you have expanded into a division without the need of knowing what the names are, it can build the names from the static type values.

# Documentation

# Bug fix

- Include how it was tested


hydroflame:
This is API breaking so ... no.

PR #2629, opened 2022-01-14
MISC: Terminal now supports ANSI-escape sequences.

heap-underflow:
Adds support for both 3- and 8-bit [ANSI escape sequences](https://en.wikipedia.org/wiki/ANSI_escape_code#Colors) for colors, as well as those for bold and underlined.

image

~Note: I marked this as a draft PR because it needs more testing.  It broke the CLI's `expr` output (which, I really think is the fault of `expr` breaking the "contract" of `Output.text` being a string and not the changes here, but I'm still putting it through its paces.~ All set to merge, imo.

I used something like:
```js
const ansi = (...c) => {
    return ns.sprintf("\u{001b}[%sm", c.join(';'))
}

ns.tprintf("%sred-on-cyan%s", ansi(31,46), ansi(0))
```
but a more friendly library could be built around it like js's chalk or python's colorama.


src/Terminal/ui/TerminalRoot.tsx
```diff
@@ -112,20 +112,136 @@ export function TerminalRoot({ terminal, router, player }: IProps): React.ReactE
...
+      }
+    }
+    if(parts.length === 0){
+      parts.push({code:null, text:item.text})
+    }
+    return (
+      <Typography classes={{ root: lineClass(item.color) }} paragraph={false}>
+        {parts.map((item, index) => {
+          return (
+            <Typography key={index} paragraph={false} component="span" sx={ansiCodeStyle(item.code)}>{item.text}</Typography>
```
surdaft, line 240: The reuse of item.text is a little confusing here. Is it possible to name it differently, just to avoid a moment of confusion? :)

src/Terminal/ui/TerminalRoot.tsx
```diff
@@ -112,20 +112,136 @@ export function TerminalRoot({ terminal, router, player }: IProps): React.ReactE
...
+    //  causes .matchAll to throw.  Ensure it's a string immediately
+    const parts = []
+    if(typeof item.text === 'string'){
+      // eslint-disable-next-line no-control-regex
+      const ANSI_ESCAPE = new RegExp('\u{001b}\\[(?<code>.*?)m', 'ug');
+      const matches = [null, ...item.text.matchAll(ANSI_ESCAPE), null]
+      if(matches.length > 2){
+        matches.slice(0, -1).forEach((m,i) => {
+          const n = matches[i+1]
+          if(m&&m.index!==undefined&&m.groups!==undefined){
```
surdaft, line 224: It would be better to fail early than to wrap the entire contents within an if statement, to reduce the amount of indentation. Eg;

```js
if (!m || m.index === undefined || m.groups === undefined) {
  return;
}

const startIndex = m ? m.index + m[0].length : 0
const stopIndex  = n ? n.index               : item.text.length
const partText   = item.text.slice(startIndex, stopIndex)
parts.push({code:m.groups.code, text:partText})
```

src/Terminal/ui/TerminalRoot.tsx
```diff
@@ -112,20 +112,136 @@ export function TerminalRoot({ terminal, router, player }: IProps): React.ReactE
...
-      return classes.warning;
+    type styleKey = "fontWeight" | "textDecoration" | "color" | "backgroundColor"
+    const style : {
+      fontWeight?: string;
+      textDecoration?: string;
+      color?: string;
+      backgroundColor?: string;
+    } = {}
+
+    if(code !== null){ /* eslint-disable yoda */
```
surdaft, line 168: It would be better to return style here if code is null, to reduce indentation as well as makes it clearer exactly what happens if code is null without scrolling to the bottom of this rather large block of code.

```js
if (code === null) {
  return style;
}
```

src/Terminal/ui/TerminalRoot.tsx
surdaft, line 240: The reuse of item.text is a little confusing here. Is it possible to name it differently, just to avoid a moment of confusion? :)

surdaft (approved):
Everything looks as though it makes sense, the only feedback is more about slight structure changes to make things clearer, earlier, but are non-blocking.

heap-underflow:
@surdaft thanks for the feedback, I made all those changes in the latest push

heap-underflow:
Along with some fixes, cleanup and code review, I made it so that text colorized with a background now uses `display:inline-block` instead of the default `display:inline` -- it makes it so that consecutive lines with backgrounds won't have gaps, e.g.
image

heap-underflow:
Marking as ready now, as I can't find any more ways to test or break it

heap-underflow:
For anyone interested, the script I threw together that generates those color swatches:

```js
/** @param {NS} ns **/
    const ansi = (...c) => {
        return ns.sprintf("\u{001b}[%sm", c.join(';'))
    }

    const lines = Array(19).fill(" ")
    const swatch = (fgbg, code, ...c) => ansi(fgbg, 5, code, ...c) + ns.sprintf("%3d", code) + ansi(0)

    function *row(fgbg, offset, ri){
        for(let i=0; i<18; i++){
            if(i%6 == 0){
                yield " "
                offset += 30
            }
            let fgc = ((fgbg == 48) && (ri>=3)) ? 30 : 0
            yield swatch(fgbg, offset + i - 30, fgc)
        }
    }

    for(let i of [0,1,17,18]){
        lines[i] += " "
    }

    for(let r=0; r<6; r++){
        lines[r+ 3] = [...row(38,  16+6*r, r), ' ', ...row(48,  16+6*r, r)].join(' ')
        lines[r+10] = [...row(38, 124+6*r, r), ' ', ...row(48, 124+6*r, r)].join(' ')
    }

    for(let i=0; i<8; i++){
        lines[ 0] += ansi(30+i) + ns.sprintf("%3d", i) + ansi(0) + " "
        lines[ 1] += ansi(40+i) + ns.sprintf("%3d", i) + ansi(0) + " "
    }

    const GREY = [...Array(12).keys()]
    const GS = ' '.repeat(33)
    lines[17] += GREY.map(i => swatch(38, 232+i)).join(" ") + GS + GREY.map(i => swatch(48, 232+i,  0)).join(" ")
    lines[18] += GREY.map(i => swatch(38, 244+i)).join(" ") + GS + GREY.map(i => swatch(48, 244+i, 30)).join(" ")


    lines[0] += "    " + ansi(1) + "BOLD TEXT"
    lines[1] += "    " + ansi(4) + "UNDERLINED TEXT"

    lines.forEach(line=>ns.tprintf("%s", line))
}
```

surdaft (approved):

Ornedan:
Are changes to any files besides `TerminalRoot.tsx` necessary? Are the DOM element IDs added to `SidebarRoot.tsx` and `Overview.tsx` used anywhere? 

And changes to the `package-lock.json`s don't belong in a feature branch unless maybe if you're adding a new dependency.

phyzical:
don't mean to be rude at all. im sure you have tested it well, but i would be a bit nervous that everything going through the terminal has all this parsing stuff, if/when we have a bug it would be very difficult to diagnose the this code.

maybe there is a way we can put this behind some sort of flag?

also just a suggestion to make some sort of class to abstract away the ascii magic into its own file rather than just a function.

tje:
This is probably the feature I'm most excited to see get rolled into the game, and I really like the idea of emulating the familiar implementations from existing terminals/shells

Some suggestions for this:
- Integration with the log box! Could abstract it to a separate component to reuse in the `LogBoxManager` as well as the `TerminalRoot`
- Theme support, at least for the shorter 4-bit color range (30-37, etc) - nice for muted colors in pastel flavored themes, or brighter ones for high contrast or whatever
- Style retention/reset markers

Elaborating on the last one (reset markers) - when I was testing this I noticed that the text style would completely reset with each new code, rather than amend and retain the current configuration like existing shells. For example, if I wanted to print white text on a magenta background, with one red word in between:

```sh
# Bash
echo -e "\x1b[37;45m White \x1b[31mRed \x1b[37mWhite \x1b[0m Reset"
```

image

Versus the current implementation in Bitburner:

```js
// Netscript
ns.tprint('\x1b[37;45m White \x1b[31mRed \x1b[37mWhite \x1b[0m Reset')
```

image

Expected sequence:
- `37;45` - white foreground, magenta background respectively
- `31` - switch to red foreground
- `37` - back to white foreground
- `0` - reset everything to default

---

I tried taking a stab at this on my own fork, though I only covered the 4-bit color range and basic weight/decoration: https://github.com/danielyxie/bitburner/compare/dev...tje:colored-log-output

Tried isolating it as a separate React component to plug in to the terminal + log box, and fetch color codes from the theme where available. Not a whole lot of confidence in what I'm doing tho, this is literally my first time working in React

PR #2630, opened 2022-01-14
Overview skill hooks breaking change + improve tooltips

nickofolas:
### Tooltips
Skill stat tooltips are now attached to the skill text instead of the progress bar, making it easier to hover over them, and also allowing them to be used when progress bars are disabled.

<table>
    <th>Before</th>
    <th>After</th>
    <tr>
      <td>
      </td>
      <td>
      </td>
    </tr>
  </table>

### Skill Overview Hooks (Breaking-ish Change)
Reworked the skill hooks system somewhat extensively, adding 2 separate hooks per skill (`overview-{skill}-hook-0` and `overview-{skill}-hook-1`). **This change in hook ID naming is the breaking change.**

This new system allows for hooks to be implemented in a way similar to the pre-existing extra hooks. Rather than only having one cell available, hooks can now be made in 2 cells, as shown in the image below.

<table>
  <th>Before</th>
  <th>After</th>
  <tr>
    <td>
    </td>
    <td>
    </td>
  </tr>
</table>

### DIFF
--- a/src/ui/React/CharacterOverview.tsx
+++ b/src/ui/React/CharacterOverview.tsx
@@ -26,6 +26,7 @@ import { BitNodeMultipliers } from "../../BitNode/BitNodeMultipliers";
 
 import { Box, Tooltip } from "@mui/material";
 import { CONSTANTS } from "../../Constants";
+import { ISkillProgress } from "src/PersonObjects/formulas/skill";
 
 interface IProps {
   save: () => void;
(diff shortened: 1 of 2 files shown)


phyzical:
had a quick scan looks good 👍

granted ive never used these hooks before, i assume its more of a developer breaking change?

nickofolas:
yeah it's not a hugely noticeable break outside of a (presumably) smaller group of people who use the current hooks implementation

but for those affected, it's an easy fix and allows more flexibility

alainbryden:
I'm pretty sensitive to backwards incompatibilities. I think you could introduce the new hooks without breaking the two existing hooks. These were deliberately added to the DOM for players, so they are (almost) as important to maintain backwards-compatilibity with as any of the API functions would be.

phyzical:
yeah i agree with @alainbryden comment, would maybe leaving the original hook name the same instead of adding `-1` work?

nickofolas:
I think that, at least for now, I'll close this PR, and when PRs are less clogged up, come up with a better implementation.

phyzical:
nah leave it open :) maybe mark it as Feature so itll be ignored in the first round of fixes

Issue #2637, opened 2022-01-15
Cannot save scripts anymore in dev


deathly809:
Apparently, this does not have a main function?  Trying to figure out what is going on.

```ts
/** @param {NS} ns **/
export async function main(ns) {
	ns.print(ns.getWeakenTime('n00dles'));
}
```

![image](https://user-images.githubusercontent.com/1381210/149603392-afe42572-c41c-4df3-848c-722c7990d858.png)

deathly809:
I think I know what is wrong.  Someone made a typo and did not test their PR.

deathly809:
@hydroflame  - Please review this PR

https://github.com/danielyxie/bitburner/pull/2638

PR #2638, opened 2022-01-15
When saving we were not saving the script content

deathly809:
# Documentation

# Bug fix

In a previous PR when formatting code to remove whitespace they developer did not reference the function parameter but rather the script parameter code.  This means that the file was never saved.

- Include how it was tested
Referenced correct parameter, saved a script and executed it.

image 

I added a test as well,

image

Issue: #2637

### DIFF
--- a/src/Script/Script.ts
+++ b/src/Script/Script.ts
@@ -118,7 +118,7 @@ export class Script {
    */
   saveScript(player: IPlayer, filename: string, code: string, hostname: string, otherScripts: Script[]): void {
     // Update code and filename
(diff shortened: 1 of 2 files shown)


deathly809:
@MartinFournier  - You seemed to have introduced a bug, can you please review this PR.  #2623

MartinFournier:
> @MartinFournier - You seemed to have introduced a bug, can you please review this PR. This is the breaking PR #2623

Damn, sorry about that and thanks for the fix!

Issue #2602, opened 2022-01-13
Doc update: function `getWeakenTime` does not accept optional params


fmmatheus:
### Issue

According to `getWeakenTime` docs:

> Returns the amount of time in milliseconds it takes to execute the weaken Netscript function on the target server. **_The function takes in an optional hackLvl parameter that can be specified to see what the weaken time would be at different hacking levels._** The required time is increased by the security level of the target server and decreased by the player's hacking level.

The function actually does not accept optional params, according to it's signature:

```js
getWeakenTime(host: string): number;
```

### Proposal

1. Add the optional param and re-write the function to calculate the weaken time in different hack levels
2. Update `docs` and remove the part where it mentions the optional param

deathly809:
I totally misread this lol.  I will do a follow-up PR tomorrow to make the hackLvl optional (I actually don't think it even takes a hackLvl at all).

deathly809:
@fmmatheus It looks like it could be annoying to do this.  Right now the code is using player object to get the players hacking time.  There is a formula that uses multiple values from the player.

```
player.hacking
player.hacking_speed_mult 
player.intelligence
```

If we did allow a way to override the hacking level it would have to assume that all other values are constant, which might not be true.

```ts
export function calculateHackingTime(server: Server, player: IPlayer): number {
  const difficultyMult = server.requiredHackingSkill * server.hackDifficulty;

  const baseDiff = 500;
  const baseSkill = 50;
  const diffFactor = 2.5;
  let skillFactor = diffFactor * difficultyMult + baseDiff;
  // tslint:disable-next-line
  skillFactor /= player.hacking + baseSkill;

  const hackTimeMultiplier = 5;
  const hackingTime =
    (hackTimeMultiplier * skillFactor) /
    (player.hacking_speed_mult * calculateIntelligenceBonus(player.intelligence, 1));

  return hackingTime;
}
```

fmmatheus:
@deathly809 I think the code should keep using the `player` object. I mean, `hacking_speed_mult` and `intelligence` values won't change for the current `run`. The third optional argument for `calculateHackingTime` will receive the `hacking` skill to simulate the hacking time, otherwise will default to the player hacking skill. Wdyt?

phyzical:
how about keeping the player object but also accept a clone object "overrides" that contains all the important props so by default it uses the player i.e current context but also allows the user to whatever theory crafting they like.

(unless this was already addressed by something else)

PR #2640, opened 2022-01-15
GetWeakenTime should take optional hostname as docs suggest

deathly809:
Issues : #2602

# Documentation

# Bug fix

- Include how it was tested - **Manaully**

# With hostname

image

image

# Without hostname

image

image


fmmatheus:
If I understood `docs` correctly, the `hostname` is not an optional params, the `hackLvl` is the optional param.

> Returns the amount of time in milliseconds it takes to execute the weaken Netscript function on the target server. The function takes in an optional hackLvl parameter that can be specified to see what the weaken time would be at different hacking levels. The required time is increased by the security level of the target server and decreased by the player's hacking level.

deathly809:
@fmmatheus Looks like @hydroflame  thinks hostname is optional now LOL

PR #2641, opened 2022-01-15
Copy terminal command history into game save

MartinFournier:
## Copy terminal command history into game save

Copy the terminal commands into the player object so that they are saved between game reloads. Helpful while developing with hot-reload to not having to retype commands.

Adds a 'history' command to display history, and a 'history -c' command to clear it from both the current terminal & the player's save. I copied the `history` command (https://ss64.com/bash/history.html), but only implemented -c since the others seemed mostly useless.

Note that it is not cleared on BN prestige, but I believe that's fine like that since we still keep our scripts.

### Testing

Tested terminal history on game reloads, page changes, augmentation & BN prestige

### Screenshots

firefox_NtE11SVZaQ

firefox_w20ha64e40

firefox_UymZY134Kt

### DIFF
--- a/src/PersonObjects/IPlayer.ts
+++ b/src/PersonObjects/IPlayer.ts
@@ -72,6 +72,7 @@ export interface IPlayer {
   sourceFiles: IPlayerOwnedSourceFile[];
   exploits: Exploit[];
   achievements: PlayerAchievement[];
+  terminalCommandHistory: string[];
   lastUpdate: number;
   totalPlaytime: number;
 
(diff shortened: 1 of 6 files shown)


heap-underflow:
If/when this gets merged, if you're open to it, I'd like to add the `!<historyId>` command that lets you quickly re-run a history entry (couldn't find a great example resource, but [this one is okay](https://www.computerworld.com/article/2717119/linux-tip--using-an-exclamation-point-----to-reference-events.html)).  So instead of hitting up arrow twenty times, you can type `history` find the entry you want from the list, then type `!<someId>`

Larandar:
As a feature creep I will add to @heap-underflow comment with adding a `CTRL+r` shortcut for fuzzy search in the history. But it might better fit in a "Terminal Enhancement" milestone with a `fish/zsh` or `fzf` style autocomplete selector

PR #2642, opened 2022-01-15
CORPORATION: Add STORE amount variable in sell product/material actions

c-goldschmidt:
Add a feature i find annoying to be missing: retaining a certain (fixed) amount in store, either to sell later when the price is better or for materials that are used for boosting but also produced (i.e. AI Cores in software).

This adds a new variable "STORE" to the sell material and sell product dialogs which will sell all in store, it can then be used in formulas (i.e. `STORE-10` would sell all in store but keep 10/s).

Fly-by fixed the "price" and "quantity" being switched up in the error messages.

# Documentation

# Not a Bug Fix
image

### DIFF
--- a/src/Corporation/Industry.ts
+++ b/src/Corporation/Industry.ts
@@ -838,6 +838,7 @@ export class Industry implements IIndustry {
                   //Dynamically evaluated
                   let tmp = (mat.sllman[1] as string).replace(/MAX/g, (maxSell + "").toUpperCase());
                   tmp = tmp.replace(/PROD/g, mat.prd + "");
+                  tmp = tmp.replace(/STORE/g, (mat.qty / (CorporationConstants.SecsPerMarketCycle * marketCycles)) + "");
                   try {
                     sellAmt = eval(tmp);
                   } catch (e) {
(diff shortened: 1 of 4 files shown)


src/Corporation/Industry.ts
```diff
@@ -838,6 +838,7 @@ export class Industry implements IIndustry {
                   //Dynamically evaluated
                   let tmp = (mat.sllman[1] as string).replace(/MAX/g, (maxSell + "").toUpperCase());
                   tmp = tmp.replace(/PROD/g, mat.prd + "");
+                  tmp = tmp.replace(/STORE/g, (mat.qty / (CorporationConstants.SecsPerMarketCycle * marketCycles)) + "");
```
phyzical, line 841: is this bonus time safe too?

src/Corporation/Industry.ts
```diff
@@ -1199,6 +1199,7 @@ export class Industry implements IIndustry {
               //Sell amount is dynamically evaluated
               let tmp = product.sllman[city][1].replace(/MAX/g, (maxSell + "").toUpperCase());
               tmp = tmp.replace(/PROD/g, product.data[city][1]);
+              tmp = tmp.replace(/STORE/g, (product.data[city][0] / (CorporationConstants.SecsPerMarketCycle * marketCycles)) + '');
```
phyzical, line 1202: is this bonus time safe too?

hydroflame:
PR made before mega fmt

src/Corporation/Industry.ts
c-goldschmidt: doesn't bonus time just reduce the time between market cycles? tbh, the code here is pretty hard to read.

phyzical: yeah tbh i have no idea, theres a reason this code causes so many bugs ;D

PR #2648, opened 2022-01-15
Add function signature for scp(files, destination)

SlyCedix:
adds
```ts
  scp(files: string | string[], destination: string): Promise<boolean>;
```
to NetscriptDefinitions.d.ts for calling scp with source omitted


SlyCedix:
Oof CI didn't like this, will have to look around a bit

hydroflame:
Yeah so the game won't compile interface with overloaded definitions.

PR #2653, opened 2022-01-16
Refactor `for ... in` loops

### LINKED ISSUE #2647: Messing with Array.prototype causes City to not render
1. Run this (minimal) script
```ts
Array.prototype.WHATEVER = function (element) {
    this.indexOf;
};

export const main = async (ns) => { ns.tprint('done') };
```

The function added to Array.prototype can be called anything however it cannot be an anonymous (arrow) function.
Inside the funciton `this.indexOf` must be referred to.
2. Click city, the game will crash.
3. Notice that the value on [this line](https://github.com/danielyxie/bitburner/blob/dev/src/Locations/ui/City.tsx#L134) will be `typeof props.city.asciiArt === 'string'` but the array returned by `split` will contain every line of the ASCII city, as expected, but also contains as last element the function we declared above.
4. Go "wtf" out loud.

nickofolas:
As discussed on Discord, this PR refactors all `for ... in` loops (save for a select few) to make use of `for ... of Object.keys()` instead. Additionally, in a few cases that used `for ... in` to iterate the items of an `Array`, I changed it to `for ... of Array`.

### Note
I have tested each variant of change I have made (that is, anything that is not just `for ... in` -> `Object.keys()`), and have tested the standard changes as well.

Having confirmed that all implementations work as expected, I believe that this PR causes no problems. If any are found, though, I would appreciate a callout!

Fixes #2647

### DIFF
--- a/src/Terminal/determineAllPossibilitiesForTabCompletion.ts
+++ b/src/Terminal/determineAllPossibilitiesForTabCompletion.ts
@@ -204,7 +204,7 @@ export async function determineAllPossibilitiesForTabCompletion(
 
   if (isCommand("buy")) {
     const options = [];
-    for (const i in DarkWebItems) {
+    for (const i of Object.keys(DarkWebItems)) {
       const item = DarkWebItems[i];
       options.push(item.program);
     }
--- a/src/Alias.ts
+++ b/src/Alias.ts
@@ -22,12 +22,12 @@ export function loadGlobalAliases(saveString: string): void {
 
 // Prints all aliases to terminal
 export function printAliases(): void {
-  for (const name in Aliases) {
+  for (const name of Object.keys(Aliases)) {
     if (Aliases.hasOwnProperty(name)) {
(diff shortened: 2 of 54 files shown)


src/Terminal/determineAllPossibilitiesForTabCompletion.ts
```diff
@@ -204,7 +204,7 @@ export async function determineAllPossibilitiesForTabCompletion(
 
   if (isCommand("buy")) {
     const options = [];
-    for (const i in DarkWebItems) {
+    for (const i of Object.keys(DarkWebItems)) {
```
paulcdejean, line 207: Good idea, but work left to be done.

For instance this should be:

```js
for (const item of DarkWebItems) {
  options.push(item.program)
}
```

nickofolas: `DarkWebItems` is not an array, rather a mapping, and cannot be iterated in that way. Save for a few instances, all changes are to iterations of map-like objects.

PR #2661, opened 2022-01-16
Added getDarkwebProgramCost and related unit tests to the api. 

neuralsim:
Gets the price of a specified darkweb program.

```js
let price = ns.getDarkwebProgramCost("brutessh.exe");
// get the price

ns.tprint(price);
// print the price to the terminal. Will output -1 if Tor has not been purchased, 
// otherwise will print 500000 if brutessh.exe has not been purchased yet, or 
// 0 if it has already been purchased. 

price = ns.getDarkwebProgramCost("bogusProgram.exe");
// throws an exception, since bogusProgram.exe isn't a darkweb program 
```

The reasoning for throwing an error if the program name is incorrect is that 99% of the time, if players get the name of the program wrong, it's probably the first time they've tried to execute the script, so giving them a big fat error will help with rapid debugging.

I followed the example of purchaseProgram in terms of the unit tests I created, and for the documentation.

### DIFF
--- a/dist/bitburner.d.ts
+++ b/dist/bitburner.d.ts
@@ -4794,7 +4794,10 @@ export declare interface Singularity {
      * You MUST have a TOR router in order to use this function. The price returned 
      * by this function is the same price you would see with buy -l from the terminal.
      * Returns the cost of the program if it has not been purchased yet, 0 if it 
-     * has already been purchased, or -1 if the program does not exist
+     * has already been purchased, or -1 if Tor has not been purchased (and thus
+     * the program/exploit is not available for purchase).
+     * 
+     * If the program does not exist, an error is thrown.
      * 
      *
      * @example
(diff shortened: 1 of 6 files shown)


alainbryden:
I feel like hard-coding the costs of these programs is easier than hard coding the names (with the correct casing). I would propose that this function on its own is not useful unless you are also adding an API for getting the names of the dark web programs.

neuralsim:
I did add a function to get the names

neuralsim:
> I feel like hard-coding the costs of these programs is easier than hard coding the names (with the correct casing). I would propose that this function on its own is not useful unless you are also adding an API for getting the names of the dark web programs.

See PR #2662

phyzical (approved):
LGTM

hydroflame:
Included elsewhere

PR #2662, opened 2022-01-16
Add function to get available darkweb programs

neuralsim:
Included:

- changes to implement the function described/sampled below
- fix minor code formatting issue in DarkwebItems.ts (make ServerProfiler entry like all the others)

Adds a new function getDarkwebProgramsAvailable, which returns a list of the programs available for purchase on the darkweb, optionally including the programs the player has already purchased.
Get a list of darkweb programs available. 
```js
let programs = ns.getDarkwebProgramsAvailable(true);
for (const program of programs) {
  // getDarkwebProgramCost was added with my last PR, uncomment the next line if it's accepted and comment the line after it to show programs and cost
  //let cost = ns.getDarkwebProgramCost(program).toMoney();
  let cost = "unavailable";
  ns.tprint(`Program ${program} is available on the darkweb for ${cost}`);
}
```


phyzical (approved):
LGTM

hydroflame:
Included elsewhere, (will still credit you)

PR #2699, opened 2022-01-19
ns: add `ns.getRecentScripts()` to return recently killed scripts

smolgumball:
## New `NS` addition: `ns.getRecentScripts(): RecentScript[]`

chrome_YpKUWGm0h5

* RAM cost: 0.2 GB
* Returns a sanitized array of `RecentScript` objects, copied from those used by the internal `RecentScriptAccordion` component and the Active Scripts -> Recent UI.
* The `RunningScript` objects attached to the `runningScript` key are sanitized using existing code from `ns.getRunningScript()` to ensure keys like `dataMap` and `dependencies` are not exposed via NS.

## New Setting

image

```ts
/**
 * Limit the number of recently killed script entries being tracked.
 */
MaxRecentScriptsCapacity: number;
```

* Added configuration option to replace existing hardcoded limit of `50` on the Recent Scripts UI. New setting defaults to 50, with increments of 25 up to a maximum of 500. Players shouldn't notice any change in performance unless they tweak the setting.

## Tested

* Local build in Chrome + local build in Electron
* Ran a test script a number of times, both before and after saving and reloading a game

chrome_cMvNrzJ189


src/ScriptEditor/ui/ScriptEditorRoot.tsx
```diff
@@ -813,20 +822,30 @@ export function Root(props: IProps): React.ReactElement {
...
           <Typography sx={{ mx: 1 }}>
             {" "}
-            Documentation:{" "}
+            <strong>Documentation:</strong>{" "}
             <Link target="_blank" href="https://bitburner.readthedocs.io/en/latest/index.html">
               Basic
-            </Link>{" "}
-            |
+            </Link>
+            {" | "}
```
smolgumball, line 848: Cleaned up a bit of spacing here, looks like Prettier did it's thing as well. Let me know and I can yoink these changes from the PR

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -99,22 +99,56 @@ interface Player {
 /**
  * @public
  */
-interface RunningScript {
+export interface RunningScript {
```
smolgumball, line 102: Added docs to the NS-facing `RunningScript` object, pulled from the internal `RunningScript` type

phyzical (approved):
LGTM

hydroflame:
PR made before mega fmt

PR #2714, opened 2022-01-19
Separate WARN/ERROR/SUCESS/INFO into their own functions

jumpsplat120:
Instead of having to prepend a message with WARN/ERROR/SUCCESS/INFO, they each have their own function.

There'd now be a tprint, twarn, tsuccess and terror, as well as format versions of all of those.


pigalot:
@jumpsplat120 should this not be backwards compatible as well? ! see no good reason to change the way tprint works to add this.

jumpsplat120:
> @jumpsplat120 should this not be backwards compatible as well? ! see no good reason to change the way tprint works to add this.

Ya, you're right, I can readd the tprint backwards compatibility. Someone closed this though.

PR #2716, opened 2022-01-20
Enhance api server

lordducky:
- Used restful patterns for handling server calls (GET/PUT/DELETE)

- Utilized Server interface for saves rather than re-implementing code

- Exposed operations for getting code-base of scripts on the home server and deleting files.

- Added a common response structure

### DIFF
--- a/electron/api-server.js
+++ b/electron/api-server.js
@@ -12,10 +12,12 @@ async function initialize(win) {
   window = win;
   server = http.createServer(async function (req, res) {
     let body = "";
+    res.setHeader('Content-Type', 'application/json');
 
     req.on("data", (chunk) => {
       body += chunk.toString(); // convert Buffer to string
     });
+
     req.on("end", () => {
       const providedToken = req.headers?.authorization?.replace('Bearer ', '') ?? '';
       const isValid = providedToken === getAuthenticationToken();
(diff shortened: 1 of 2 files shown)


surdaft (approved):
Looks good! Would be nice to be able to trust VSCode to delete the "- Copy" files when I copy/paste a file and then rename it. :')

smolgumball:
looks great, would love to be able to do a proper bi-directional sync w/ the vscode extension and this feels like solid progress to that end

surdaft (approved):
Dang, good spot! Missed that when reviewing.

lordducky:
> looks great, would love to be able to do a proper bi-directional sync w/ the vscode extension and this feels like solid progress to that end

I was looking to get to that point. I'm looking to hit a nice middle ground for the moment. But if you ask me, the proper approach to this would be a websocket model.

Right now the communication is omnidirectional (VScode making requests to the bitburner http server), but for true bidirectional communication, you would ideally use a websocket layer to stream events between the VScode client and the Bitburner web client so that terminal changes are also picked up.

But that would mean a big rewrite of the vscode module and the api-server as it stands.

lordducky:
@surdaft Made one more addition, added ramUsage and hash as part of the returns.

This probably is needed for dirty checking based on my own dev-server prototypes around these changes.

surdaft (approved):

PR #2727, opened 2022-01-20
Script Editor Tab Actions

nickofolas:
Improves the functionality of script editor tabs in a couple of ways:
- Middle clicking on an open tab will now close it (like browser tabs)
- Added a new "sync" button to tabs that will replace the editor content with the content of the file by the same name on the server
  - Useful for cases like working via the API server, so scripts updated over the API can be viewed properly in the editor without having to close and reopen them
  - If the editor is up to date with the server content, then the button press will be ignored
- Slight style adjustments to editor tabs for consistency and continuity
image

### Demo
firefox_Ru6zq3IU5C

### Testing
Tested the following cases:
- Middle clicking active and inactive tabs
- Middle clicking with only one tab
- Syncing the currently active tab
- Syncing an inactive tab

Tested on Firefox 97.0

### DIFF
--- a/src/ScriptEditor/ui/ScriptEditorRoot.tsx
+++ b/src/ScriptEditor/ui/ScriptEditorRoot.tsx
@@ -702,7 +702,8 @@ export function Root(props: IProps): React.ReactElement {
 
     if (openScript.code !== serverScriptCode) {
       PromptEvent.emit({
-        txt: "Do you want to overwrite " + openScript.fileName + " with the contents saved on the server? This cannot be undone.",
+        txt: "Do you want to overwrite the current editor content with the contents of " +
+          openScript.fileName + " on the server? This cannot be undone.",
         resolve: (result: boolean) => {
           if (result) {
             // Save changes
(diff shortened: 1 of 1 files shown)


smolgumball:
hey @nickofolas, does this degrade gracefully if you open a file on one server, then connect to another, then attempt to "sync" the contents? i've run into odd RAM calc issues in the past due to opening a file on server A and then attempting to save on server B, so just thought I'd raise this particular edge case for testing

nickofolas:
could you explain what you mean by degrade? sorry - i'm happy to test and make sure the case is accounted for and fix it if not, just not sure what you mean :)

smolgumball:
ah apologies, i suppose i meant the opposite of "enhance", i.e. "progressive enhancement" -- like if the syncing operation isn't supported while connected to a different server than the one you opened the file from, does the sync operation fail gracefully, succeed without issues, notify the user, etc?

nickofolas:
ah gotcha. in my limited testing, it *seems* like it's accounted for. testing process:
1. run script that will edit file on `home` in 2.5 seconds
2. immediately change connection to CSEC
3. wait for edit operation to complete
4. sync, new contents of script on `home` replace open editor content
by my understanding, i think it handles it appropriately?

PR #2745, opened 2022-01-21
Fix incorrect result being detected on the roulette table

### LINKED ISSUE #2086: Roulette in Aevum
While playing roulette in Iker Molina Casino, if I bet red a few times consecutively, then switch to black then land on black, I still lose.  It only happens that one time after switching, just kinda annoying to lose when I should've won.

### LINKED ISSUE #2382: Casino Roulette Seems Broken
I have tested this multiple times and it appears all over the place. If I bet 1-12, I'll win when it is 24B, if I go black, i'll lose when it shows 13B, etc, etc. seems borked

edit: reading the code it looks like the CPU does "cheat" but I would expect to not be cheated in such a ridiculous manner. First of all, if it comes up 24B and I choose 1-12 , I shouldn't be winning. Everything in the code looks fine to me except the cheating section. I'm just not sure how that affects the game or why it would make me win when I should lose. idk if there's a disconnect between the result and the display, if there is I can't see how by my cursory reading of the code.

jamie-mac:
Resolves #2382
Resolves #2086
Also fixes an issue where choosing black would result in a win if 0 came in.

Fixes were manually tested in the browser after running `npm run start:dev`.

The main issue (for the first 2 fixes) was that it always used the previous strategy, rather than the current one, when doing the check.

The issue was easy to recreate as the first spin always lost no matter the result:
1. go to the casino
2. select roulette
3. choose black (or anything that has to high chance, to speed up process)
4. if black comes in, notice you lose
5. if red comes in, select "stop playing" and go back to step 2.

I solved it by removing the reliance on state (as it wasn't updating before the result was checked), and just using the strategy directly. The slots suffer from a similar issue, so i'll submit a fix for that too at some point.

### DIFF
--- a/src/Casino/Roulette.tsx
+++ b/src/Casino/Roulette.tsx
@@ -40,13 +40,13 @@ const strategies: {
 } = {
   Red: {
     match: (n: number): boolean => {
-      if (n === 0) return false;
       return redNumbers.includes(n);
     },
     payout: 1,
   },
   Black: {
     match: (n: number): boolean => {
+      if (n === 0) return false;
       return !redNumbers.includes(n);
     },
     payout: 1,
--- a/src/Casino/Roulette.tsx
+++ b/src/Casino/Roulette.tsx
@@ -118,12 +118,6 @@ export function Roulette(props: IProps): React.ReactElement {
   const [status, setStatus] = useState<string | JSX.Element>("waiting");
   const [n, setN] = useState(0);
   const [lock, setLock] = useState(true);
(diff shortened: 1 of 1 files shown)


src/Casino/Roulette.tsx
```diff
@@ -40,13 +40,13 @@ const strategies: {
 } = {
   Red: {
     match: (n: number): boolean => {
-      if (n === 0) return false;
```
jamie-mac, line 43: this wasn't required, as redNumbers is an array that doesn't contain a zero, so would never match anyway

PR #2752, opened 2022-01-22
NetscriptDefinitions.d.ts: CrimeStats.name is a string, not a number

tigercat2000:
Title pretty much says it all, this was incorrectly typed as a number in NetscriptDefinitions.d.ts while getCrimeStats() in CrimeHelpers.ts returns a Crime class, which uses a string for the name.


tigercat2000:
And naturally, I completely missed the fact that #2737 already exists! Nevermind then.

Issue #2719, opened 2022-01-20
Expressions for setting material quantity not working anymore


ltoni90:
**Game Version**: v.1.4.0
**Source File**: [Corporations/Actions.ts](https://github.com/danielyxie/bitburner/blob/dev/src/Corporation/Actions.ts)
**Line Number**: 120
**OS**: Windows 10
**Description**:
Before v1.4.0 it was possible to use expressions like 'PROD-300'. This does not work anymore because of commit [98e8910c3aed719003730acd785d7aa9f1d118b7](https://github.com/danielyxie/bitburner/commit/98e8910c3aed719003730acd785d7aa9f1d118b7#diff-741b7886774b08ffc9963810bfac1e2700943d5effa7a653ff3569e522e538dc). At [line 120 in Actions.ts](https://github.com/danielyxie/bitburner/blob/34f18a6000a1c14ecd1a12712e8a86747a54f02e/src/Corporation/Actions.ts#L120) a new condition `parseFloat(tempQty) < 0` was added and this is causing the exception "Invalid value or expression for sell price field" as the variables MAX and PROD will be both replaced by 1 for a temporary evaluation of the expression. Either remove condition or rewrite logic.

Thank you!

ltoni90:
Hey guys, as this is a game breaking bug for me regarding corporations, I fixed the bug by myself and created a pull request. If something should not be correct, let me know that so I can fix that.
#2757

spinear:
+1 Please fix this or approve his pull request.

ChrissiQ:
This bug is quite game breaking - I am unable to export materials other than MAX.

I have a software division.  I want to keep a certain amount of AI cores in my warehouse because it contributes to the production multiplier.

In my game, in Aevum, MP of AI Cores is $11k, while in Sector-12 it is $52k, so obviously I would like to be able to export this material from Aevum to Sector-12 and sell it in Sector-12.

However, AI Cores contribute to the production multiplier, so I do not want to export or sell MAX - only prod.

I can't export PROD (that would be a great feature!!!), so instead I want to export a number less than prod, say 32, and sell PROD-32.

Since I cannot sell PROD-(amount), this is not possible, so I can't balance my sell and export.

We really need this fix!

Penumbra69:
[...]

PR #2757, opened 2022-01-22
Bug fix corporation expressions

ltoni90:
Fixed bug that caused an error when trying to set a dynamic expression to sell materials or products (e.g. PROD-500). [See Issue](https://github.com/danielyxie/bitburner/issues/2719)

**Changes**
- Added property "maxsll" to Material.ts and Product.ts
- The result of the calculation of maxSell will now be written to the property "maxsll" of the corresponding material or product
- Temporary quantity evaluation does not use "1" anymore. Instead it uses the effective production rate respectively the maximum amount of the product or material that can be sold.

**How it was tested**
Entered expressions where the evaluation leads to negative result and error message was displayed as expected.

### DIFF
--- a/src/Corporation/Product.ts
+++ b/src/Corporation/Product.ts
@@ -96,6 +96,8 @@ export class Product {
   marketTa2 = false;
   marketTa2Price: IMap<number> = createCityMap<number>(0);
 
+  // Determines the maximum amount of this product that can be sold in one market cycle
+  maxsll = 0
   constructor(params: IConstructorParams = {}) {
     this.name = params.name ? params.name : "";
     this.dmd = params.demand ? params.demand : 0;
--- a/src/Corporation/Product.ts
+++ b/src/Corporation/Product.ts
@@ -97,7 +97,7 @@ export class Product {
   marketTa2Price: IMap<number> = createCityMap<number>(0);
 
   // Determines the maximum amount of this product that can be sold in one market cycle
-  maxsll = 0
+  maxsll = 0;
   constructor(params: IConstructorParams = {}) {
     this.name = params.name ? params.name : "";
     this.dmd = params.demand ? params.demand : 0;
--- a/src/Corporation/Actions.ts
+++ b/src/Corporation/Actions.ts
@@ -109,8 +109,8 @@ export function SellMaterial(mat: Material, amt: string, price: string): void {
   if (amt.includes("MAX") || amt.includes("PROD")) {
     let q = amt.replace(/\s+/g, "");
     q = q.replace(/[^-()\d/*+.MAXPROD]/g, "");
-    let tempQty = q.replace(/MAX/g, "1");
-    tempQty = tempQty.replace(/PROD/g, "1");
+    let tempQty = q.replace(/MAX/g, mat.maxsll.toString());
+    tempQty = tempQty.replace(/PROD/g, mat.prd.toString());
     try {
       tempQty = eval(tempQty);
     } catch (e) {
--- a/src/Corporation/Actions.ts
+++ b/src/Corporation/Actions.ts
@@ -174,8 +174,8 @@ export function SellProduct(product: Product, city: string, amt: string, price: 
     //Dynamically evaluated quantity. First test to make sure its valid
     let qty = amt.replace(/\s+/g, "");
     qty = qty.replace(/[^-()\d/*+.MAXPROD]/g, "");
-    let temp = qty.replace(/MAX/g, "1");
-    temp = temp.replace(/PROD/g, "1");
+    let temp = qty.replace(/MAX/g, product.maxsll.toString());
+    temp = temp.replace(/PROD/g, product.data[city][1].toString());
     try {
       temp = eval(temp);
     } catch (e) {
--- a/src/Corporation/Industry.ts
+++ b/src/Corporation/Industry.ts
@@ -828,7 +828,7 @@ export class Industry implements IIndustry {
                   }
                 }
 
-                const maxSell =
+                mat.maxsll =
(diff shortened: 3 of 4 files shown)


src/Corporation/Product.ts
```diff
@@ -96,6 +96,8 @@ export class Product {
   marketTa2 = false;
   marketTa2Price: IMap<number> = createCityMap<number>(0);
 
+  // Determines the maximum amount of this product that can be sold in one market cycle
+  maxsll = 0
```
ChrissiQ, line 100:
```suggestion
  maxsll = 0;
```

ltoni90:
Is there any reason why this PR won't get approved? Can I do something to speed this up?

phyzical:
sadly you will just have to be patient :)

Penumbra69:
Sadly this is a game breaker.

phyzical:
@Penumbra69 passive aggressive responses like this wont help anything..

phyzical (commented):
LGTM, once theres, a test script and some sort of gif showing how it was tested i can approve.

(need to consider things like too small numbers, too big, -ves ect.

Penumbra69:
> @Penumbra69 passive aggressive responses like this wont help anything..

Neither is being told to "sadly" wait for a game breaking bug to be fixed when a PR is offered to fix it without any advice on timing or priority.

If your words, which I just echoed back, were seen as passive aggressive - what would that mean about yours?
Aggressively passive?! :D 

Anyhow - thanks for looking at the PR, I hope this bug fix makes it through the gauntlet soon so we can all re-enjoy this part of the game!

Thanks,
P

phyzical:
yes... bark and you will get barked back at

ltoni90:
Hey @phyzical ,

At the time where I opened this pull request, my corporation wasn't as advanced as of now and while creating the test GIFs I stumbled upon another problem regarding big floating numbers.

This are my current stats:
grafik

Which results in maxsll of 1969593915067574800. At this point expressions like MAX-(MAX+0.0001) or even MAX-(MAX+10) will eval to 0 and thus validating as true instead of false. I know it's a silly expression and also a highly advanced corporation I'm talking about here, but nonetheless the validation logic wouldn't work in this case. If you say that this is a no-go, I'll need more time to implement a way that satisfies such edgy cases. What do you think?

phyzical:
@ltoni90 ah damn this is why i like the idea for forcing a gif :) you end up wanting to test more than you thought and find those edge cases that just cant bee seen otherwise.

Given the time it will take to get into prod/dev anyway i think it would be reasonable to have another think about how to solve it. (simpler the better :) ) i feel like if there is a way to achieve this without storing at against the mat itself also would be better, just feel like it may be exploitable 😆 

there no rush :)

if i was to guess its toString messing with you, maybe theres some helper function in the codebase exactly for this  "convert number to string safely"

@pigalot do you have any suggestions (incase this just isnt the right way to solve this)

pigalot:
I would not bother trying to solve for very stupid edge cases like "MAX-(MAX+0.0001) or even MAX-(MAX+10)".
If they evaluate to 0 then that's fine it will just set it to 0, if the user wants too shoot them self in the foot then let them as long as it prevents negative numbers it doesn't matter.

ltoni90:
@pigalot @phyzical 
sounds reasonable for me to ignore some edgy cases. The only way I see right now to handle big numbers would be to use shifted BigInts. But as we need to evaluate mathematical expressions it would be necessary to compile the expression in some sort of math expression tree as point and dash operations have to be calculated in a special way (like [here](https://stackoverflow.com/a/66939244)) when working with shifted BigInts. And that's way to exaggerated for this use case....

@phyzical 
regarding your concerns about the possibilty of exploitation. The additional property on Material.ts and Product.ts is only used for the validation of a sell expression and won't be accessed afterwards (it only gets updated in each market cycle). And before selling the actual material or product the expression is getting evaluated again anyway. If the result should be negative than it will be set to 0 (see [here](https://github.com/danielyxie/bitburner/blob/ee887de151f00c508e4f37ba94885ceeaffa36fd/src/Corporation/Industry.ts#L870) or [here](https://github.com/danielyxie/bitburner/blob/ee887de151f00c508e4f37ba94885ceeaffa36fd/src/Corporation/Industry.ts#L1228)). Because of that I don't really believe that the property can be exploited in any way. I've also looked into some alternatives like calculating maxsll in Actions.ts, but that's not possible as some variables used to calculate maxsll are only available during [processMaterials()](https://github.com/danielyxie/bitburner/blob/ee887de151f00c508e4f37ba94885ceeaffa36fd/src/Corporation/Industry.ts#L523) or [processProdcuts()](https://github.com/danielyxie/bitburner/blob/ee887de151f00c508e4f37ba94885ceeaffa36fd/src/Corporation/Industry.ts#L993). So I think that the simplest solution to this problem is to create the maxsll property on Material or Product instance.

If that's okay for you I will create the test GIFs and post them here.

phyzical:
👍

ltoni90:
@phyzical here you go:

Test 1
Division: Real Estate
Type: Material - Real Estate
Expression: PROD-100
Expected Result: OK
Result: OK
Test_1


Test 2
Division: Real Estate
Type: Material - Real Estate
Expression: PROD-60000
Expected Result: FAIL
Result: FAIL
Test_2

Test 3
Division: Real Estate
Type: Product
Expression: PROD-100
Expected Result: OK
Result: OK
Test_3


Test 4
Division: Real Estate
Type: Product
Expression: PROD-31000
Expected Result: FAIL
Result: FAIL
Test_4

phyzical:
looking good GJ thanks for the gifs, i wonder if we should improve the error message too. also how about api support?

Issue #2732, opened 2022-01-21
'Find All Valid Math Expressions' contract description does not specify order of evaluation, unary minus


macdjord:
The description for the 'Find All Valid Math Expressions' contract type asks the player to "Return all possible ways you can add the +, -, and * operators to the string such that it evaluates to the target number". It does not specify what order of operations the player should assume.

My initial solution assumed that the generated expressions would be evaluated in plain left-to-right order like a calculator, e.g. `1*2+3*4` = `(((1*2)+3)*4)` = 20. This worked for some problems, but later failed. My next solution, which assumes that all multiplication is done first, then addition and subtraction are done left-to-right, and this has worked so far, so I'm assuming its the right answer.

Similarly, the way the contract is worded doesn't make clear whether unary minus is allowed, e.g. whether `2*-2` ('two times minus-two') is a valid way to get -4, or `-1+3` to get 2. I am assuming it's not, since my solver doesn't consider them and its solutions have been accepted, but the contract ought to make that clear.

zeddrak:
PR #2769

PR #2769, opened 2022-01-23
In response to Issue #2732

zeddrak:
This simply clarifies the description for the find all valid math expressions in response to issue #2732 
#2732


macdjord:
"(Normal order of operations applies.)" is less than helpful. Suggest actually specifying what that is, i,e, multiplication first, then addition and subtraction evaluated left-to-right.

PR #2844, opened 2022-01-27
MISC: Crimes can now be unfocused.

phyzical:
# Feature
Add optional focus to commitCrime via api
Added ability to "do something else" via the ui
converted `Work` over to a switch from a  series of if statements and added support for crime
# Testing
manually tested committing a crime, canceling and new do something else + re focus
also tested commitCrime from api with and without focus
also tested calling  commitCrime via the api while a crime was initiated via the ui
* add a 25% penalty when committing a crime unfocused
* refactored implementation of success rate to take "focus" external of player object to allow api usage/ static ui usage and make sleeves compatible
* changes the output of getCrimeChance to return a percentage rather than a flat 0 - 1 number
* moved focus + router logic into the Crime helper for simplier implementation
* hard coded sleeves to a "focus" true when calculating the success chance as not applicable

Apr-16-2022 15-18-31
Apr-16-2022 15-37-31


phyzical:
Also thought i should note, i attempted to handle the focus deeper down but i noticed there are disconnects between the data i.e in CharacterOverview.tsx you never seem to get the final focus = false for some reason, and based on other functions from the api context i.e createProgram they seemed to kind of just overwrite the focusness at the end

src/NetscriptFunctions/Singularity.ts
```diff
@@ -1300,7 +1300,15 @@ export function NetscriptSingularity(
         throw helper.makeRuntimeErrorMsg("commitCrime", `Invalid crime: '${crimeRoughName}'`);
       }
       workerScript.log("commitCrime", () => `Attempting to commit ${crime.name}...`);
-      return crime.commit(Router, player, 1, workerScript);
+      crime.commit(Router, player, 1, workerScript);
```
Faustvii, line 1303: This is not taking into consideration if there was focus before hand, so every time the commitCrime triggers it'll send you back to the terminal even if you aren't focusing or were looking at a different screen.

For instance see line 1278 on how it's implemented for working on a program.

src/NetscriptFunctions/Singularity.ts
```diff
@@ -1300,7 +1300,15 @@ export function NetscriptSingularity(
...
+      if (focus) {
+        player.startFocusing();
+        Router.toWork();
+      } else {
```
Faustvii, line 1307: This is not taking into consideration if there was focus before hand, so every time the commitCrime triggers it'll send you back to the terminal even if you aren't focusing or were looking at a different screen.

For instance see line 1278 on how it's implemented for working on a program.

phyzical: ah good point >< 👍

alainbryden:
I think fundamentally this might not be merged, because crimes are supposed to require focus. You aren't supposed to be able to personally do crimes in the background.

phyzical:
@alainbryden the same could be said about working for a company or going a gymWorkout  or going to a university though?

phyzical:
Granted i am only at BN4: and have not reeched sleeves so i understand the point of view, but it just feels silly as you still can only be doing one thing.

src/ui/React/CharacterOverview.tsx
```diff
@@ -132,51 +132,62 @@ function Work(): React.ReactElement {
...
+    case CONSTANTS.WorkTypeCreateProgram:
+      details = <>Coding {player.createProgramName}</>;
+      header = <>Creating a program</>;
+      innerText = (
+        <>
+          {player.createProgramName}{" "}
+          {((player.timeWorkedCreateProgram / player.timeNeededToCompleteWork) * 100).toFixed(2)}%
+        </>
+      );
+      break;
```
tyrope, line 185: Why are we changing the entire flow just to add an extra thing?

alainbryden:
> @alainbryden the same could be said about working for a company or going a gymWorkout or going to a university or creating a program though?
> 
> the reason i made the pr is when i am grinding to create a gang the game is a PIA to say "killall" and reboot if your other auto scripts if things go wrong

Maybe not the best place for me to be saying this, but note that a good solution to the problem you are having is to call `ns.tail()` before or after any calls to `ns.commitCrime` or other focus-stealing functions. This will ensure the tail window for your script is always visible (even if you accidentally closed it) and therefore can be easily killed you're struggling to exit your crime loop.

src/ui/React/CharacterOverview.tsx
phyzical: yeah ill roll back the change to a case statement back to if statements, its just more readable imo

phyzical:
@alainbryden perfectly valid workaround, sadly doesn't change the fact this singularity logic is missing this optional focus  that the rest seem to have

alainbryden:
If we are to go ahead with merging this PR, then I propose we first introduce a penalty to un-focused crime when the user does not yet have the augmentation "Neuroreceptor Management Implant" - similar to the penalty when working for factions or companies.

This penalty should have been put in place when the "do something else" option was added to gym/study (which previously required focus), but we missed the boat on that one for now.

phyzical:
we can always do that as part of this tbh, this one hasnt been approved yet so theres every chance it just closed.

But if we do want it im happy to add some penalties to those also

cc @hydroflame

DavidGrinberg:
Suggestion for the penalty: significantly reduced crime success change.

phyzical:
sure! what do we reckon it should be 70% as effective when unfocused?

alainbryden:
I believe the rep focus penalty is 75% (25% as effective) so we should be consistent with that maybe.

phyzical:
@alainbryden mind giving the changes another looksie, should be ready for review

src/PersonObjects/Player/PlayerObjectGeneralMethods.tsx
```diff
@@ -1496,13 +1496,14 @@ export function startCrime(
   chaExp: number,
   money: number,
   time: number,
+  focus: true,
```
Snarling, line 1499: focus:true in the type definitions of the params for startCrime... focus should be type defined as a boolean and set to a default value of true.

Snarling:
* 25% reduction in success chance is only a penalty until you can achieve a 125% success chance. To match up with other work tasks, any penalty should be applied to the actual rewards (money, karma, xp). It should also take into account Neuroreceptor Management Implant.
* Focus doesn't work well with crime because unlike other work tasks, crime is a discrete task. The player can just re-focus moments before the end of the crime and they avoid any penalty, since that's the only time any rewards are given out. Other tasks are continuous and the total applied penalty applies continuously only during the time the player is unfocused. 
* I'm also not sure why this needs to also include API breaking changes to getCrimeChance (changing it from a fractional chance to a percentage) 
* I personally don't think focus/unfocus should be added to crime.

phyzical:
>25% reduction in success chance is only a penalty until you can achieve a 125% success chance. To match up with other work tasks, any penalty should be applied to the actual rewards (money, karma, xp). It should also take into account Neuroreceptor Management Implant.

happy to change 

>Focus doesn't work well with crime because unlike other work tasks, crime is a discrete task. The player can just re-focus moments before the end of the crime and they avoid any penalty, since that's the only time any rewards are given out. Other tasks are continuous and the penalty applies for the entire duration the player is unfocused and nothing more.

yeah this is correct but i suppose i decided if the player wants to be smart about it so be it reward users for better time management, metawise  i look at it this way rob a store and "unfocus" pull out your phone to run more scripts and if you dont focus at the last second the store owner might knock the gun out of your hand reducing the chance of success

>I'm also not sure why this needs to also include API breaking changes to getCrimeChance (changing it from a fractional chance to a percentage)

i guess i dont see the point in returning a non % value when the ui displays it as such, users would just be using this value as a > value anyway

>I personally don't think focus/unfocus should be added to crime.

fair enough i just added this as spending 16 hours murdering to start a gang was boring AF

hydroflame:
Closed because wont work with new work system (plus implemented there)

PR #2845, opened 2022-01-27
Added the ability to exportGame via api

phyzical:
* added the ability to check if there is a bonus avaible via the api
* Added the ability to exportGame via api

# Feature
tested exportBonus before exporting returns true
tested exportBonus after triggering an export returns false
tested exportGame triggers the same functionality as the ui


src/NetscriptFunctions/Singularity.ts
```diff
@@ -250,7 +252,12 @@ export function NetscriptSingularity(
       workerScript.running = false; // Prevent workerScript from "finishing execution naturally"
       killWorkerScript(workerScript);
     },
-
+    backupSave: function (): void {
+      return exportGame(player)
+    },
+    backupSaveBonus: function (): boolean {
+      return canGetBonus();
+    },
```
nickofolas, line 260: These should probably call
```ts
helper.updateDynamicRam("backupSave", getRamCost(player, "backupSave"));
helper.checkSingularityAccess("backupSave")
```
and same for `backupSaveBonus`, to keep in line with the other Singularity functions.

nickofolas (changes requested):
Looks good, just one small suggestions for the Singularity access.

src/NetscriptFunctions/Singularity.ts
phyzical: ah yep! oops good spot 👍
i've never added a new function that needed ram

phyzical:
not sure why it fails, locally it does not fail to build...

PR #2848, opened 2022-01-27
[Balance] Add CalculateShareMult to Field Work and Security Work

### LINKED ISSUE #2708: Proposal: share() should affect Security/Field work too, but at lower scale
For example, 100% effective for "Hacking Contracts" (current formula), 50% effective for "Field work" and 30% effective for "Security work".

Rationale: all three jobs require hacking skill at some weight, so our protagonist can utilize his computer power to do them. But in other two jobs the computer is not the only thing is used, so it should boost at the lower scale.

Can be easy implemented:
```
export function CalculateShareMult(mult: number = 1.0): number {
  return CSM(sharePower, mult);
}
```
```
export function CalculateShareMult(power: number, mult: number = 1.0): number {
  const x = 1 + mult * Math.log(power) / (8 * Math.log(1000));
  if (isNaN(x) || !isFinite(x)) return 1;
  return x;
}
```
(and probably a field to adding new augmentations that affect the share multiplier bonus in the future)

SagePtr:
Add CalculateShareMult to Field Work and Security Work.
It boosts only hacking and intelligence skills (like it basically does in Hacking Work, as they are the only skills used there), but not the other skills. So if the player is low on hacking but high on combat stats it almost doesn't affect a lot.
Closes #2708

### DIFF
--- a/src/PersonObjects/formulas/reputation.ts
+++ b/src/PersonObjects/formulas/reputation.ts
@@ -25,26 +25,19 @@ export function getHackingWorkRepGain(p: IPlayer, f: Faction): number {
 export function getFactionSecurityWorkRepGain(p: IPlayer, f: Faction): number {
   const t =
     (0.9 *
-      (p.hacking / CONSTANTS.MaxSkillLevel +
-        p.strength / CONSTANTS.MaxSkillLevel +
-        p.defense / CONSTANTS.MaxSkillLevel +
-        p.dexterity / CONSTANTS.MaxSkillLevel +
-        p.agility / CONSTANTS.MaxSkillLevel +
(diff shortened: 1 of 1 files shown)


phyzical:
I thought based on the ticket it would only affect the stats that are directly related to "hacking" but this would affect them all? or am i miss understanding it

SagePtr (commented):
@phyzical it multiplies only p.hacking and p.intelligence, leaving the other skills outside the brackets.

phyzical:
Ah derp 😆 LGTM, but this would be one for higher powers to decide

phyzical (approved):

phyzical:
ah actually would this then need to be shown on any stat overviews? or does it call hook into these two functions

PR #2855, opened 2022-01-28
growthAnalyze (choice 1 / simple fix) - only merge one of these

zeddrak:
Implementation of the simplest fix for the growthAnalyze issue.
It will take longer to implement the others.
(Not tested, not carefully examined for knock-on effects)

### DIFF
--- a/src/Server/ServerHelpers.ts
+++ b/src/Server/ServerHelpers.ts
@@ -63,7 +63,7 @@ export function numCycleForGrowth(server: Server, growth: number, p: IPlayer, co
       BitNodeMultipliers.ServerGrowthRate *
       coreBonus);
 
-  return cycles;
+  return Math.min(server.moneyMax, cycles);
 }
 
 //Applied server growth for a single server. Returns the percentage growth
--- a/src/Server/ServerHelpers.ts
+++ b/src/Server/ServerHelpers.ts
@@ -75,7 +75,7 @@ export function processSingleServerGrowth(server: Server, threads: number, p: IP
   }
 
   if (server.moneyAvailable < 1) { //can grow a server reduced to $0
-    server.server.moneyAvailable = 1;
+    server.moneyAvailable = 1;
   }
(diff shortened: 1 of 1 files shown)


zeddrak:
The main point of this version of the fix is to be as simple as possible.
It's still up to the player to hand growthAnalyze a "correct" multiplier (ie, they should hand it infinity if they want to grow a $0 server - and growthAnalyze should handle that).
this fix also brings the actual grow functionality in-line with the behavior growthAnalyze reports.

zeddrak:
BTW: An alternative to using $1 (if that's too punishing) is to use a minimum percentage (like servers are always considered to have at least 0.001% of their max when growing) - this is effectively the same as capping the growth multiplier required to max a zeroed server to a given amount (ie, it'll never take a growth multiplier > 1e5 to cap a zeroed server...) But the simple truth is, even at $1, multiple grows can eventually recover the server given enough time.

The key point of this solution option is to avoid the server's current moneyAvailable as a dependency and thus avoid the irreversible (a+x)**(x*m) issue.

zeddrak:
(PS: The percentage option has potential for exploitation)

src/Server/ServerHelpers.ts
```diff
@@ -74,8 +75,11 @@ export function processSingleServerGrowth(server: Server, threads: number, p: IP
     serverGrowth = 1;
   }
 
+  if (server.moneyAvailable < 1) { //can grow a server reduced to $0
+    server.moneyAvailable = 1;
+  }
   const oldMoneyAvailable = server.moneyAvailable;
-  server.moneyAvailable += 1 * threads; // It can be grown even if it has no money
+  //server.moneyAvailable += 1 * threads; // It can be grown even if it has no money
```
phyzical, line 82: we might aswell just remove this

phyzical (commented):
LGTM,

will approve when test script and gif of it working

PR #2858, opened 2022-01-28
(ALPHA) growthAnalyze (choice 2 / less vodoo) - only merge one of these

zeddrak:
This implementation involves a simple (and poorly optimized) 50% partition search approximation methodology to determine the number of threads actually required to grow a server from one amount to another.

The advantages of this approach over option 1 is that it allows for the current mechanic of +$1/grow thread pre-multiplier.
The disadvantage is that it can't seamlessly replace the previous functionality (this is just a fact of the current implementation of grow).

The advantage over option 2 is just less math wizardry. Partition searches are a relatively entry level concept compared to using advanced approximation techniques. The disadvantage is that it's almost certainly slower.

### DIFF
--- a/src/Server/ServerHelpers.ts
+++ b/src/Server/ServerHelpers.ts
@@ -106,7 +106,51 @@ export function numCycleForGrowthTransition(server: Server, growth: number, p: I
  * @returns Number of "growth cycles" needed
  */
 export function numCycleForGrowthCorrected(server: Server, targetMoney: number, startMoney: number, p: IPlayer, cores = 1): number {
-  return 0; //left off here.
+ 	if (startMoney == server.moneyMax) { return 0; } //no growth possible, no threads needed
+	if (startMoney < 0) { startMoney = 0; } // servers "can't" have less than 0 dollars on them
+	if (targetMoney > server.moneyMax) { targetMoney = server.moneyMax; } // can't grow a server to more than its moneyMax
+
+	const growthMultiplier = Math.max(1.0, Math.min(targetMoney, targetMoney / startMoney)); //need a starting point, this is worst case grow multiplier
+
+	const adjGrowthRate = (1 + (CONSTANTS.ServerBaseGrowthRate - 1) / server.hackDifficulty); // adj exponential base for security
+	const exponentialBase = Math.min(adjGrowthRate, CONSTANTS.ServerMaxGrowthRate); //cap growth rate
+
+	const serverGrowthPercentage = server.serverGrowth / 100.0;
+	const coreMultiplier = 1 + ((cores -1) / 16);
+	const threadMultiplier = serverGrowthPercentage * p.hacking_grow_mult * coreMultiplier * BitNodeMultipliers.ServerGrowthRate; //total of all grow thread multipliers
+
+	let cycles = Math.log(growthMultiplier) / Math.log(exponentialBase) / threadMultiplier; //this is the completely naive cycle amt and is always >= the real cycles required
+	let cycleAdjust = 0.5 * cycles;
+
+	let overGrowth = 0;
+	while (cycleAdjust > 0.5) { //go until we get an overage of less than $1 or we're adjusting by less than half a thread
+		overGrowth = (startMoney + cycles) * Math.pow(exponentialBase, cycles * threadMultiplier) - targetMoney;
+		if (overGrowth < 0) { cycles += cycleAdjust; }
+		else if (overGrowth > 1) { cycles -= cycleAdjust; }
+		else { break; } //we're over by less than $1, return cycles
+		cycleAdjust *= 0.5; //basic 50% partition search
+	}
+
+	cycles = Math.ceil(cycles);
+  //might be worth doing some checks here +/- 1 thread just to make sure we didn't quit too early
+	return cycles;
+}
+
+/**
+ * This function calculates the number of threads needed to grow a server based on a pre-hack money and hackAmt
+ * (ie, if you're hacking a server with $1e6 moneyAvail for 60%, this function will tell you how many threads to regrow it
+ * PROBABLY the best replacement for the current ns.growthAnalyze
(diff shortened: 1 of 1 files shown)


src/Server/ServerHelpers.ts
```diff
@@ -66,6 +66,92 @@ export function numCycleForGrowth(server: Server, growth: number, p: IPlayer, co
   return cycles;
 }
 
+/**
+ * Replacement function for the above function that accounts for the +$1/thread functionality of grow
+ * with parameters that are the same (for compatibility), but functionality is slightly different.
+ * This function can ONLY be used to calculate the threads needed for a given server in its current state,
+ * and so wouldn't be appropriate to use for formulas.exe or ns.growthAnalyze (as those are meant to
+ * provide theoretical scenaarios, or inverse hack respectively). Players COULD use this function with a
+ * custom server object with the correct moneyAvailable and moneyMax amounts, combined with a multplier
```
tyrope, line 75: typo: multplier => multiplier

src/Server/ServerHelpers.ts
```diff
@@ -66,6 +66,92 @@ export function numCycleForGrowth(server: Server, growth: number, p: IPlayer, co
...
+ * correctly calculated to bring the server to a new moneyAvailable (ie, pasing in moneyAvailable 300 and x2
```
tyrope, line 76: typo: pasing => passing

src/Server/ServerHelpers.ts
```diff
@@ -66,6 +66,92 @@ export function numCycleForGrowth(server: Server, growth: number, p: IPlayer, co
...
+ * when you want the number of threads required to grow that particular server from 300 to 600), and this
+ * function would pass back the correct number of threads. But the key thing is that it doesn't just
+ * inverse/undo a hack (since the amount hacked from/to matters, not just the multiplier).
+ * The above is also a rather unnecessarily obtuse way of thinking about it for a formulas.exe type of
+ * application, so another function with different parameters is provided for that case below this one.
+ * Instead this function is meant to hand-off from the old numCycleForGrowth function to the new one
+ * where used internally for pro-rating or the like. Where you have applied a grow and want to determine
+ * how many htreads were needed for THAT SPECIFIC grow case using a multiplier.
```
tyrope, line 84: typo: htreads => threads

src/Server/ServerHelpers.ts
```diff
@@ -66,6 +66,92 @@ export function numCycleForGrowth(server: Server, growth: number, p: IPlayer, co
...
+ * Idealy, this function, and the original function above will be depreciated to use the methodology
```
tyrope, line 85: typo: Idealy => Ideally

src/Server/ServerHelpers.ts
```diff
@@ -66,6 +66,92 @@ export function numCycleForGrowth(server: Server, growth: number, p: IPlayer, co
...
+export function numCycleForGrowthTransition(server: Server, growth: number, p: IPlayer, cores = 1): number {
+  return numCycleForGrowthCorrected(server, server.moneyAvailable * growth, server.moneyAvailable, p, cores);
+}
+
+/**
+ * This function calculates the number of threads needed to grow a server from one $amount to a the same or higher $amount
+ * (ie, how many threads to grow this server from $200 to $600 for example). Used primarily for a formulas (or possibly growthAnalyze)
+ * type of application. It lets you "theorycraft" and easily ask what-if type questions. It's also the one that implements the
+ * main thread calculation algorith, and so is the fuinction all helper functions should call.
+ * It protects the inputs (so putting in INFINITY for targetMoney will use moneyMax, putting in a nagitive for start will use 0, etc.)
```
tyrope, line 101: typo: nagitive => negative

tyrope (commented):
Code looks good; caught a couple of typos in your comments though. ;)

zeddrak:
Just realized that numCycleForGrowthByHackAmt can be made to function similar to the current growthAnalyze function with the same parameter list. The main difference being that it would return the number of threads needed to reverse the hack amt as if the server were at moneyMax when hacked. Since this is the "worst case" (as well as the most common case), this would mean it returns too many threads for lower value servers, but this has the side effect of slowly growing the server (as the extra threads mean more growth than loss). This would make that function inappropriate for pro-rating, but would make it a very good choice for replacing the current growthAnalyze without breaking scripts (since parameter list is unchanged).

src/Server/ServerHelpers.ts
```diff
@@ -66,6 +66,109 @@ export function numCycleForGrowth(server: Server, growth: number, p: IPlayer, co
...
+ * @param server - Server being grown
+ * @param hackAmt - the amount hacked (total, not per thread) - as a decimal (like 0.60 for hacking 60% of available money)
+ * @param prehackMoney - how much money the server had before being hacked (like 200000 for hacking a server that had $200000 on it at time of hacking)
+ * @param p - Reference to Player object
+ * @returns Number of "growth cycles" needed to reverse the described hack
+ */
+export function numCycleForGrowthByHackAmt(server: Server, hackAmt: number, prehackMoney: number, p: IPlayer, cores = 1): number{
+	if (prehackMoney > server.moneyMax) prehackMoney = server.moneyMax;
+	const posthacMoney = Math.floor(prehackMoney * Math.min(1, Math.max(0, (1 - hackAmt))));
+	return numCycleForGrowthCorrected(server, posthacMoney, prehackMoney, p, cores);
```
dwRchyngqxs, line 152: I think that swapping the argument here was not correct: we want to know how to grow from a lower value to a greater value, but hacking reduces the money on a server so posthacMony < prehackMoney.
targetMoney should be prehackMoney as before.

zeddrak: You are correct... Don't code tired... ;)

zeddrak:
ok, parameters look better now...

phyzical:
tbh im in favour of the simple soultion.

PR #2864, opened 2022-01-29
Don't hide irrelevant materials if their stock is not empty and hide irrelevant divisions from Export

### LINKED ISSUE #2860: Warehouse full because of wrong export of "Metal"
I accidently exported "Metal" to an "Agriculture"-coporation.
Problem is that I can not get it away from storage.
There is no "Metal" entry on the list of the UI to export it back.

So it looks like a bug.
I guess I should not be able to export "Metal" to "Agriculture" anyway (but it was selected by default).

SagePtr:
Fixes #2860

### DIFF
--- a/src/Corporation/ui/Helpers.tsx
+++ b/src/Corporation/ui/Helpers.tsx
@@ -4,7 +4,7 @@ import { IIndustry } from "../IIndustry";
 // current industry.
 export function isRelevantMaterial(matName: string, division: IIndustry): boolean {
   // Materials that affect Production multiplier
-  const prodMultiplierMats = ["Hardware", "Robots", "AICores", "RealEstate"];
+  const prodMultiplierMats = ["Hardware", "Robots", "AICores", "RealEstate", "AI Cores", "Real Estate"];
 
   if (Object.keys(division.reqMats).includes(matName)) {
     return true;
--- a/src/Corporation/ui/IndustryWarehouse.tsx
+++ b/src/Corporation/ui/IndustryWarehouse.tsx
@@ -83,8 +83,10 @@ function WarehouseRoot(props: IProps): React.ReactElement {
   const mats = [];
   for (const matName of Object.keys(props.warehouse.materials)) {
     if (!(props.warehouse.materials[matName] instanceof Material)) continue;
-    // Only create UI for materials that are relevant for the industry
-    if (!isRelevantMaterial(matName, division)) continue;
+    // Only create UI for materials that are relevant for the industry or in stock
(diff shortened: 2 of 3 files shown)


src/Corporation/ui/Helpers.tsx
```diff
@@ -4,7 +4,7 @@ import { IIndustry } from "../IIndustry";
 // current industry.
 export function isRelevantMaterial(matName: string, division: IIndustry): boolean {
   // Materials that affect Production multiplier
-  const prodMultiplierMats = ["Hardware", "Robots", "AICores", "RealEstate"];
+  const prodMultiplierMats = ["Hardware", "Robots", "AICores", "RealEstate", "AI Cores", "Real Estate"];
```
phyzical, line 7: hmm why was this needed? shouldn't we be standardizing the mats somewhere already?

src/Corporation/ui/IndustryWarehouse.tsx
```diff
@@ -83,8 +83,10 @@ function WarehouseRoot(props: IProps): React.ReactElement {
   const mats = [];
   for (const matName of Object.keys(props.warehouse.materials)) {
     if (!(props.warehouse.materials[matName] instanceof Material)) continue;
-    // Only create UI for materials that are relevant for the industry
-    if (!isRelevantMaterial(matName, division)) continue;
+    // Only create UI for materials that are relevant for the industry or in stock
+    const isInStock = props.warehouse.materials[matName].qty > 0;
+    const isRelevant = isRelevantMaterial(matName, division);
+    if (!isInStock && !isRelevant) continue;
```
phyzical, line 89: wouldn't this stop people from wanting to make an export, say they have 0 but want it to export next tick?

SagePtr: This is applied to unrelevant materials only. People won't have this kind of materials in their warehouses anyway unless they accidentally exported them from another division by mistake.

src/Corporation/ui/Helpers.tsx
SagePtr: Yes, it's subject to refactor some code, i did only the quick bugfix

phyzical: sorry i mean Ai Cores and Rel Estate are already in the array? but you have added them again?

src/Corporation/ui/IndustryWarehouse.tsx
phyzical: ah good point 👍

Issue #2718, opened 2022-01-20
Gang member tasks don't update in gang screen when changed through script


tvanderpol:
**Game Version**: Bitburner v1.4.0
**Steps to reproduce**:

- Run a script that changes a gang member's task on a timer
- Go to gang screen

**Expected:** See the task that the member is assigned to change as the script changes it
**Actual:** The stats (money, respect, wanted level, total respect) and the description seem to change, but the actual task in the drop down does not.

zeddrak:
I have also noticed this, though I don't know that updating the actual dropdown is a good solution (anyone tested what this would do if you had the drop down open when a script update happened).
I did notice that the descriptions to the right update correctly. So perhaps just adding a label with the "current task" listed below the dropdown that updates with api calls (and manual dropdown selection,)

zeddrak:
I have uploaded a potential fix (see link above this comment), but I need someone to test it for me (I haven't got Node working yet). I'll create the pull request once someone has confirmed this fix works as desired.

phyzical:
just wanted to add i tried to fix this on the dropdown directly 👍

PR #2868, opened 2022-01-29
(Not Tested) Show current task on gang management screen

zeddrak:
Possible fix for #2718
I haven't been able to test it, so please have someone test before implementing.
(To test, simply open the gang management page and see if it looks ok)


phyzical:
hmm im not sure this is needed though is it? you can infer the current task from the dropdown right?

PR #2869, opened 2022-01-29
Fix ps documentation

### LINKED ISSUE #2854: Minor copy/paste issue in documentation for NS.ps() method
The documentation for the ps method has a small copy/paste error in the NS2 example section.

The example is:
```
// NS2:
const ps = ns.ps("home");
for (let script of ps) {
    ns.tprint(`${script.filename} ${ps[i].threads}`);
    ns.tprint(script.args);
}
```
The NS2 example does not iterate with a `for (let i=0;...)` loop, so it should be:
```
const ps = ns.ps("home");
for (let script of ps) {
    ns.tprint(`${script.filename} ${script.threads}`);
    ns.tprint(script.args);
}
```

Thank you ever so much for this great game!!!

SagePtr:
It fixes two errors:
1) NS1 doesn't support ES6 let/const
2) NS2 version wasn't refactored properly

Closes #2854

### DIFF
--- a/src/ScriptEditor/NetscriptDefinitions.d.ts
+++ b/src/ScriptEditor/NetscriptDefinitions.d.ts
@@ -5017,8 +5017,8 @@ export interface NS extends Singularity {
    * @example
    * ```ts
    * // NS1:
-   * const scripts = ps("home");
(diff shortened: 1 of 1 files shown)


phyzical (approved):
LGTM

PR #2873, opened 2022-01-29
Add user-friendly invalid argument number message to purchaseServer

### LINKED ISSUE #2415: Misleading error message for Nescript function missing required argument
# Reproduction
* Create `tmp.script`:
  ```javascript
  purchaseServer("test-serv")
  ```
* Execute `run tmp.script`

# Expected result
Error message informing me that `purchaseServer()` is missing the required argument `ram`.

# Actual result
```text
RUNTIME ERROR
tmp.script@home

purchaseServer: ram should be a number
```

# Notes
I originally discovered this when I tried to call `purchaseServer(8)`, having briefly forgotten that `purchaseServer()` requires a server name as well as the amount of RAM. I got the same error message there, as well, which made it even more confusing - 8 was clearly a number, so what was it complaining about?

SagePtr:
Fixes #2415

### DIFF
--- a/src/NetscriptFunctions.ts
+++ b/src/NetscriptFunctions.ts
@@ -1628,6 +1628,7 @@ export function NetscriptFunctions(workerScript: WorkerScript): NS {
       return cost;
     },
     purchaseServer: function (aname: any, aram: any): any {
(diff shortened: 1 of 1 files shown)


phyzical (approved):
LGTM, will approve when gif showing it working

PR #2875, opened 2022-01-30
fix: electron `windowTracker` crash when reloading and killing scripts

smolgumball:
## resolves crash in `electron/windowTracker.js` during reload

* currently crashes when running `Reloads -> Reload & Kill All Scripts` menu command
* this fix adds extra checks to window position tracking code to ensure Electron window handles are valid before attempting to access their properties

**Before fix**
image

**After fix:**
l6fet9gVOI

### DIFF
--- a/electron/windowTracker.js
+++ b/electron/windowTracker.js
@@ -29,6 +29,11 @@ const windowTracker = (windowName) => {
   };
 
   const saveState = debounce(() => {
+    if (!window || window.isDestroyed()) {
+      log.silly(`Saving window state failed because window is not available`);
+      return;
+    }
+
     if (!windowState.isMaximized) {
       windowState = window.getBounds();
     }
@@ -41,7 +46,7 @@ const windowTracker = (windowName) => {
 
   const track = (win) => {
     window = win;
-    ['resize', 'move', 'close'].forEach((event) => {
+    ["resize", "move", "close"].forEach((event) => {
       win.on(event, saveState);
     });
   };


phyzical (approved):
LGTM, will approve when gif showing what the new result looks like.

smolgumball:
@phyzical New result would be a blank screenshot of my desktop :) Did you still need that?

phyzical:
ah i assumed it would resolve the issue an reload the scripts safely, or it just more about handling the issue more gracefully? 

if so yeah as silly as it sounds it will atleast show you've tested it properly :) (not that i dont trust you havnt :P )

smolgumball:
@phyzical Added screenshot to original issue

You mentioned "resolve the issue [and] reload the scripts safely", but I believe the desired outcome from the `Reload & Kill All Scripts` command is that the game reboots with no running scripts. This was the behavior previously, and still is. 

This fix is solely focused on addressing a regression due to dc2f00057aa298e30cd864db65ace15f7fad3eb0

phyzical:
thanks for the picture, i actually was getting this the other day on the dev branch. buuuut i think it actually has since been fixed.

is there any chance you can try to replicate it on the latest dev? (make sure you rebase dev properly ect :))

smolgumball:
@phyzical Still happening on latest `dev`

Current:
image

After merging 3431997315e0e3591409ac66c06a7be289f027d2:
image

phyzical:
coolies thx

PR #2876, opened 2022-01-30
[Feature] Script Editor Theme Customization

nickofolas:
This PR implements a system for customizing the theme of the included Monaco editor via a modal accessed from the options menu.

firefox_SN2smrH575

The customizer works by designating a new theme `customTheme`, which is nearly the same as the vscode dark theme by default. `customTheme` is now exposed to the user via the theme editor modal, which allows for the theme to be adjusted very similarly to the way the game theme is changed.

The editor theme data is stored in a new attribute in the save file, `EditorTheme`.

To protect against invalid color data, a sanitization function is executed whenever the editor is mounted, as well as whenever the options window is saved/closed. This replaces any invalid values with a valid placeholder value (i.e. bright red for invalid colors).

### Testing
- [x] Tested loading an old save file into the new version
- [x] Tested the various theme-able options to ensure that actually work
- [x] Tested with invalid color inputs

### More Screenshots
image
image
image
image


smolgumball:
nick the UI ninja strikes again. this is awesome!

total spitball scope-creep [don't hate me] suggestion: 
* add a button to convert current bitburner theme into an editor theme

not sure if the editor theme keys line up to global theme keys in any sensible way but 🤷

nickofolas:
thanks, i'm glad you're a fan! :)

deriving an editor theme from the global theme was something i had thought about and wanted to do, but, as you said, there's pretty much no reliable correlation between the keys :/

a bandage solution could be something like slapping the global `backgroundprimary` and friends onto general ui stuff, but im not sure if it's worth adding a shortcut for something so easy to do manually

phyzical:
may also fix/resolve https://github.com/danielyxie/bitburner/issues/2927

src/ScriptEditor/ui/ScriptEditorRoot.tsx
```diff
@@ -893,7 +895,11 @@ export function Root(props: IProps): React.ReactElement {
         </Box>
         <OptionsModal
           open={optionsOpen}
-          onClose={() => setOptionsOpen(false)}
+          onClose={() => {
+            sanitizeTheme(Settings.EditorTheme);
```
phyzical, line 899: should we be saving it when we close?

src/ScriptEditor/ui/ThemeEditorModal.tsx
```diff
@@ -0,0 +1,276 @@
...
+              setThemeCopy(defaultMonacoTheme);
+              rerender();
+            }}
+            startIcon={<HistoryIcon />}
+          >Reset to default</Button>
+        </Box>
+      </Paper>
+    </Modal>
+  )
+}
```
phyzical, line 276: nit: missing a lint?

phyzical (approved):
LGTM but to be honest im not super familiar with this code.

src/ScriptEditor/ui/ScriptEditorRoot.tsx
nickofolas: I think it's fine to save on close, since the user already clicks "save" within the theme editor modal, so it would be odd for the changes to not be applied.

Beyond that, the theme is saved in the save object when the theme is saved from within the editor modal, so not applying it on close would lead to an odd situation where the theme is applied next time the editor is loaded.

src/ScriptEditor/ui/ThemeEditorModal.tsx
nickofolas: Fixed.

PR #2878, opened 2022-01-30
Fix for ui of gang members current task when set via api

### LINKED ISSUE #2149: Gang dropdown needs to keep up with gang task.

### LINKED ISSUE #2718: Gang member tasks don't update in gang screen when changed through script

phyzical:
currently when using the api to set a members task it is updating everything but the dropdown of the current selected task, Note: this is only an issue when setting the task while you have the gang menu open.

# Bug fix
fixes #2718 
fixes #2149
* Added check to manually update interal state of dropdown on change of a members task via api

Tested by setting a gang members task manually then running 

```
sleep ns.sleep(5000)
ns.gang.setMemberTask("hghgm", "Deal Drugs" )
```

Before: you see the description of the task update but not the dropdown itself
After: the dropdown reflects the current task correctly

### DIFF
--- a/src/Gang/ui/TaskSelector.tsx
+++ b/src/Gang/ui/TaskSelector.tsx
@@ -21,6 +21,13 @@ export function TaskSelector(props: IProps): React.ReactElement {
   const gang = useGang();
   const [currentTask, setCurrentTask] = useState(props.member.task);
 
+  const contextMember = gang.members.find(member => member.name == props.member.name)
+  if (contextMember &&
+    contextMember.task != currentTask
+  ) {
+    setCurrentTask(contextMember.task)
+  }
+
   function onChange(event: SelectChangeEvent<string>): void {
     const task = event.target.value;
     props.member.assignToTask(task);


phyzical:
There may be a better way to go about this... though so if anyone has a better idea, i will happly change it :)

zeddrak:
Have you tested what it does if you have the dropdown open at the time of a script update?
(ie)
Set Manually
run script (with longer wait if needed)
Open dropdown and wait for script to update it.

phyzical:
oooo that's a good thought, looks like we are good though.


Animation

phyzical:
and if i was to wager a guess as to why this happens/ is needed, maybe its the whole gang being an array of members , from personal experience react doesn't like objects of arrays being updated, rather it doesn't know that the item inside changed only the amount of items changes. i cant 100% remember what caused this in the past for me just to be careful with arrays of objects being passed down to child components

Penumbra69 (approved):

PR #2880, opened 2022-01-30
Adding support for text/select options in Prompt command

PhilipArmstead:
Adds `options` parameter to `ns.prompt` function which allows the user to override the default Boolean display with Text and Select prompts. Select prompts require the `options` property from which to build the choices.

```js
const serverIndex = await ns.prompt('Select a server to target', { 
	type: 'select',
	options: ns.scan(), // options can be an array
 })

const scripts = { 'Hack': 'hack.js', 'Grow': 'grow.js', 'Weaken': 'weaken.js' }
const scriptKey = await ns.prompt('Select a script to run', {
	type: 'select',
	options: scripts, // options can be an object
})

const threadCount = await ns.prompt('How many threads should be used?', { 
	type: 'text',
 })
```


smolgumball (changes requested):
hey phil, looking good! 

i think you'll want to also update the [NetscriptDefinitions](https://github.com/danielyxie/bitburner/blob/dev/src/ScriptEditor/NetscriptDefinitions.d.ts#L5880-L5893) so in-game autocomplete and TS users get info about the API changes

phyzical:
LGTM, will approve with a gif showing them working.

one suggestion if your open/up to it. maybe we should make 3 distinct prompt components and 3 distinct functional calls to make it suimplier for the end user?

PhilipArmstead:
> LGTM, will approve with a gif showing them working.
> 
> one suggestion if your open/up to it. maybe we should make 3 distinct prompt components and 3 distinct functional calls to make it suimplier for the end user?

What did you have in mind for the names of the separate functions? `prompt`, `promptText` and `promptSelect` or something? For the record, I don't think those are good names. Having one prompt function which can be configured makes the most sense to me, but I'm not a good judge of these matters. I will yield to the wishes of the maintainers.

phyzical:
@PhilipArmstead yeah just a thought (i would have done the names as you had guessed) from a user perspective, your implementation looks to work fine. 

maybe leave it as is and others can chime in if they think otherwise

Master-Guy (changes requested):
LGTM, but keep #3156 in mind please.

PR #2881, opened 2022-01-30
growthAnalyze (choice 2 / documented math) - only merge one of these

dwRchyngqxs:
This implementation involves an initial approximation step then a Newton-Raphson iteration to determine the number of threads actually required to grow a server from one amount to another.

The advantages of this approach over option 2 / less voodoo is that it should be more efficient (to be tested, I don't know how to compare perf).
The disadvantage is that the code isn't easy to understand because there is some non-trivial math involved. I did my best effort to explain the math and document the choices though.

The log provided shows the results the algorithm spits (before capping to the number of threads used) on two savefiles only running weaken/grow/hack loops (standard use I suppose). It took at most 4 refinement iterations to give a result in my tests.
[console-export-2022-1-31_0-19-48.txt](https://github.com/danielyxie/bitburner/files/7967173/console-export-2022-1-31_0-19-48.txt)
[console-export-2022-1-31_0-21-13.txt](https://github.com/danielyxie/bitburner/files/7967174/console-export-2022-1-31_0-21-13.txt)


trhr:
this = a, w'(e) = s((o + m) * e^(w*o) - r)/k

dwRchyngqxs:
I did not use the correct value as exit condition for the refinement loop, this should be fine now (side-effect: less refinement iterations are performed)

dwRchyngqxs:
> this = a, w'(e) = s((o + m) * e^(w*o) - r)/k

What do you mean? Are the comment cryptic or too notations heavy?

dwRchyngqxs:
This PR is not ready to be merged yet. There have been changes on the branch I implemented my changes on (PR #2858). I won't be able to rebase until this WE.

src/Server/ServerHelpers.ts
```diff
@@ -66,6 +66,199 @@ export function numCycleForGrowth(server: Server, growth: number, p: IPlayer, co
...
+	 * n * ln(b) * t * b^(o*t) = w*exp(w), Thus w = W(n * ln(b) * t * b^(o*t))
+	 * Finally we invert the variable change: c = W(n * ln(b) * t * b^(o*t))/(ln(b)*t) - o
+	 *
+	 * There is still an issue left: b^(o*t) doesn't fit inside a double precision float
+	 * because the typical amount of money on servers is arround 10^6~10^9
+	 * We need to get an approximation of W without computing the power when o is huge
+	 * Thankfully an approximation giving ~30% error uses log immediately so we will use
+	 * W(n * ln(b) * t * b^(o*t)) ~= log(n * ln(b) * t * b^(o*t)) = log(n * ln(b) * t) + log(exp(ln(b) * o * t))
+	 * = log(n * ln(b) * t) + ln(b) * o * t
+	 * (thanks to Drak for the grow formula, f4113nb34st and Wolfram Alpha for the rewrite, dwRchyngqxs for the explanation)
```
xmodar, line 156: I have implemented [`WLog(x) = W(ln(x))`](https://github.com/howion/lambert-w-function/issues/6) which allows us to handle `b^(o*t)` with ease. So `W(n * ln(b) * t * b^(o*t))` becomes `WLog(ln(n * ln(b) * t) + o * t * ln(b))`.

dwRchyngqxs: I will reuse [your code](https://gist.github.com/xmodar/baa392fc2bec447d10c2c20bbdcaf687) of the initial approximation given in [Iacono, R., Boyd, J.P.](https://doi.org/10.1007/s10444-017-9530-3) (20). I doubt I will use their iteration on lambert W though because the code is intended to be read by people and maybe teach them something, like using results from research articles and a classical algorithm (Newton's iteration).

xmodar: My very first solution to this problem was actually Newton's method. Then, I attempted using an arbitrary-precision floating-point library like [mpmath](https://mpmath.org/doc/current/functions/powers.html?highlight=lambert#mpmath.lambertw). However, I found [the paper](https://doi.org/10.1007/s10444-017-9530-3) and their solution seems to work well but suffers from the same issue. Therefore, I thought about the log scale and had to derive this implementation myself. The idea behind it is to leverage the [LogSumExp trick](https://en.wikipedia.org/wiki/LogSumExp#:~:text=log-domain%20calculations) to define `log1Exp(x) = ln(1 + exp(x))`. The rest is a bunch of careful substitutions to reach the final form.

P.S. I just update [the gist](https://gist.github.com/xmodar/baa392fc2bec447d10c2c20bbdcaf687) to improve handling boundary conditions.

PR #2886, opened 2022-01-31
QoI: Increase HacknetNodeElem name colSpan

DrCuriosity:
Spanning `{node.name}` across all three table columns in the Hacknet Node UI uses a bit less screen real estate, so that on larger font sizes you don't end up with e.g. "hacknet- " on one line and "node-11" on the next.

image

### DIFF
--- a/src/Hacknet/ui/HacknetNodeElem.tsx
+++ b/src/Hacknet/ui/HacknetNodeElem.tsx
@@ -172,7 +172,7 @@ export function HacknetNodeElem(props: IProps): React.ReactElement {
       <Table size="small">
         <TableBody>
           <TableRow>
-            <TableCell>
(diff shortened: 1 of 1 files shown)


phyzical:
LGTM, will approve with gif showing how it looks when resizing

DrCuriosity:
Thanks @phyzical, here's a side-by-side comparison with the change made to the leftmost column only:
image

Obviously there are some other bits of squished layout elsewhere in the table cells, but it at least helps to preserve the node name for longer without breaking it across lines. It expands into empty table cell space that's not being used for anything else, so there's no risk of the change blowing layouts out any further than they already are.

phyzical:
thanks for the pic 👍

phyzical (approved):

PR #2891, opened 2022-02-01
Terminal: There is now an interface definition for the autocomplete() 'data' argument.

tigercat2000:
This is just typescript stuff, all it should do is allow users to have editor autocompletion for the different attributes on `data`.
Usage is as follows,
```javascript
/**
 * @param {AutocompleteData} data
 * @param {string[]} args
 */
export function autocomplete(data, args) {
    return data.servers;
}
```


phyzical:
LGTM, will approve with a  gif showing it working, maybe some examples of other autocompletes being unaffected by changes

phyzical (approved):
thanks for the gif :)

PR #2895, opened 2022-02-01
Prevent Sleeves from Getting Netburner Augs

waffleattack:
quick and dirty fix, It works, but could probably be better


phyzical:
could you please add test scripts and a gif testing via the ui and api of it working?

hydroflame:
Really the check should be that we verify if the aug even has any effect on the sleeve.

Aka does it increase any of the stats or any of the exp

PR #2913, opened 2022-02-04
MISC: FIX #1378,#2918 refactored applyFor implementations to return IApplyForJob

### LINKED ISSUE #1378: refactor applyFor* functions to return [gotJob, reason]

### LINKED ISSUE #2918: Typo when stats do not match job requirements
Reads Unforunately where it should be Unfor_t_unately.

phyzical:
# BugFix

* added new interface object for response
* attempted to replicate optional message and default to original messages to preserve functionality

fixes #1378
fixes #2918


phyzical:
@hydroflame im not super familiar with TS part of the reason for contributing, (plus i find refactoring almost as fun as play the game itself 😄 

few things i was unsure about
* location of where to put the new interface i assume types is the valid location? 
* didn't change the response of the applyToCompany as was unsure if this is the ultimate change desired, if so just let me know and ill update that too

PR #2914, opened 2022-02-05
Fix research tree of product industries post-prestige

pd:
After destroying a BitNode, the `resetIndustryResearchTrees` reset all
industries to the _base_ research tree, leaving the product industries
without the `uPgrade` researches in the graph. So corporations started
in the next BitNode were unable to expand product capacity.

This fixes that reset function to use the correct product industry tree.

## Reproduce

Dev menu, formed a corp with a software division named `sw` and looked at the research tree. All is well:

image

Hack daemon to prestige and repeat the corp formation:

image

Fulcrum and below are gone.

## After

On this branch, same process and fulcrum is back:

image


phyzical (approved):
good spot 👍 
LGTM

PR #2915, opened 2022-02-05
add list of available crimes to crime details

wdpk:
added list of available crimes as of 1.4.0 and their descriptions


hkgnobody:
These names matches Singularity.commitCrime() method, somehow it dont match even i match those roughName defined in CrimeHelpers.findCrime()

phyzical (approved):
LGTM

PR #2916, opened 2022-02-05
MISC: Add websockets to interact with the filesystem

lordducky:
This is a cut at a Websocket server for bitburner.
---
The makes the Bitburner game operate as a Websocket client which connects to a websocket server (maintained by the IDE connection) Operating as a client allows both web and electron games to use an external ide editor, bringing this feature to the web game as well.

I have a working non-extension version of this server working as a prototype and testing field (still has a few bugs) that is relatively stable for the most part.

Documentation for integration is in a [`Readme.md`](https://github.com/danielyxie/bitburner/blob/a59f121280bdde9b6714d1a14f3f72965d189223/src/APIServer/README.md) that is included in the `src/APIServer` folder. This file can be moved wherever appropriate but included it there for now since I am unsure of the proper location.


phyzical:
@lordducky so to confirm the idea behind this is to allow files to be sent to the game from the editor? 

just wondering why isn't the current vscode integration good enough?

or is this more about making it generic for other IDE's?

lordducky:
@phyzical 
**1. Allows for the game to utilize VSCode**
I'm working on updating the VSCode integration to be able to utilize this method, but ideally any version of the game, whether electron or in browser can use this method to integrate with and IDE.

That's because right now, running a http server is something that is reserved for an application, and browsers only operate as a client. The current model where the game acts as the server and the IDE acts as a client is backwards architecturally. 

The game running in the browser has the tightest security sandbox and the least amount of features, while IDEs and other custom setups have a much wider breadth of functions available. Feature development wise, its much easier to use the browser as client.

2. It allows for 2 way communication. So not only can the VSCode integration push code, but the game can also send code back to the VSCode integration.

This allows for things such as creating a file  on bitburner writing to files in game. Those files would not be synced VSCode 

Alternatively, VSCode or another ide could conceivably poll for such changes, but that would still require a two way communication method which is not possible if we want to support IDEs for web browsers.

phyzical:
hey thanks for the overview,

+1 for the two way comms (though itll be interesting to see how that goes given some peoples scripts) 

is there any prs vscode integration side to link the two up?

lordducky:
Not at the moment. Still working on it. Will link the pr here when ready.

phyzical:
no problem, no rush.
 there is a few spring cleaning tasks that must be performed before prs will start being merged TBH (so no ETA)

lordducky:
Sure would also add that this does not interfere with the server API that exists at the moment. They can coexist without issue.

phyzical:
ah coolies, is it in a testable state? the two way comms? (might give this one a go myself) 

or do you mean the code doesnt come into effect until the vscode integration is updated?

PR #2920, opened 2022-02-05
Issues with log windows and running same script from terminal

### LINKED ISSUE #1775: Tail window won't pick up new run when running the same script from terminal.

phyzical:
# Bug fix
* added logic to close log if it already had a window open

Wasn't sure if this is the desired result.

Another option i stumbled across was to simply not early out and this will instead open a new log window each time its called. (though there was lingering PID issue with this approach so theres additional logic required for this approach)

If we would prefer that i can redo this PR 👍 

i also tried playing aroundthe tail function specifically but it didnt get anywhere

fixes #1775
Mar-17-2022 21-33-56

### DIFF
--- a/src/ui/React/LogBoxManager.tsx
+++ b/src/ui/React/LogBoxManager.tsx
@@ -145,7 +145,7 @@ function LogWindow(props: IProps): React.ReactElement {
       const server = GetServer(script.server);
       if (server === null) return;
       const s = findRunningScript(script.filename, script.args, server);
-      if (s !== null) {
+      if (s !== null && script != s) {
         setScript(s);
       }
       setPoll(!poll)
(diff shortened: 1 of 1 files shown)


Master-Guy:
Why would you close the existing log file, instead of re-using it?
It would have my preference to append to an existing log, just like hitting the "Run" button in the existing Log button.

phyzical:
its just not possible to preserve the logs due to how killscript works in its corrent implementation

hydroflame:
Yeah not quite the best way to solve this.

`WorkerScriptStartStopEventEmitter.ts` This file contains an event emitter.

The log box needs to subscribe to that event and when it receives something it need to check if the scripts it's representing is still active.

Closing because essentially no lines in this PR will be in common with the other solution.

PR #2923, opened 2022-02-05
Update ls to use --grep like ps

### LINKED ISSUE #1944: ls -l

phyzical:
# Bug fix
updated grep for ls
* implemented flags logic for grep
* added --grep and -g
* updated docs

fixes #1944

image


phyzical:
Ah there is one regression still to fix (use via api seems to add the filename as the directory oops)
image

PR #2925, opened 2022-02-06
Necro script fix

Ornedan:
Script death is signaled by throwing that script's `WorkerScript` instance. This thrown object can pass through player code, where the player can modify the script's state, and any parts of the game accessible through a `WorkerScript`.

Sample code to illustrate the problem, a simple exploit that restores the script back to running status:

```js
/** @param {NS} ns **/
  debugger;

  function nekromanteia(f, ...args) {
    try {
      return ns[f](...args);
    }
    catch(e) {
      if(e && e.env && e.env.stopFlag != null) {
        e.env.stopFlag = false;
        ns.tprintf("Call to %s failed due to script death, raised script back from the dead", f);
      }
    }
  }

  ns.exit();

  nekromanteia("tprint", "Arise"); // Whatever throws when called in a dead script is fine

  ns.tprint("Braains");
}
```

The problem is solved by instead throwing instances of a dedicated `ScriptDeath` type, that does not carry direct references to game engine objects.

After the fix the sample exploit no longer works.

### DIFF
--- a/src/NetscriptEvaluator.ts
+++ b/src/NetscriptEvaluator.ts
@@ -4,8 +4,10 @@ import { ScriptDeath } from "./Netscript/ScriptDeath";
 import { WorkerScript } from "./Netscript/WorkerScript";
 
 export function netscriptDelay(time: number, workerScript: WorkerScript): Promise<void> {
-  if (workerScript.delayReject)
-    workerScript.delayReject(new ScriptDeath(workerScript));
+  // Cancel any pre-existing netscriptDelay'ed function call
+  // TODO: the rejection almost certainly ends up in the uncaught rejection handler.
+  //       Maybe reject with a stack-trace'd error message?
+  if (workerScript.delayReject) workerScript.delayReject();
 
   return new Promise(function (resolve, reject) {
     workerScript.delay = window.setTimeout(() => {
--- a/src/NetscriptEvaluator.ts
+++ b/src/NetscriptEvaluator.ts
@@ -1,15 +1,18 @@
 import { isString } from "./utils/helpers/isString";
 import { GetServer } from "./Server/AllServers";
+import { ScriptDeath } from "./Netscript/ScriptDeath";
 import { WorkerScript } from "./Netscript/WorkerScript";
 
 export function netscriptDelay(time: number, workerScript: WorkerScript): Promise<void> {
-  if (workerScript.delayReject) workerScript.delayReject();
+  if (workerScript.delayReject)
+    workerScript.delayReject(new ScriptDeath(workerScript));
+
   return new Promise(function (resolve, reject) {
     workerScript.delay = window.setTimeout(() => {
       workerScript.delay = null;
       workerScript.delayReject = undefined;
 
-      if (workerScript.env.stopFlag) reject(workerScript);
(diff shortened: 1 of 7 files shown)


src/NetscriptWorker.ts
```diff
@@ -525,29 +527,30 @@ function createAndAddWorkerScript(
...
     }
   }
 
   // Once the code finishes (either resolved or rejected, doesnt matter), set its
   // running status to false
-  p.then(function (w: WorkerScript) {
-    w.running = false;
-    w.env.stopFlag = true;
+  scriptExecution.then(function (w: WorkerScript) {
+    if(w !== workerScript) console.error("!BUG! Wrong WorkerScript instance !BUG!")
```
phyzical, line 551: How about 
```suggestion
    if(w !== workerScript) console.error("!BUG! Wrong WorkerScript instance !BUG! (This means we missed a reference somewhere)")
```

just to help the next person if this gets accepted

phyzical, line 551: How about 
```suggestion
    if(w !== workerScript) console.error("!BUG! Wrong WorkerScript instance !BUG! (This means we missed a reference somewhere)")
```

just to help the next person

Ornedan:
> i didn't personally test it, but code wise and concept wise seems logical.
> 
> my only concern would be if/when we wanted to add "pause" id assume this change would make that impossible due to freeze?

This patch makes no change to game engine functional behaviour. Whatever "pause" feature you're envisioning is just as possible after as before.

phyzical:
updated comment

src/NetscriptEvaluator.ts
```diff
@@ -1,15 +1,20 @@
 import { isString } from "./utils/helpers/isString";
 import { GetServer } from "./Server/AllServers";
+import { ScriptDeath } from "./Netscript/ScriptDeath";
 import { WorkerScript } from "./Netscript/WorkerScript";
 
 export function netscriptDelay(time: number, workerScript: WorkerScript): Promise<void> {
+  // Cancel any pre-existing netscriptDelay'ed function call
+  // TODO: the rejection almost certainly ends up in the uncaught rejection handler.
+  //       Maybe reject with a stack-trace'd error message?
```
phyzical, line 9: still need to handle this todo?

phyzical:
lgtm will approve with a gif showing it in action

src/NetscriptEvaluator.ts
Ornedan: Yeah, but in a different PR. It's a functional change (if minor) that'd be off-topic for this one.

src/NetscriptWorker.ts
Ornedan: Cleaned the code up a bit by switching the `startNetscript*Script` functions to return `Promise<void>` instead of `Promise<WorkerScript>`.

Ornedan:
Rebased on current dev. Checked that the changes still work correctly, which they do:

- exploit remains fixed
- normal script termination still works

Ornedan:
Re-re-based

Ornedan:
Rebased to current dev. Verified that:
- the exploit fix still works, running the sample exploit does not grant access to game internal state
- normal script execution and death works OK, tested by running my HWGW setup for a few cycles with no anomalies detected

PR #2930, opened 2022-02-07
Add tab switching support to vim mode

### LINKED ISSUE #2929: Vim mode tab changing commands do not work
In vim, you can change between open tabs (not windows/split panes) by using something like `gt` or `gT` in normal mode (or even `4gt` to change 4 tabs over). This doesn't work in vim mode, which makes working on something where you attempt to not have a monolith script quite annoying.

Bitburner v1.4.0 (49de4a28), Steam version.

Steps to reproduce:
1. Open a file in vim, eg `vim f1.js`
2. Go back to terminal
3. Open another file in vim, eg `vim f2.js`
4. Try gT/gt. Nothing happens

JParisFerrer:
This fixes #2929.

This lets you, in vim normal mode, type `gt` to go to the next tab (to the right), and `gT` to go to the previous tab (to the left). This also works with `3gt` to go three tabs to the right, etc. This behavior wraps around the tabs.

This is a standard vim mapping.

You can view the manual tests (re: the tab wrapping & repeating behavior) I performed in this recording:
https://capture.dropbox.com/oIgfov7Z6aHrUoKx

I also tested that `gt` while in INSERT (typing) or EX (:command) mode were not affected by this change.

I took a quick look around but didn't see any tests for the script editor I could try to extend for this. If they exist let me know and I can try to put some test in there

GIF version of demo:  
Recording-2022-2-6-at-41903-PM

### DIFF
--- a/src/ScriptEditor/ui/ScriptEditorRoot.tsx
+++ b/src/ScriptEditor/ui/ScriptEditorRoot.tsx
@@ -206,6 +206,32 @@ export function Root(props: IProps): React.ReactElement {
             save();
             props.router.toTerminal();
           });
+
+          // Setup "go to next tab" and "go to previous tab". This is a little more involved
+          // since these aren't Ex commands (they run in normal mode, not after typing `:`)
+          MonacoVim.VimMode.Vim.defineAction("nextTabs", function (_cm: any, args: {repeat?: number}, _vim: any) {
+            const nTabs = args.repeat ?? 1;
+            // Go to the next tab (to the right). Wraps around when at the rightmost tab
+            const currIndex = currentTabIndex();
+            if (currIndex !== undefined) {
+              const nextIndex = (currIndex + nTabs) % openScripts.length;
+              onTabClick(nextIndex);
+            }
+          });
+          MonacoVim.VimMode.Vim.defineAction("prevTabs", function (_cm: any, args: {repeat?: number}, _vim: any) {
+            const nTabs = args.repeat ?? 1;
+            // Go to the previous tab (to the left). Wraps around when at the leftmost tab
+            const currIndex = currentTabIndex();
+            if (currIndex !== undefined) {
+              let nextIndex = (currIndex - nTabs);
+              while (nextIndex < 0) {
+                nextIndex += openScripts.length;
+              }
+              onTabClick(nextIndex);
+            }
+          });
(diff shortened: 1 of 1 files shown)


src/ScriptEditor/ui/ScriptEditorRoot.tsx
```diff
@@ -605,16 +631,26 @@ export function Root(props: IProps): React.ReactElement {
...
       // Save currentScript to openScripts
-      const curIndex = openScripts.findIndex(
-        (script) =>
-          currentScript !== null &&
-          script.fileName === currentScript.fileName &&
-          script.hostname === currentScript.hostname,
-      );
-      openScripts[curIndex] = currentScript;
+      const curIndex = currentTabIndex();
+      if (curIndex !== undefined) {
```
phyzical, line 651: maybe throw an error if the tab click didnt work for whatever reason?
`dialogBoxCreate("ERRORTHING");`

phyzical:
LGTM will approve with a gif showing it in action

src/ScriptEditor/ui/ScriptEditorRoot.tsx
JParisFerrer: let me look into this (although, I believe the logic hasn't changed, as undefined is only returned when currentScript === null, which is the same check this function did before calling it; unless there is some other async-ness involved on modifying currentScript)

JParisFerrer:
Hi @phyzical , I linked a demo here: https://capture.dropbox.com/oIgfov7Z6aHrUoKx . Does it not load for you without an account? I can try to convert it explicitly into a GIF if it would help.

phyzical:
thanks for that if you dont mind, a gif would be easier you should just be able to drag n drop straight into the description

src/ScriptEditor/ui/ScriptEditorRoot.tsx
phyzical: yeah i dont think it was doing it before, but i felt it might aswell tell the user something went wrong. feel free to disregard :)

JParisFerrer:
Apologies for the delay @phyzical, I converted the demo into a GIF and added to the description

phyzical (approved):
LGTM, granted im not sure if its the right place either but im sure higher powers will let us know for next time :)

Issue #2261, opened 2022-01-01
Corporation dividend payout amount and taxes seem incorrect


norbby42:
Platform: Steam, PC Build ID 7928103 (Dec 23, 2021)
BitNode: 3, Corporatocracy

I have a corporation running with these stats:
```
Total Revenue: $21.792m / sec
Total Expenses: $4.466m / sec
Publicly Traded: Yes
Owned Stock Shares: 900.000m
Total Stock Shares: 1.500b
```

Setting Dividends to 50% yields the following information:
```
Retained Profits (after dividends): $8.611m / sec
Dividend Percentage: 50%
Dividends per share: $0.006 / sec
Your earnings as a shareholder: $3.599m / sec
```

----

**Issue 1:**  The tax rate does not behave as expected

I own 900 / 1500 = 60% of the company's shares, and expect to receive a baseline of $8.611m * 0.6 = $5.166m / sec

BUT!  I have not purchase either tax upgrade, so the expectation is that value is decreased by 15% to $4.391m / sec.

The actual amount that the UI tells me I will receive is $3.599m / sec, which is a noteworthy difference from $4.391m / sec.

It appears the use of Math.pow to apply the tax rate in addition to the BitNode multiplier in Corporation.tsx is applying a pretty hefty penalty beyond what is expected.
It's important to note that I am in BiteNode 3, so CorporationSoftCap == 1.0.  Were I in another bitnode, this would be a fair bit harsher.

----

**Issue 2:** The reported payout does not match actual payout

The UI says I will earn $3.599m / sec.
However, careful observation of my monetary value via the player widget at the top right shows that I only appear to gain that amount roughly once per corp cycle - every 10 seconds.

I verified this with the following script while no other scripts were running on any server, I had no hacknet nodes purchased, no gang, not working or performing any task/crime/etc, or any other source of income that I could identify:

```
export async function main(ns) {
	let prevMoney = ns.getPlayer().money;
	while (true) {
		await ns.asleep(10000); // 10 seconds = 1 corp cycle
		const money = ns.getPlayer().money;
	
		ns.tprint("Player funds delta since last cycle: " + ns.nFormat(money - prevMoney, "($0.00a)"));	

		prevMoney = money;
	}
}
```

Results:
```corptest.js: Player funds delta since last cycle: $3.61m
corptest.js: Player funds delta since last cycle: $3.62m
corptest.js: Player funds delta since last cycle: $3.62m
[...]

PR #2932, opened 2022-02-07
correct tax calculation of corporation dividends

Yirka:
This PR attempts to fix the inconsistent dividend tax rates.

To begin with, the game uses the following formula to calculate dividends after taxes:
```
return Math.pow(dividends, BitNodeMultipliers.CorporationSoftCap + upgrades);
```
Where (with values from my corporation)
```
dividends = 107718488.83520724
BitNodeMultipliers.CorporationSoftCap = 1
upgrades = -0.15
```
Yields the mathematical expression
```
107718488.83520724^0.85 = 6721198.38429
```
Which is 
```
6721198.38429 / 107718488.83520724 = 0.0624
or
93.76% effective tax rate
```
This 93.76% tax rate is far, far off the `dividendTaxPercentage = 50;`found within the same file.

Additionally, having had both of the upgrades (reducing tax by 5% and 10% respectively) would've changed the expression to:
```
107718488.83520724^1 = 107718488.83520724
```
Which, quite obviously, is the equivalent of paying _**no tax at all**_, rather than the expected 35%

As such, I propose the formula to be changed to:
```
return Math.pow(dividends, BitNodeMultipliers.CorporationSoftCap) * (1 - (this.dividendTaxPercentage / 100));
```
Alternatively (but more disruptively),
```
return dividends * BitNodeMultipliers.CorporationSoftCap * (1 - (this.dividendTaxPercentage / 100));
```

This change maintains the power relation between dividends and BitNodeMultipliers.CorporationSoftCap, but changes the taxes to scale more linearly with upgrades.
As a consequence, this makes corporation dividends much more profitable early game (50% tax instead of >90%), but much less profitable late game (35% tax instead of 0%).

I've looked through past commits, but wasn't able find any reference to this.
This leaves the possibility that this behavior is 100% inteded, and the problem lies within the user interface rather than the formulas.

Furthermore, with both upgrades present I've identified in the overview that dividends per second are an order of magnitude greater than my actual income.
This is because the dividends get calculated once every 10 seconds, but the overview pretends that this happens once per second instead.

image

fixes https://github.com/danielyxie/bitburner/issues/2261

### DIFF
--- a/src/Corporation/Corporation.tsx
+++ b/src/Corporation/Corporation.tsx
@@ -148,14 +148,7 @@ export class Corporation {
     const totalDividends = (this.dividendPercentage / 100) * cycleProfit;
     const dividendsPerShare = totalDividends / this.totalShares;
     const dividends = this.numShares * dividendsPerShare;
-    let upgrades = -0.15;
-    if (this.unlockUpgrades[5] === 1) {
-      upgrades += 0.05;
-    }
-    if (this.unlockUpgrades[6] === 1) {
-      upgrades += 0.1;
-    }
-    return Math.pow(dividends, BitNodeMultipliers.CorporationSoftCap + upgrades);
+    return Math.pow(dividends, BitNodeMultipliers.CorporationSoftCap) * (1 - (this.dividendTaxPercentage / 100));
   }
 
   determineValuation(): number {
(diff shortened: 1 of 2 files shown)


pigalot:
I would vote to reject this as it is effectively a nerf on corps and the current code looks fairly intended to me.

A good bug fix would be to fix the UI.

Yirka:
The problem with fixing the UI is that this issue is practically indescribable without putting the formula directly within the dividends display and hoping that the player will understand what it means. This is a problematic solution when understanding isn't guaranteed due to the effective tax rate approaching either 100% or 0% depending on upgrades and/or bitnode multipliers.

pigalot:
Your solution is more problematic, directly nerfing corps without fixing any of its 100s of problems is pointless.

phyzical:
(fyi no one is dissing your solution) we just know that it would become a problem again and then we need to understand another concoction of variables to fix the issue that could just be resolved in another way.

or rather lets say in a month we do balance corps, then we need to rework this again. but maybe there's another contributor that does it and misinterprets something.

all just to fix a ui bug 😆, and at the end of the day we can just make such big decisions those are up to higher powers

hydroflame:
The UI needs fixing but the code is correct. This is meant to have harsh diminishin return.

PR #2934, opened 2022-02-07
added check for NS interface with user defined guard

### LINKED ISSUE #1955: ports can send NS through them

### LINKED ISSUE #2011: tryWritePort doesn't check for basic types

phyzical:
# Bug fix
fixes #1955 
fixes #2011 
```
/** @param {NS} ns **/
	await ns.writePort(1,ns)
}
```

image

### DIFF
--- a/src/NetscriptFunctions.ts
+++ b/src/NetscriptFunctions.ts
@@ -1841,6 +1841,12 @@ export function NetscriptFunctions(workerScript: WorkerScript): NS {
     },
     tryWritePort: function (port: any, data: any = ""): any {
       updateDynamicRam("tryWritePort", getRamCost(Player, "tryWritePort"));
+      if ((typeof data !== "string" && typeof data !== "number") || isNS(data)) {
+        throw makeRuntimeErrorMsg(
+          "writePort",
+          `Trying to write invalid data to a port: only strings and numbers are valid.`,
+        );
+      }
       if (!isNaN(port)) {
         port = Math.round(port);
         if (port < 1 || port > CONSTANTS.NumNetscriptPorts) {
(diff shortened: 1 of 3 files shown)


src/NetscriptFunctions.ts
```diff
@@ -98,6 +98,11 @@ interface NS extends INS {
   stanek: IStanek;
 }
 
+//interface guard
+function isNS(object: any): boolean {
+  return (object as NS).gang !== undefined;
```
alainbryden, line 103: Can you explain why this is needed? `typeof data` when data is an NS object should return neither "string" nor "number" (presumably it returns "NS") - so the existing checks look like they would suffice.

phyzical: hmmmmmm ill double check... but the ticket implied that you could pass through the whole ns object, is it because string and number is a object lower down under the hood maybe?... (not sure if js works this way)

its possible this one is no longer an issue and i just derped and never tried it before the fix 😆 .

But the custom check was the recomended way online to enforce that something is not a certain type of interface.

src/NetscriptFunctions.ts
```diff
@@ -1836,6 +1841,12 @@ export function NetscriptFunctions(workerScript: WorkerScript): NS {
     },
     tryWritePort: function (port: any, data: any = ""): any {
       updateDynamicRam("tryWritePort", getRamCost(Player, "tryWritePort"));
+      if ((typeof data !== "string" && typeof data !== "number") || isNS(data)) {
+        throw makeRuntimeErrorMsg(
```
alainbryden, line 1845: I think these two lines of code (without the `isNS` bit) is all that was missing to cause the reported bug. I've heard a few people in chat say that "writePort" is fine, but "tryWritePort" is bugged.

phyzical:
@alainbryden this should be now approval able  👍

PR #2938, opened 2022-02-07
Don't check only gangs for sleeve augs

reacocard:
It is possible for a player to qualify for an aug through a faction and
not the gang, e.g. when the gang is newly formed and still getting up to
speed. So, we need to consider both gang qualifications and faction
qualifications when constructing the sleeve augmentation list.

tested by loading my save which doesn't have enough gang rep to qualify for any augs but does quality through factions. before this fix the sleeve augs pane was empty, after the fix it is correctly populated. then used the dev menu to add gang faction rep so it would qualify for augs through there and verified that those showed up and did not create duplicate entries.


phyzical:
mind adding a gif of it in action?

hmm but, reading this it seems to imply that your gangs faction should be the only place your allowed to get augs from?
```
// If player is in a gang, then we return all augs that the player
// has enough reputation for (since that gang offers all augs)
```

PR #2949, opened 2022-02-10
MISC: FIX #2205 It is no longer possible to get into debt by going to the gym or taking a class.

### LINKED ISSUE #2205: Courses let you go into negative balance

### LINKED ISSUE #3515: Sleeve--Negative numbers when selecting activities
Having a sleeve lets you go into the negatives when training at the gym or taking classes.

v 1.6.4

phyzical:
* taking courses and going to gym without money
* logic to cancel work when out of money

closes #2205
closes #3515 

Animation4


pigalot:
Have you added checks to the api and sleeves?

phyzical:
api yes i think you mean singularity correct?
sleeves no i have very little exp with them yet. 
but i will take a crack at that also, just for my understanding  reading the description of the docs sleeves are just "multiple versions of yourself" 
so basically just another version of ui functions that that have another entry point to most parts of the game?

phyzical:
ah damn good call sleeves is still able to go into -ve will fix now

alainbryden:
I'm so sorry for the effort you put in my dude, but this one is actually by design.

There's a steam achievement "Massive debt" - "Be $1b in debt.", and the intended way to earn this is by studying with little-to-no income.

If not for that, I would have thought this to be a change for the better.

Perhaps a way to salvage the PR is to indeed forbid working out at the gym (student debt is a real thing, but no one gives out a loan for working out at the gym) - and the same should probably apply to sleeves. Hydro will have the final call though, he might not be in favour of either.

alainbryden (changes requested):
I'm so sorry for the effort you put in my dude, but this one is actually by design.

There's a steam achievement "Massive debt" - "Be $1b in debt.", and the intended way to earn this is by studying with little-to-no income.

If not for that, I would have thought this to be a change for the better.

Perhaps a way to salvage the PR is to indeed forbid working out at the gym (student debt is a real thing, but no one gives out a loan for working out at the gym) - and the same should probably apply to sleeves. Hydro will have the final call though, he might not be in favour of either.

phyzical:
all good issue, do we know how this related the issue that was left in the bug list?

phyzical:
guess i just assumed based on this it was all good 😆 

image

alainbryden:
I hadn't seen that comment so this may actually stick.

In that case, we might just want to make sure there's an alternate route to that achievement? (Should be some way that require exploits, or else it should probably be converted to an SF-1 achievement)

phyzical:
yep i agree,
 tbh hydro may have just given in and his opinion might  change once hes back.
can just sit till then

illusori:
There's other in-game ways to get that achievement, although personally I do like the social commentary on student debt.

<details><summary>Spoilers</summary>Short selling on the stock market is the quickest way to get massively in debt.</details>

src/ui/WorkInProgressRoot.tsx
```diff
@@ -23,18 +23,21 @@ const CYCLES_PER_SEC = 1000 / CONSTANTS.MilliPerCycle;
 
 export function WorkInProgressRoot(): React.ReactElement {
   const setRerender = useState(false)[1];
+  const player = use.Player();
+  const router = use.Router();
   function rerender(): void {
     setRerender((old) => !old);
+    if (player.className != CONSTANTS.ClassStudyComputerScience && player.workType == CONSTANTS.WorkTypeStudyClass && player.money <= 0) {
```
phyzical, line 30: hmm after making more prs to the codebase i think doing this here may be a mistake....

hydroflame:
PR made before mega fmt

alainbryden:
A lot of players are enjoying strategies around going in to debt with gym/studies, so I propose that in order to not have this PR get anybody angry:
- Add a new setting to the options menu "overdraft protection" that defaults to true. When "overdraft protection" is enabled, the new code in this PR prevents the player from going into the negative and forcibly leaves the gym/studies.
- If "overdraft protection" is disabled by the user, the current behaviour continues.

phyzical:
happy to add a setting if we feel its needed but imo it just forces the user to adjust their scripts accordingly i.e instead of relying on free money you commit some crimes for a bit then switch back to training

taralx:
Am I reading this correctly? It looks like you can still go into debt by 1 tick.

phyzical:
yeah you would probably end up going into a little bit of debt, i could enforce 0, but that could conflict anything else that could cause debt in the same tick

taralx:
It's an improvement over infinite debt.

apdjz:
Sorry for the double post. I thought I did a search for the issue.

phyzical:
@apdjz no need to say sorry :) double is better than none

hydroflame:
While this issue is still current, it massively conflict with the new work system so I'm going to close.

PR #2966, opened 2022-02-12
Grammar Fix

MatthewTh0:
Product is the noun, while the correct part of speech here is the verb "produce". 
Also of note, the limit (from what I've tested, at least) doesn't change unless data is entered and then removed (otherwise the previous limit is kept) even if the button is clicked (which isn't expected behavior based on the text). However, I'm not sure how to convey this succintly and I'm not familiar enough with React to change the value when only the button is clicked (especially since it is probably the intended behavior to not change the value when the modal is closed by clicking outside of it).


phyzical (approved):
LGTM

PR #2967, opened 2022-02-13
Fix log message in ns.bladeburner.startAction

AdamTReineke:
Fixes a logging issue where the 2nd parameter isn't wrapped in single quotes. Technically untested but it's a trivial logging change. Before was 1 quotation mark at the end of the word, now I've wrapped the word with single quotes.

Before / Expected After:

```diff
-bladeburner.startAction: Starting bladeburner action with type 'Contracts' and name Tracking"
+bladeburner.startAction: Starting bladeburner action with type 'Contracts' and name 'Tracking'
```

Technically untested in the game, I haven't tried setting up a dev instance yet.


phyzical (approved):
LGTM

PR #2955, opened 2022-02-12
Bugfix/corp api fixes

phyzical:
# Bug fix
* added early out in UpgradeWarehouse when not enough cash
* added checks to enforce maxProducts
* added checks to enforce export material is valid for location
* added checks for market ti reasearches* fixed the export material fix
* changed teh dev menu fund adjuster for more granular control
* added checks for smartSupply and setSmartSupply
* few random autolints

closes #2860 
closes #2943 
closes #2942 
closes #2965
closes #2974
closes #3059 

corp

annnd test script 

```
/** @param {NS} ns **/
export async function main(ns) {
	//given new corp
	// with all apis and export api access
	// and where a is argiculture
	// and where b is mining
	// and where c is software
	const industry_a = "a"
	const city = "Sector-12"
	// will error
	

	await testWrapper(ns, async () => ns.corporation.setMaterialMarketTA1(industry_a, city, "Water", true))
	await testWrapper(ns, async () => ns.corporation.setMaterialMarketTA2(industry_a,city, "Water", true))
	await testWrapper(ns, async () => {
		ns.corporation.makeProduct("c", city, "a", 1, 1)
		ns.corporation.makeProduct("c", city, "b", 1, 1)
		ns.corporation.makeProduct("c", city, "c", 1, 1)
		ns.corporation.makeProduct("c", city, "d", 1, 1)
	})

	await testWrapper(ns, async () =>  ns.corporation.setProductMarketTA1(industry_a, "a", true) )

	await testWrapper(ns, async () =>  ns.corporation.setProductMarketTA2(industry_a, "a", true) )
	await testWrapper(ns, async () => ns.corporation.exportMaterial("b", city, industry_a, city, "Metal", 1) )
	

	await testWrapper(ns, async () =>  await ns.corporation.setAutoJobAssignment(industry_a, city,"Operations", 0) )
	await testWrapper(ns, async () =>  ns.corporation.setSmartSupply(industry_a, city,true) )
	await testWrapper(ns, async () => ns.corporation.setSmartSupplyUseLeftovers(industry_a, city, "Water", true) )
	await testWrapper(ns, async () => ns.corporation.issueDividends(69) )
	await testWrapper(ns, async () => {
		while (1) {
			// will error eventually when no cash
			ns.corporation.upgradeWarehouse(industry_a, city)
			await ns.sleep(1000)
		}
	})
}

async function testWrapper(ns, func) {
	try {
		await func()
	} catch (e) {
	}
}
```

[...]

PR #2969, opened 2022-02-13
QoL pass on corp API

Mawootad:
* Adds documentation for setSmartSupplyUseLeftovers
* Fixes several typos/inaccuracies in corp error messages
* Changes the following functions to return true/false for succeed/fail and changes error throwing methods to not throw exceptions due to simple state errors such as duplicate item existing or insufficient funds:
  * purchaseWarehouse
  * upgradeWarehouse
  * upgradeOfficeSize
  * hireAdVert
  * research
  * expandIndustry
  * expandCity
  * unlockUpgrade
  * levelUpgrade
* Changes purchaseMaterial to use 0 for the purchase amount if the purchase amount is less than 0
* Require an office in a city to purchase a warehouse in that city
* Don't buy duplicate offices in cities with offices
* Don't allow purchasing non-levelable upgrades multiple times
* Don't allow purchasing the same research multiple times
* Require funds to upgrade a warehouse

Documentation


phyzical:
i don't mean to be rude, appreciate the input but did you look for existing prs/ issues?

also there are no test scripts to prove you tested anything, maybe some gifs would help in the future.

(saying that it looks like there is some things that ware missed in the existing prs) https://github.com/danielyxie/bitburner/pull/2955

i will look at adding another pr with the subset of those changes.

as for all the changes to return values and removing errors being thrown did you happen to ask anyone else if this is what they wanted after all there are many players.

Mawootad:
I did not see the previous pull request, as it was submitted after I started work on this. The return values vs throwing errors is much closer in line with a majority of existing functionality. I asked in the Discord and got very limited feedback, but they also pointed out that most functions return true/false when you ask them to do something that might fail due to simple variable state (eg buying a server returns false when you are out of funds because that state can vary wildly and the game typically punishes players for asking questions that they don't need to). I have added demonstration that the documentation is updated and will add a test script shortly.

PR #2978, opened 2022-02-14
Clarify math expression operator precedence for coding contract

factubsio:
Clarify math expression operator 

The description + examples leave this ambiguous, this change adds a hint.


phyzical:
suggestion: how about an example explaining what is meant by "operator precedence" 

maybe something simple like 

`'1+2*3' and '3*2+1'  target: 7`

factubsio:
Yep will write it up 👍

PR #2989, opened 2022-02-16
UI: FIX #2228,#2958 Fix tab highlights and highlight files not on home.

### LINKED ISSUE #2228: Game should it make it clear that files not on home should not really be edited
Text editor maybe should have the file name in red?

### LINKED ISSUE #2958: Incorrect editor tab highlighting when editing files with same name on different servers
* Open a file in the editor on one server: `home; nano foo.js` 
* Connect to a different server: `connect iron-gym`
* Open a file with the same name as the first one: `nano foo.js` 
* Go to the editor and select the tab for either file
* The tabs for both files will be highlighted

Version: a3c599e6

phyzical:
fixes #2228
fixes #2958 

added logic to also check hostname for "current script"# Bug fix
added logic to display external servers as red
added logic to display server on save dialog

editFileUpdates

note: screenshots out of date and will show info color instead of error color now


src/ScriptEditor/ui/ScriptEditorRoot.tsx
```diff
@@ -822,22 +822,29 @@ export function Root(props: IProps): React.ReactElement {
...
+                    borderColor: Settings.theme.button,
+                    color: Settings.theme.primary
+                  } : {
+                    background: Settings.theme.backgroundsecondary,
+                    borderColor: Settings.theme.backgroundsecondary,
+                    color: Settings.theme.secondary
+                  }
+
+                  if (externalScript) {
+                    colorProps.color = Settings.theme.error
```
nickofolas, line 839:
```suggestion
                    colorProps.color = Settings.theme.info
```

Using `Settings.theme.error` feels a bit out of place here. It makes it seem like something's wrong with the tab, rather than just indicating a difference of origin.

I think that changing it to `Settings.theme.info` or something else less "alarming" would help it still draw attention to the difference without giving the wrong idea :)

hydroflame:
PR made before mega fmt

PR #2990, opened 2022-02-16
Add A Simple Button menu to the Bitverse Screen if ascii art is disabled

TheMas3212:
this fixed the issue that was noticed that the current ascii art page is
not possible to navigate with a screen reader

image

### DIFF
--- a/src/BitNode/ui/BitverseRoot.tsx
+++ b/src/BitNode/ui/BitverseRoot.tsx
@@ -10,6 +10,9 @@ import makeStyles from "@mui/styles/makeStyles";
 import createStyles from "@mui/styles/createStyles";
 import Typography from "@mui/material/Typography";
 import Tooltip from "@mui/material/Tooltip";
+import { Settings } from "../../Settings/Settings";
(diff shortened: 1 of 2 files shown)


phyzical:
@TheMas3212 sorry i dindt see this one, mind also adding these changes to it?
https://github.com/danielyxie/bitburner/pull/3001 
(its just i confirmed with the user who reported the bug that this fixed their issue about clickablility but they did raise the fact it was still full of symbols which yours resolves)

Also should we maybe add the description of each bitnode next the each button?

TheMas3212:
clicking the button opens up the normal bitnode panel with the full description
i could add the name + subtitle of each node next to each button though?

phyzical:
yeah that's what i was thinking tbh, i'm not sure if there's an accessibility way to provide this to the button itself though.

its just if i had to use a screen reader it could be quite annoying being like "wait which bit node was the one with the stuff i wanted again"

TheMas3212:
Closing this as it was merged into #3001

PR #3024, opened 2022-02-23
Fix 3022: Missable achievements on hacknet nodes after beating BN9

### LINKED ISSUE #3022: Missable Achievement: "That's the new limit"
If a player beats BN9 (Hacktocracy) before getting all the hacknet node related achievements, then they become impossible to get when you have Hashnet servers instead of the original Hacknet nodes.

Other bitburner devs told me that the intention is to not have any missable achievements in this game, so it would be good to fix this so it's no longer missable.

image
image

BrianLDev:
Fix so that any hacknet node related achievements are no longer missable after a player beats BN9 and hacknet nodes are converted into hashnet servers.

Fixes #3022

### DIFF
--- a/src/Achievements/Achievements.ts
+++ b/src/Achievements/Achievements.ts
@@ -339,24 +339,23 @@ export const achievements: IMap<Achievement> = {
   FIRST_HACKNET_NODE: {
     ...achievementData["FIRST_HACKNET_NODE"],
     Icon: "node",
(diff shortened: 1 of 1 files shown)


phyzical (approved):
lgtm

BrianLDev:
Update: it was pointed out that the HACKNET_SERVER achievements have additional unlocks to cover the old hacknet node achievements.  So I'm closing out the issue and related PR.
image

PR #3025, opened 2022-02-23
Fix 3023: hacknet servers incorrectly being named 'hacknet-node-x'

### LINKED ISSUE #3023: Hacknet Servers are incorrectly named "nodes"
After beating BN9 (Hacktocracy) and getting upgrade from Hacknet nodes to Hashnet servers, the new servers show up under the name `hacknet-node-x`.  For clarity and consistency, it's better if they're named `hacknet-server-x` since nodes and servers have very different functionality.

image

BrianLDev:
quick and easy fix

fixes #3023

### DIFF
--- a/src/PersonObjects/Player/PlayerObjectServerMethods.ts
+++ b/src/PersonObjects/Player/PlayerObjectServerMethods.ts
@@ -46,7 +46,7 @@ export function getUpgradeHomeCoresCost(this: IPlayer): number {
 
 export function createHacknetServer(this: IPlayer): HacknetServer {
   const numOwned = this.hacknetNodes.length;
-  const name = `hacknet-node-${numOwned}`;
+  const name = `hacknet-server-${numOwned}`;
   const server = new HacknetServer({
     adminRights: true,
     hostname: name,


phyzical:
hmm it does get pushed to `this.hacknetNodes.push(NEWSERVER)`

if you search "hacknet-node" i don't think this is as simple as it seems on the surface

for example this code
https://github.com/danielyxie/bitburner/blob/master/src/Server/Server.ts#L61

Feodoric:
I think fixing this is also likely to break user scripts (big red flag unless the issue being fixed is truly dire). For example, anyone who uses the string 'hacknet-node' to prevent scripts from running on their hacknet will find their hash rate dropping to 0 if the names are changed.

BrianLDev:
@phyzical @Feodoric @ChrissiQ since this one is a bit controversial I'll just close out the issue and related PR.

In general, it's still best to avoid inconsistent terminology and also avoid using "magic strings" in scripts, but there are bigger fish to fry at the moment so I'm not going to fight for this change.

phyzical:
i agree, just an fyi wasnt against the idea. i just saw the magic strings being matched in code so who knows if thats by design or not :D

PR #3034, opened 2022-02-25
docs: fix `@returns` of `bladeburner.getBonusTime()`

dlchen:
# Documentation fix

* update `@returns` of `ns.bladeburner.getBonusTime()` to correctly identify the return value in seconds (not milliseconds)
* you can see on line [663](https://github.com/danielyxie/bitburner/blob/dev/dist/bitburner.d.ts#L663) it correctly says seconds

### DIFF
--- a/dist/bitburner.d.ts
+++ b/dist/bitburner.d.ts
@@ -668,7 +668,7 @@ export declare interface Bladeburner {
      * For example, if an action takes 30 seconds to complete but you’ve accumulated over
      * 30 seconds in bonus time, then the action will only take 6 seconds in real life to complete.
      *
-     * @returns Amount of accumulated “bonus time” (milliseconds) for the Bladeburner mechanic.
+     * @returns Amount of accumulated “bonus time” (seconds) for the Bladeburner mechanic.
      */
     getBonusTime(): number;
 }


phyzical:
just incase have you 100% confirmed it does return seconds (incase that part of the docs is meant to say `30000 milliseconds`

dlchen:
@phyzical yes I confirmed this.

I was confused because I was trying to convert `getBonusTime()` into minutes (while assuming it was ms) but it didn't match up with what was displaying on the Bladeburner tab.

Also I found an old message in the Discord from BigD saying that "[getBonusTime() returns seconds, not ms](https://discord.com/channels/415207508303544321/415213413745164318/502244261283102730)".

dlchen:
Adding visual proof:
image

This is a screenshot of a log of `bonusTime` alongside the bonus time displayed in the UI.

```
18280 assumed to be seconds / 60 = 304.667 minutes
304.667 minutes / 60 = 5.0778 hours // lines up with the UI

18280 assumed to be ms / 1000 = 18.28 seconds // does not line up with the UI
```

phyzical:
thanks for the extra info makes explicitly clear :)

hydroflame:
In this case the code is incorrect. The game shoudl always return milliseconds units of time.

phyzical:
@dlchen feel free to fix it on the other side instead :)

phyzical:
actually looks like someones got you covered :) https://github.com/danielyxie/bitburner/pull/3053/

PR #3038, opened 2022-02-25
Fix #3037: wrap user-supplied atExit function so that we can't accide…

### LINKED ISSUE #3037: `atExit` leaks `WorkerScript` as `this`
Currently player-supplied function given to `atExit` is stored as a property of a `WorkerScript` object and called from there. As a consequence `this` is bound to the `WorkerScript` object during the call to player code. A player can then modify the parts of game state that are accessible from a `WorkerScript` object.

Sample exploit:
```js
/** @param {NS} ns **/
	ns.atExit(function() {
		this.getServer().maxRam = Number.MAX_SAFE_INTEGER;
	});
}
```
localhost_8000__noscripts

Ornedan:
Fixes #3037, see the bug report for a sample exploit.

Approach chosen is to wrap the user-supplied function in a new function, thus isolating it and preventing any leak of game objects via `this` when `WorkerScript.atExit` is called.

### DIFF
--- a/src/NetscriptFunctions.ts
+++ b/src/NetscriptFunctions.ts
@@ -2279,7 +2279,7 @@ export function NetscriptFunctions(workerScript: WorkerScript): NS {
       if (typeof f !== "function") {
         throw makeRuntimeErrorMsg("atExit", "argument should be function");
       }
-      workerScript.atExit = f;
+      workerScript.atExit = () => { f(); }; // Wrap the user function to prevent WorkerScript leaking as 'this'
     },
     mv: function (host: string, source: string, destination: string): void {
       updateDynamicRam("mv", getRamCost(Player, "mv"));


phyzical (approved):
LGTM nice simple fix, 

could you please provide a gif of a few usages just to show it all working

Ornedan:
> LGTM nice simple fix,
> 
> could you please provide a gif of a few usages just to show it all working

This is not an UI change, there is nothing to GIF.

phyzical:
a gif of it being used in a script i mean, showing its outcome

Ornedan:
localhost_8000__noscripts (1)
Post-change `this` is `undefined` in player code `atExit` function, as intended.

phyzical:
coolies, any proof you can get values still after the change?

Ornedan:
> coolies, any proof you can get values still after the change?

Please clarify. Values of what, from where?

phyzical:
sorry, just checked the issue code sample (maybe add to the the pr desc too) i would assume people use the getServer to then spawn another script given the host as one quick example
```
/** @param {NS} ns **/
	ns.atExit(function() {
		alert(this.getServer().name)
	});
}
```

but obviously this wont work now hmm those users could store it outside and pass it in i guess. ive just got a feeling this change would cause pain to those that expect values to be "updated" at the time exit is run.

can we pass in ns inside the atExit?

phyzical:
while i agree with the intention, i wonder if it shouldn't be handled by making the exploitable properties immuatable i.e 

`Object.defineProperty(obj, "maxRam", { configurable: false, writable: false });` 

though this isnt a simple change to implement... many moving parts..

 its just it feels like the usefulness of atExit drops dramatically, i.e no more port writing for those that use magic uis (for dynamic changing values) ect.

Ornedan:
> while i agree with the intention, i wonder if it shouldn't be handled by making the exploitable properties immuatable i.e
> 
> `Object.defineProperty(obj, "maxRam", { configurable: false, writable: false });`
> 
> though this isnt a simple change to implement... many moving parts..

Uh, it's the live game internal server object, it can't be made immutable.

> 
> its just it feels like the usefulness of atExit drops dramatically, i.e no more port writing for those that use magic uis (for dynamic changing values) ect.

Writing to ports during `atExit` is just as possible after change as before.

phyzical:
>Uh, it's the live game internal server object, it can't be made immutable.

let me rephrase it as maybe im not understanding, if a user wanted to get the say current server cash at the exit of script how would they do that now?

Ornedan:
> > Uh, it's the live game internal server object, it can't be made immutable.
> 
> let me rephrase it as maybe im not understanding, if a user wanted to get the say current server cash at the exit of script how would they do that now?

`ns.getServerMaxMoney(ns.getHostname())`, same as any other time.

src/NetscriptFunctions.ts
```diff
@@ -2279,7 +2279,7 @@ export function NetscriptFunctions(workerScript: WorkerScript): NS {
       if (typeof f !== "function") {
         throw makeRuntimeErrorMsg("atExit", "argument should be function");
       }
-      workerScript.atExit = f;
+      workerScript.atExit = () => { f(); }; // Wrap the user function to prevent WorkerScript leaking as 'this'
```
ApamNapat, line 2282: Wouldn't 
```suggestion
      workerScript.atExit = () => f(); // Wrap the user function to prevent WorkerScript leaking as 'this'
```
be equivalent and simpler?

PR #3045, opened 2022-02-28
sort player's joined factions for cleaner rendering

jjayeon:
Each time a player joins a faction, their list of factions will be sorted according to a preset ranking, mostly identical to the one in the documentation.  That way, when the factions are rendered in the UI, they'll appear in that order as well.  This will make it much easier to keep track of which factions players have joined.

### DIFF
--- a/src/Faction/FactionHelpers.tsx
+++ b/src/Faction/FactionHelpers.tsx
@@ -19,6 +19,42 @@ import { SourceFileFlags } from "../SourceFile/SourceFileFlags";
 import { dialogBoxCreate } from "../ui/React/DialogBox";
 import { InvitationEvent } from "./ui/InvitationModal";
 
+const factionOrder = [
+  "CyberSec",
+  "Tian Di Hui",
+  "Netburners",
+  "Sector-12",
+  "Chongqing",
+  "New Tokyo",
+  "Ishima",
+  "Aevum",
+  "Volhaven",
+  "NiteSec",
+  "The Black Hand",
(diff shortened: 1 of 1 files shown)


phyzical:
what if instead we sort  them alphabetically that way if any new ones are ever added they will slot in the appropriate place

jjayeon:
I chose this method because I think a very common use case will be to have the wiki open in one tab and Bitburner in the other, so having factions being in the same order in both would be very handy.

One possibility is to somehow pull the order on the wiki into the code here --- I don't know how to do that, but if someone else does I'd be thrilled.

phyzical:
why not make a pr to update the wiki to be alphabetical too?

jjayeon:
There's an idea too... the way I see it, here are the options and drawbacks:

- Sort alphabetically
  - + easier to code
  - + visually pleasing
  - - lacks information clustering
- Sort by category (hacking, city, crime, etc) (how the wiki does it atm)
  - + good information clustering
  - - requires hard-coding faction order
- Sort by requirements
  - + also good clustering
  - + better for new players --- relevant factions at the top
  - + rough sense of timeline for progression
  - - also might require hard coding, or a better quantification of what "requirements" are

I'll mull over this for a bit.  Having sorting at all is a plus tho imo.

phyzical:
ah granted i haven't looked at the wiki mind including a link to help other visualize what your trying to replicate. (my biggest concern is that this list may become out of date one day and its not very obvious it needs more items is all :) )

PR #3046, opened 2022-02-28
add box showing remaining augments per faction

jjayeon:
On the "Factions" page next to the "Details" button, the player will see a box containing how many Augmentations a faction offers that they haven't purchased.  Neuroflux won't count as long as they have at least one level.

Screenshot from 2022-02-28 07-05-54

### DIFF
--- a/src/Faction/ui/FactionsRoot.tsx
+++ b/src/Faction/ui/FactionsRoot.tsx
@@ -75,6 +75,15 @@ export function FactionsRoot(props: IProps): React.ReactElement {
                       <Button onClick={() => openFaction(Factions[faction])}>Details</Button>
                     </Box>
                   </TableCell>
+                  <TableCell align="right">
+                    <Typography noWrap ml={2} mb={1}>
+                      Augmentations Left: {Factions[faction]
+                        .augmentations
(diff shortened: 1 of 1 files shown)


DJ-Laser:
Cool! I would like this very much when trying to get all the augs from my factions

nickofolas:
I like this! I wonder if we could take this one step further by showing this information in a button that can double as a shortcut to the faction augmentations page? I know I would love the extra QoL.

phyzical:
oo @nickofolas thats a cool idea, but we would start running out of space or are you think its just the same text as a button?

nickofolas:
the latter yeah, take the text that says the number of remaining augmentations and replace it with a shortcut button with a label like `Augmentations (n left)`

jjayeon:
> LGTM, just a suggestion on adding some more min width so that when theres more than 10/100 it doesn't miss align

done!

TheRealMaxion:
Nice, though when the faction has a gang it still only counts the augments that came originally with the faction.

nickofolas:
> Nice, though when the faction has a gang it still only counts the augments that came originally with the faction.

fixed by #3119

PR #3053, opened 2022-03-01
Made getBonusTime return miliseconds

### LINKED ISSUE #3050: Bladeburner getBonusTime returns seconds instead of millis.

waffleattack:
Fixes getBonusTime to properly return milliseconds instead of seconds.
fixes #3050


phyzical (approved):
LGTM, nice spot :)

PR #3054, opened 2022-03-01
Removed redundant repetitions of | in alias regex

ApamNapat:
Repeating `|` makes no sense inside `[]` as it is treated like any other character and not a meta one. Probably it shouldn't have ended in the regex in the first place but by now it should probably be kept not to break interface

### DIFF
--- a/src/Alias.ts
+++ b/src/Alias.ts
@@ -36,7 +36,7 @@ export function printAliases(): void {
 
 // Returns true if successful, false otherwise
 export function parseAliasDeclaration(dec: string, global = false): boolean {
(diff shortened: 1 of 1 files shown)


phyzical:
isn't the  | needed though? mind adding a gif of some sort of online regex tester proving its the same?

ApamNapat:
> isn't the | needed though? mind adding a gif of some sort of online regex tester proving its the same?

Inside `[]` `|` is just a regular character as `[]` implies an alternative already as I wrote in description. This example should make it clear (from regex101) 
image

phyzical (approved):
TIL, LGTM

PR #3055, opened 2022-03-01
Simplified some booleans and if elses

ApamNapat:
I'm positive that this doesn't change the way code works but clearly reduces boilerplate and redundancy

### DIFF
--- a/src/ExportBonus.tsx
+++ b/src/ExportBonus.tsx
@@ -6,8 +6,7 @@ export let LastExportBonus = 0;
 const bonusTimer = 24 * 60 * 60 * 1000; // 24h
 export function canGetBonus(): boolean {
   const now = new Date().getTime();
-  if (now - LastExportBonus > bonusTimer) return true;
-  return false;
+  return now - LastExportBonus > bonusTimer;
 }
 
 export function onExport(p: IPlayer): void {
--- a/src/data/codingcontracttypes.ts
+++ b/src/data/codingcontracttypes.ts
@@ -307,15 +307,7 @@ export const codingContractTypesMetadata: ICodingContractTypeMetadata[] = [
         reach = Math.max(i + data[i], reach);
       }
       const solution: boolean = i === n;
-
-      if (ans === "1" && solution) {
-        return true;
-      }
-      if (ans === "0" && !solution) {
-        return true;
-      }
-
-      return false;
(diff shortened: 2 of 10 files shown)


inigo:
Those changes all look correct to me.

phyzical:
i understand the reasoning for the changes, but i think this pattern is done so that the "last comparison" return doesn't cause issues if any new ones need to be added after it, as in if the developer needs a new "last comparison" they just add it before the default return. if im making any sense 😆

src/data/codingcontracttypes.ts
```diff
@@ -307,15 +307,7 @@ export const codingContractTypesMetadata: ICodingContractTypeMetadata[] = [
...
-
-      if (ans === "1" && solution) {
-        return true;
-      }
-      if (ans === "0" && !solution) {
-        return true;
-      }
-
-      return false;
+      return ans === "1" && solution || ans === "0" && !solution;
```
phyzical, line 310: shouldn't we add parenthesis for better readability

ApamNapat:
> i understand the reasoning for the changes, but i think this pattern is done so that the "last comparison" return doesn't cause issues if any new ones need to be added after it, as in if the developer needs a new "last comparison" they just add it before the default return. if im making any sense 😆

This applies to some cases only, but anyways I think this is easy to revert if needed (YAGNI kinda thing) and I avoided the changes for cases where there were long `if {} else if {} ...` chains for the reasons you gave

src/data/codingcontracttypes.ts
ApamNapat: You convinced me me with JS doing weird stuff

phyzical:
nah i agree proably should have said its more of a `nit pick comment :)`

src/ExportBonus.tsx
```diff
@@ -6,8 +6,7 @@ export let LastExportBonus = 0;
 const bonusTimer = 24 * 60 * 60 * 1000; // 24h
 export function canGetBonus(): boolean {
   const now = new Date().getTime();
-  if (now - LastExportBonus > bonusTimer) return true;
-  return false;
+  return now - LastExportBonus > bonusTimer;
```
phyzical, line 9: parentheses for this one too :)

ApamNapat: Now this one really doesn't convince me

phyzical (approved):
lgtm

PR #3066, opened 2022-03-05
Fix editor "close" naming

### LINKED ISSUE #2961: Editor 'close' button doesn't actually close file
The editor has a big button labelled 'Close' at the bottom of the screen next to 'Save', but hitting it doesn't actually close the current file; it just returns the user to the terminal screen.

Note: I suppose it's possible that that button is supposed to be 'Close editor and return to terminal' and this is the intended behaviour, but in that case:
* Why do we even need such a button? 'Go to terminal' already has a dedicated button on the left sidebar.
* The presentation of the button is confusing; its name, along with its positioning right next to the 'save' button, implies it means 'close file'. If it's supposed to be 'close editor' then the button should be renamed 'Close editor' or 'Quit to terminal' so that its purpose is clear

Version: a3c599e6

phyzical:
* evened out spacing of buttons
* renamed close to Terminal

closes #2961 

image

### DIFF
--- a/src/ScriptEditor/ui/ScriptEditorRoot.tsx
+++ b/src/ScriptEditor/ui/ScriptEditorRoot.tsx
@@ -892,8 +892,8 @@ export function Root(props: IProps): React.ReactElement {
             {ram}
           </Button>
           <Button onClick={save}>Save (Ctrl/Cmd + s)</Button>
-          <Button onClick={props.router.toTerminal}>Close (Ctrl/Cmd + b)</Button>
-          <Typography sx={{ mx: 1 }}>
+          <Button sx={{ mx: 1 }} onClick={props.router.toTerminal}>Terminal (Ctrl/Cmd + b)</Button>
+          <Typography>
             {" "}
             Documentation:{" "}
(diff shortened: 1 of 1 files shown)


src/ScriptEditor/ui/ScriptEditorRoot.tsx
```diff
@@ -892,8 +892,8 @@ export function Root(props: IProps): React.ReactElement {
             {ram}
           </Button>
           <Button onClick={save}>Save (Ctrl/Cmd + s)</Button>
-          <Button onClick={props.router.toTerminal}>Close (Ctrl/Cmd + b)</Button>
-          <Typography sx={{ mx: 1 }}>
+          <Button sx={{ mx: 1 }} onClick={props.router.toTerminal}>Terminal (Ctrl/Cmd + b)</Button>
```
ApamNapat, line 895: I think this is "truer" but is it better? Maybe something like `Go to terminal` would work? Or would that be to verbose?

phyzical: i wanted to the same originally but theres not enough real estate, if we make it a multiline button flex then resizes the other buttons too and they look gross. happy to try something else if you have any ideas

PR #3069, opened 2022-03-05
Added support to connect to ip addresses again

### LINKED ISSUE #2877: connect to IP doesn't work
Heyho. I just tried to connect to a server using it's IP as suggested by the help page of connect. It told me that I couldn't directly connect to that IP.

Terminal-input and -output, relevant passages marked:
[home ~/]> help connect
Usage: connect [hostname]
Connect to a remote server. The hostname **or IP address** of the remote server must be given as the argument 
to this command. Note that only servers that are immediately adjacent to the current server in the network can be connected to. To 
see which servers can be connected to, use the 'scan' command.

[home ~/]> scan
Hostname        IP       Root Access
**n00dles         3.3.9.6**  Y
foodnstuff      12.9.1.0 Y
sigma-cosmetics 66.4.3.0 Y
joesguns        81.7.0.1 Y
hong-fang-tea   80.3.7.5 Y
harakiri-sushi  84.2.9.9 Y
iron-gym        91.0.5.3 Y
[home ~/]> connect 3.3.9.6
**Cannot directly connect to 3.3.9.6**

reproduced in steam-version 1.4.0 (49de4a28), and browser-version 1.4.0 (a3c599e6)

phyzical:
* Added support to connect to ip addresses again

closes #2877 

image

### DIFF
--- a/src/NetscriptFunctions/Singularity.ts
+++ b/src/NetscriptFunctions/Singularity.ts
@@ -587,7 +587,7 @@ export function NetscriptSingularity(
       for (let i = 0; i < server.serversOnNetwork.length; i++) {
         const other = getServerOnNetwork(server, i);
         if (other === null) continue;
-        if (other.hostname == hostname) {
+        if (other.hostname == hostname || other.ip == hostname) {
           player.getCurrentServer().isConnectedTo = false;
           player.currentServer = target.hostname;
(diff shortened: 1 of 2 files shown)


ApamNapat (approved):
Maybe if both `hostname` and `ip` are valid, then the variable named `hostname` could be renamed. But I don't have a good idea

PR #3074, opened 2022-03-06
Present offline message as list

DSteve595:
I just recently started playing. After taking a quick break and starting the game again, I got this message:

image

The info's great, but the bit about the unchanged reputation is redundant, and irrelevant to new players that already have a bunch of new concepts to wrap their heads around. I think showing this info as a list would solve a bunch of little problems here.

image

- More glanceable
- ~Conditionally shows reputation~
  - Scrapped this. Could cause confusion of its own.
- Fewer changes needed if/when another element is added

Awesome game by the way!

### DIFF
--- a/src/engine.tsx
+++ b/src/engine.tsx
@@ -405,9 +405,7 @@ const Engine: {
               <ul>
                 <li>Your scripts generated{" "} <Money money={offlineHackingIncome} /></li>
                 <li>Your Hacknet Nodes generated {hacknetProdInfo}</li>
-                {offlineReputation != 0 &&
-                  <li>You gained{" "} <Reputation reputation={offlineReputation} /> reputation divided amongst your factions
-                  </li>}
+                <li>You gained{" "} <Reputation reputation={offlineReputation} /> reputation divided amongst your factions</li>
               </ul>
             </>,
           ),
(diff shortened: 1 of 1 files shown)


phyzical:
hmmmm i like the changes much cleaner, but i feel like we should show 0.00 rep as it just makes it more obvious that they earned no rep.

i get the "dont over confuse new users" but you will still end up finding out about rep even before exiting your first bitnode.

if you reaaallly wanted to fight for this, maybe we can check if they have ever earned any repuation before? then show it rather than basing it on if there was any offline earnings?

phyzical (approved):
lgtm just the one comment about always hiding 0.00 rep

DSteve595:
>check if they have ever earned any repuation before

I think this is totally ideal. I don't know enough about the project to know how to hook this up, though.
Changed the reputation to be unconditional.

phyzical:
👍  i had a quick scan and there isnt any magical functions to work this out, it would require looping all factions to call `getFactionRep` and looping all companies to check `getCompanyRep"

If your feeling up for it it would be something along the lines of the following

```
const hasAnyRep = (Factions.reduce((factionName) => getFactionRep(factionName)) + Companies.reduce((companyName) => getCompanyRep(companyName))) > 0
```

PR #3075, opened 2022-03-06
Simplified some relative imports

ApamNapat:


phyzical (approved):
im surprised eslint doesn't try to fix this up

ApamNapat:
> im surprised eslint doesn't try to fix this up


I searched for `import` in the eslint config file and tried to google this but apparently eslint doesn't help with that

PR #3076, opened 2022-03-06
Trailing whitespace and files not ending with \n are now illegal

ApamNapat:
Most projects across languages disallow trailing whitespace, same goes for files that don't end with `\n` . Cf. https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline and https://eslint.org/docs/2.0.0/rules/no-trailing-spaces


phyzical (approved):
lgtm, one suggestion though. maybe in the future it might be better to just update the linting rules and just wait for a developer to touch the files.

its just all these linting prs will probably cause conflicts, where as the developer touching files *should* lint it as they go

ApamNapat:
> lgtm, one suggestion though. maybe in the future it might be better to just update the linting rules and just wait for a developer to touch the files.
> 
> its just all these linting prs will probably cause conflicts, where as the developer touching files _should_ lint it as they go

I searched for `import` in the eslint config file and tried to google this but apparently eslint doesn't help with that

ApamNapat:
> lgtm, one suggestion though. maybe in the future it might be better to just update the linting rules and just wait for a developer to touch the files.
> 
> its just all these linting prs will probably cause conflicts, where as the developer touching files _should_ lint it as they go

Do you know how to even set this up "natively" without filtering with console commands like proposed here https://stackoverflow.com/questions/54511168/enable-eslint-only-for-edited-files? I'm more of a "do it once start to finish and take some pain if needed" guy, but I hear you. Hopefully ~24 lines of changes won't be that painful.

phyzical:
@ApamNapat i use eslint extension with vscode and just enable lint of save tbh.

not native... i know, not sure if there is a antive way to target the file that was saved without the IDE handling it.

but, i understand where your coming from also easier to just have a git hook to just normalizes the repo.

Not a big deal tbh there isnt that many changes :)

PR #3082, opened 2022-03-07
Fix Static Ram Circumventing for all NS functions

CrafterKolyan:
Example of script which circumvents static ram calculation and is not triggered by dynamic ram cost calculation
image
image

```
/** @param {NS} ns **/
    ns.tprint(eval("ns.hackAnalyzeSecurity")(12))
}
```


phyzical:
and picture showing it fixed after the change if possible :)

CrafterKolyan:
@phyzical Here it is
image

phyzical (approved):
lgtm

PR #3091, opened 2022-03-07
Fix extra space on editor tabs

nickofolas:
Removes the minWidth attribute of editor tabs so that the border doesn't extend beyond the tab

<table>
  <th>Before</th>
  <th>After</th>
  <tr>
    <td>
      
    </td>
    <td>
      
    </td>
  </tr>
</table>


phyzical (approved):
lgtm

PR #3092, opened 2022-03-08
🐛  (Bug: Corp) Fixes sufficient player money check to buy back shares.

ChrissiQ:
`BuyBackShares()` checks the corp for sufficient funds, but since shares are purchased with player money, it should be checking the player's money instead.

Sorry no screenshot... having trouble setting up a save file for good testing here.  I'm pretty sure this works and is pretty simple bug fix.


phyzical (approved):
/facepalm my bad good catch

holgero:
I just got bitten by this bug. So I will need to let my corporation make money to be able to buy its shares back, right? This might become difficult. I hope the share price doesn't rise faster than the funds of my corporation...

phyzical:
atm in live yes, the fix is merged so it should start using player money again going forward

Issue #3077, opened 2022-03-06
expr terminal command returns incorrect value for remainder/modulo operator (%)


ChrissiQ:
`expr 121 % 2` returns `1212` instead of `1`.

<img width="249" alt="Screen Shot 2022-03-06 at 11 40 06 AM" src="https://user-images.githubusercontent.com/1127719/156937349-64d0baf7-037f-452b-903a-7df77f9c2dfb.png">

mbrannen:
The regex sanitization didn't allow that operator.  I created a PR with a fix, hopefully it's accepted.

PR #3093, opened 2022-03-08
updated regex sanitization

mbrannen:
-allows %
-updated escape characters to \

# Documentation

- ISSUE #3077   

# Bug fix

- tested in localhost

### DIFF
--- a/src/Terminal/commands/expr.ts
+++ b/src/Terminal/commands/expr.ts
@@ -17,7 +17,7 @@ export function expr(
   const expr = args.join("");
 
   // Sanitize the math expression
-  const sanitizedExpr = expr.replace(/s+/g, "").replace(/[^-()\d/*+.]/g, "");
+  const sanitizedExpr = expr.replace(/s+/g, "").replace(/[^-()\d\*+.\%]/g, "");
   let result;
   try {
     result = eval(sanitizedExpr);


ChrissiQ:
I haven't tested this but I think there's something wrong - those characters do **not** need escaping and now the regex does not match `/` for division.

mbrannen:
You were correct that they didn't need escaping.  I must have missed division the first time I tested.  Updated with a new commit and ive attached a screenshot this time of testing
image

mbrannen:
Another testing screenshot with some more complicated expressions
image

ChrissiQ:
Thank you! That looks a lot better. Just a minor note that I don't think `%` needs escaping either so the `\` is likely unnecessary.

I'm not sure if there could be repercussions from this change, since the result goes straight into `eval` - did you happen to check if this might enable new exploits?

ApamNapat:
> Thank you! That looks a lot better. Just a minor note that I don't think `%` needs escaping either so the `\` is likely unnecessary.
> 
> I'm not sure if there could be repercussions from this change, since the result goes straight into `eval` - did you happen to check if this might enable new exploits?

`\` is certainly unnecessary

src/Terminal/commands/expr.ts
```diff
@@ -17,7 +17,7 @@ export function expr(
   const expr = args.join("");
 
   // Sanitize the math expression
-  const sanitizedExpr = expr.replace(/s+/g, "").replace(/[^-()\d/*+.]/g, "");
+  const sanitizedExpr = expr.replace(/s+/g, "").replace(/[^-()\d*/+.\%]/g, "");
```
ApamNapat, line 20: When you remove the unnecessary escape and reordering that doesn't change anything as well, you and up with:

```suggestion
  const sanitizedExpr = expr.replace(/s+/g, "").replace(/[^-()\d/*+.%]/g, "");
```
Which is just adding `%`

ApamNapat (changes requested):

mbrannen:
> I'm not sure if there could be repercussions from this change, since the result goes straight into `eval` - did you happen to check if this might enable new exploits?

the eval function itself is a rather exploitable function to begin with, but the allowed characters from the regex expression are very narrow in scope.  If there are any vulnerabilities, I don't believe this would add any additional ones as % only functions as a remainder to my knowledge.

ApamNapat (approved):

PR #3095, opened 2022-03-08
added ram charge for stanek width and height

phyzical:
something noticed taking a stab at fixing the static ram calc issues


hydroflame:
missing NetscriptDefinitions.d.ts

phyzical:
image

i think its there, lmk if you mean something else 👍

hydroflame:
Ah, that's my real name :(

phyzical:
@hydroflame  ah sorry, itll be your git config on a computer somewhere (proably your work pc 😆) same thing happens with me when i switch between my home pc and work pc

PR #3096, opened 2022-03-08
Change license to Apache 2 with Commons Clause

danielyxie:


phyzical (approved):
lgtm, you want to charge for something your responsible for the damages from those changes. otherwise do what you want

PR #3099, opened 2022-03-09
Fixes #3098 missing space in Smart Supply

### LINKED ISSUE #3098: Error when checking for Smart Supply upgrade
https://github.com/danielyxie/bitburner/blob/570b8caf781e14fce0d5f410d0ec495931515a2e/src/NetscriptFunctions/Corporation.ts#L417

This line gives an error because the upgrade actually has a space between the two words.
It should be "Smart Supply" here, not "SmartSupply".
The same applies to the line below it and the two similar lines in the next function below this.

image

TheRealMaxion:
Quick fix to stop this from erroring due to the missing space in the upgrade name.


phyzical (approved):
good spot 👍

PR #3104, opened 2022-03-09
Attempt to fix `bladeburner.process()` early routing issue

MartinFournier:
### Attempt to fix `bladeburner.process()` issue

For a certain user, it appears that the engine attempts to route to the
bitverse screen before the router is properly initialized. I could not
reproduce the problem on my side and we were not able to extract his
save game, so this is a blind shot if it'll work or not.

I tried to replicate his situation by completing Daedalus blackop, waiting for an autosave and reloading the game, but I don't get the error on my side and I get redirected to the Bitverse screen successfully, both before and after the modification included in the PR.

This is a screenshot from the user, where he can't retrieve the save game and has the `Router called before initialization` error in the console.

IMG_3347

### DIFF
--- a/src/Bladeburner/Bladeburner.tsx
+++ b/src/Bladeburner/Bladeburner.tsx
@@ -1914,6 +1914,9 @@ export class Bladeburner implements IBladeburner {
   }
 
   process(router: IRouter, player: IPlayer): void {
+    // Edge race condition when the engine checks the processing counters and attempts to route before the router is initialized.
+    if (!router.isInitialized) return;
+
     // Edge case condition...if Operation Daedalus is complete trigger the BitNode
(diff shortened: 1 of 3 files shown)


phyzical (approved):
change looks sensible to me

hydroflame:
Are we still in touch with that user?

hydroflame:
Are we still in touch with that user?

hydroflame:
Maybe the real problem is that the code runs concurrently with the UI but shouldn't really

hydroflame:
Maybe the real problem is that the code runs concurrently with the UI but shouldn't really

MartinFournier:
@hydroflame The user is DeeK in discord https://discord.com/channels/415207508303544321/415213413745164318/951245212817195008

PR #3107, opened 2022-03-11
Add shortcut to Faction augmentations page from FactionsRoot

nickofolas:
This builds upon the QoL in #3046, turning the "Augmentations Left" text into a button that, when clicked, routes the user directly to the faction's respective augmentations page.

firefox_wNmW6Gqb8P

This also required a tiny adjustment to `router.toFaction` to allow it to take an argument. To accommodate this, I ended up making the `faction` parameter required, and updated the code where it had not been provided an argument to reflect this.

### Testing
For the UI change, I was able to verify that it worked, and that nothing broke when more factions were added.

For the `router.toFaction` change, I verified it by ensuring that, when `Player.process` is called, the game still properly routes to the expected faction page.

### DIFF
--- a/src/PersonObjects/Player/PlayerObjectGeneralMethods.tsx
+++ b/src/PersonObjects/Player/PlayerObjectGeneralMethods.tsx
@@ -590,7 +590,7 @@ export function process(this: IPlayer, router: IRouter, numCycles = 1): void {
   if (this.isWorking) {
     if (this.workType == CONSTANTS.WorkTypeFaction) {
       if (this.workForFaction(numCycles)) {
-        router.toFaction();
+        router.toFaction(Factions[this.currentWorkFactionName]);
       }
     } else if (this.workType == CONSTANTS.WorkTypeCreateProgram) {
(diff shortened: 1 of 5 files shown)


src/PersonObjects/Player/PlayerObjectGeneralMethods.tsx
```diff
@@ -590,7 +590,7 @@ export function process(this: IPlayer, router: IRouter, numCycles = 1): void {
   if (this.isWorking) {
     if (this.workType == CONSTANTS.WorkTypeFaction) {
       if (this.workForFaction(numCycles)) {
-        router.toFaction();
```
phyzical, line 593: careful with this, ive noticed references to Factions causes issues when augmenting as Factions[this.currentWorkFactionName] can return null, so the same tick you augment itll then throw the recovery screen.

phyzical: but int his case toFaction has a if check so i think this ones safe

PR #3114, opened 2022-03-11
Mixing spaces and tabs, extra semicolons and lonely ifs are now illegal

ApamNapat:
Mixing spaces and tabs is generally frowned upon (e.g. it's syntax error in Python 3) and there is no reason to keep the useless semicolons. The last one is a bit more a matter of style, but again, removing "noise", extra indent etc. while keeping the code doing the same thing is IMHO almost always a good idea


phyzical (approved):
lgtm

PR #3128, opened 2022-03-12
Escape forward slashes in regex

TheRealMaxion:
All documentation and regex testers I have found state that forward slashes ("/") inside regex patterns should be escaped, since the forward slash is used to define the start and end of the pattern.
The [ECMA standard](https://262.ecma-international.org/12.0/#sec-escaperegexppattern) only states that it "shall be escaped as necessary" for it to parse correctly, but I think the better practice is to always escape it, even inside a character class like it is here.
Other posts I have seen about it say it has to be escaped when in the /pattern/ notation but not in the RegExp constructor, but again, I think the best practice is to always escape it.
If there are regex patterns containing "/" elsewhere in the code they may yet need to also be corrected.

### DIFF
--- a/src/Corporation/Actions.ts
+++ b/src/Corporation/Actions.ts
@@ -91,7 +91,7 @@ export function SellMaterial(mat: Material, amt: string, price: string): void {
   if (price === "") price = "0";
   if (amt === "") amt = "0";
   let cost = price.replace(/\s+/g, "");
-  cost = cost.replace(/[^-()\d/*+.MPe]/g, ""); //Sanitize cost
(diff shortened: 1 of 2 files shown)


phyzical:
mind just adding a gif showing the updated functions still in action after the fix :)

ApamNapat (changes requested):
Those escapes are perfectly redundant and all they do is make the regexes less readable, which is a big issue with the whole thing. The parser is smart enough to know that `/` inside `[]` is not the end of the regex. I'd remove those if they were already in the code instead of adding them

TheRealMaxion:
Alright, it seemed to work fine without them, so that's fair enough I guess.

PR #3129, opened 2022-03-12
Fix unique ip generation

InDieTasten:
The previous implementation had a bug in it, where the generated IP
wouldn't be unique, if the generated IP wouldn't be unique two times in
a row.

Conducted tests:
Quick check in the game:
image

Test suite result:
image

### DIFF
--- a/src/Server/AllServers.ts
+++ b/src/Server/AllServers.ts
@@ -78,12 +78,11 @@ export function ipExists(ip: string): boolean {
 }
 
 export function createUniqueRandomIp(): string {
-  const ip = createRandomIp();
-
-  // If the Ip already exists, recurse to create a new one
-  if (ipExists(ip)) {
-    return createRandomIp();
-  }
+  let ip: string;
+  // Repeat generating ip, until unique one is found
+  do {
+    ip = createRandomIp();
+  } while (ipExists(ip));
 
   return ip;
 }


phyzical (approved):
LGTM, one thing though if for whatever reason the issue was the randomization the game would get stuck in an infinite loop, maybe add a max of like 15-20 tries?

the randomization looks to be this
https://github.com/danielyxie/bitburner/blob/dev/src/utils/helpers/getRandomInt.ts#L10

src/Server/AllServers.ts
```diff
@@ -78,12 +78,11 @@ export function ipExists(ip: string): boolean {
 }
 
 export function createUniqueRandomIp(): string {
-  const ip = createRandomIp();
-
-  // If the Ip already exists, recurse to create a new one
-  if (ipExists(ip)) {
-    return createRandomIp();
-  }
+  let ip: string;
```
ApamNapat, line 81:
```
  let ip = createRandomIp();
  // Repeat generating ip, until unique one is found
  while (ipExists(ip)) {
    ip = createRandomIp();
  }
```
No need to use a quite rare `do...while` when a more typical loop would suffice

ApamNapat (changes requested):

ApamNapat:
> LGTM, one thing though if for whatever reason the issue was the randomization the game would get stuck in an infinite loop, maybe add a max of like 15-20 tries?
> 
> the randomization looks to be this https://github.com/danielyxie/bitburner/blob/dev/src/utils/helpers/getRandomInt.ts#L10

`getRandomInt` is used all over the project, the whole game would break if it stopped working in such a simple case, this is too much caution IMHO

phyzical:
just to be clear im talking about the while loop generating the random ip address,

 not changing getRandomInt. just meant thats where its randomness resides

InDieTasten:
@phyzical I thought about it, but limiting the loop would add more complexity for a system (server IPs), that isn't really widely used.

The bug of non-unique ip being assigned was already rare, because it required the createRandomIp function to produce a duplicate twice in a row. With every iteration of the loop, the overall chance of continuing gets even slimmer, especially since we have such a large IP space with only a couple dozen servers being assigned an IP.

As for using a while loop instead of do...while, I can post an update in a couple hours.

PR #3131, opened 2022-03-12
Refresh "Create Programs" page UI

nickofolas:
This PR updates the UI for the program creation page with a couple of goals:
- Display more information at a glance
- Waste less screen space

<table>
  <th>Before</th>
  <th>After</th>
  <tr>
    <td>
      
    </td>
    <td>
      
    </td>
  </tr>
</table>

*Note: "Before" is shown with higher hacking level for comparison's sake*

The entire page is now wrapped in a `<Container>` to improve readability of the top paragraph.

The "Unlocks in:" value is found via the calculation used to determine if a program can be created, which takes `Player.intelligence` into account. Note: this shows a value rounded by `Math.ceil` for the sake of simplicity.

The "Current completion:" value is found in the same way that it's found when program work is being resumed, by parsing the value of the incomplete file on the user's computer.

By continuing to show program info, even after creation, it's easier for players to quickly remind themselves of what that program does, in the event they've forgotten. Completed programs are shown with 75% opacity, and a checkmark icon.

Programs are sorted by level requirement, and completed programs are always sorted to the end of the list.

Additionally, the `Create Programs` tab will now be available immediately, instead of being hidden until a program is available. **Rationale:** The dark web's `buy -l` already provides a list of all available programs for purchase, so it seems perfectly reasonable to provide the same information here.

Also, when the user doesn't meet the requirements for `BitFlume` (they're in BN 1.1), the `b1t_flum3.exe` program will not be shown.

**In BN1.1**
image

### Testing

- [x] I have tested that `b1t_flum3.exe` is properly hidden when it should be
- [x] I have tested that programs are locked until they are allowed to be created (the `Create program` button will not be shown)
- [x] I have verified that the page is properly reset upon prestige

### DIFF
--- a/src/Programs/ui/ProgramsRoot.tsx
+++ b/src/Programs/ui/ProgramsRoot.tsx
@@ -1,5 +1,5 @@
 import React, { useState, useEffect } from "react";
-import _ from "lodash";
+import { find } from "lodash";
 
 import {
   Box,
@@ -57,7 +57,7 @@ export function ProgramsRoot(): React.ReactElement {
   }
 
   const getProgCompletion = (name: string): number => {
-    const programFile = _.find(player.getHomeComputer().programs, p => {
+    const programFile = find(player.getHomeComputer().programs, p => {
       return (p.startsWith(name) && p.endsWith("%-INC"));
     });
     if (!programFile) return -1;
--- a/src/Programs/ui/ProgramsRoot.tsx
+++ b/src/Programs/ui/ProgramsRoot.tsx
@@ -1,9 +1,19 @@
 import React, { useState, useEffect } from "react";
-import { use } from "../../ui/Context";
-import { getAvailableCreatePrograms } from "../ProgramHelpers";
+import _ from "lodash";
 
-import { Box, Tooltip, Typography } from "@mui/material";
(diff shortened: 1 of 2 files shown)


phyzical:
i guess the one thing to note would be spoilers?

>The dark web's buy -l already provides a list of all available programs for purchase, so it seems perfectly reasonable to provide the same information here.
i think the darkweb is hidden away enough that new users dont just stumble across them, but i dont care too much tbh 😆

phyzical (approved):
lgtm

nickofolas:
> i guess the one thing to note would be spoilers?
> 
> > The dark web's buy -l already provides a list of all available programs for purchase, so it seems perfectly reasonable to provide the same information here.
> 
> i think the darkweb is hidden away enough that new users dont just stumble across them, but i dont care too much tbh laughing

fair, but like you said, it's probably not an issue. the only real "spoiler"-y one would be bitflume IMO, the rest aren't like, secrets or anything

besides, it might add more incentive to players to level up to see what they're working toward :D

phyzical:
true, at the same time less people might miss bit_flume entirely now (idk how that happens but it does 😆 )

PR #3137, opened 2022-03-15
Add the ability to filter open scripts

phyzical:
* icon that opens up on click and reset and closes on X
Mar-15-2022 20-38-55


nickofolas (approved):
looks good

phyzical:
@nickofolas while i have you, one thing i couldn't get right was making the text-inputs height to match the tab height 
image
 
after about an hour of fiddling with all kinda of settings i just gave up 😆

PR #3141, opened 2022-03-16
[Feature] Add ability to Bitverse and provide next bitnode

phyzical:
# Reasoning
was a suggestion in discord and i thought it was a good one to basically to allow further automation of the game

* added additional current bitnode display below bitburner version, only displays if any sourcefiles collected 
* optional param that will only have any effect when hacking the daemon, kept the param description vague on purpose to avoid spoilers
* also supports the operation method of bitversing
* has checks to enforce a bitnode between 1-13 inclusive, with a global int in the list of bitbnodes for easy updating
* added logic to the dev menu to allow for even faster node changing

Mar-16-2022 17-55-17

Mar-16-2022 21-13-02


TheRealMaxion:
If adding a way to automate selection of next bitnode, should also add a param for a script to run after entering the new bitnode, like for singularity resets/augment installs.

hydroflame:
PR made before mega fmt

phyzical:
@TheRealMaxion hmm good idea, ill have a look at what i did and see how complicated it becomes

phyzical:
i cant see an easy way to pull off also starting a script at this time, ill leave that up to the next person

hydroflame:
Ok so I'm going to close this.

The reason is it's trying to rope all the UI code in it. When really it should just be a new NS function that checks takes  a slightly different route.

PR #3144, opened 2022-03-16
Fix for BitNode prestige of corporation research

TheRealMaxion:
BitNode prestige currently prevents product industries from ever researching product research ever again.
This fixes that, allowing them to do product research after BitNode prestige.

(currently running saves that have already prestiged will have to prestige their bitnode after this fix is applied before their corporations will be able to do this research again)


phyzical:
@TheRealMaxion doesn't seem to be any changes in the diff?

pd:
This was fixed in https://github.com/danielyxie/bitburner/pull/2914 which merged yesterday.

TheRealMaxion:
Ah okay, thanks pd, closing this then.

PR #3156, opened 2022-03-17
Changed .keyCode to .key for files in "src"

### LINKED ISSUE #2308: Keyboard mappings and game play.
Bitburner v1.2.0 (4fb1dbf2) - Windows 10 - Steam

The script editing screen has changed the way it captures input.
The mapping for ctrl/cmd + s and ctrl/cmd + b now seem to capture the physical keycode and not the intended character.
The issue for me is that I use a dvorak layout on a qwerty keyboard.
 - The qwerty b and the dvorak x use the same key; no more using ctrl - x to cut text in a script file because it closes the editor.
 is it possible to make these keys configurable?

Master-Guy:
Changed .keyCode to .key for files in "src"

Did NOT change the files in "dist", or the files in the root directory ("edit.main.js", "main.bundle.js" and "main.bundle.js.map"). Not sure if these are required, and they also have very different keycodes in use compared to the rest.

Keyboard shortcuts are still working on my US-international layout, and should be tested by someone with a dvorak layout.
If approved, fixes #2308

tBC0vs7R1f


phyzical:
mind adding a small gif showing things working via keyboard

src/Sidebar/ui/SidebarRoot.tsx
```diff
@@ -276,54 +275,54 @@ export function SidebarRoot(props: IProps): React.ReactElement {
     function handleShortcuts(this: Document, event: KeyboardEvent): any {
       if (Settings.DisableHotkeys) return;
       if ((props.player.isWorking && props.player.focus) || redPillFlag) return;
-      if (event.keyCode == KEY.T && event.altKey) {
+      if (event.key === "t" && event.altKey) {
```
phyzical, line 278: how about instead we change the KEY constants to be these static strings, there are other devs that have a pet peeve about magic strings

Master-Guy: Added the GIF and changed it to the KEY constant.

phyzical (approved):
lgtm

PR #3158, opened 2022-03-17
Tabelize linked output of `ls`

Master-Guy:
Undocumented bug:
The clickable files in `ls` were not put in table form, even though it was expected them to be.
The other files and folders were correctly formatted.

image

The new setup does correctly form "tables":

image

### DIFF
--- a/src/Terminal/commands/ls.tsx
+++ b/src/Terminal/commands/ls.tsx
@@ -139,10 +139,12 @@ export function ls(
       }),
     )();
 
-    const rowSplit = row
-      .split(" ")
-      .map((x) => x.trim())
-      .filter((x) => !!x);
+    console.log(row);
+
+    const rowSplit = row.split("~");
+    let rowSplitArray = rowSplit.map((x) => [x.trim(), x.replace(x.trim(), "")]);
+    console.log(rowSplitArray);
+    rowSplitArray = rowSplitArray.filter((x) => !!x);
 
     function onScriptLinkClick(filename: string): void {
       if (player.getCurrentServer().hostname !== hostname) {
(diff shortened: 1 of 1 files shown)


src/Terminal/commands/ls.tsx
```diff
@@ -139,10 +139,12 @@ export function ls(
       }),
     )();
 
-    const rowSplit = row
-      .split(" ")
-      .map((x) => x.trim())
-      .filter((x) => !!x);
+    console.log(row);
```
ApamNapat, line 142: Are this `console.log` and the one below supposed to be merged or were they for debugging?

src/Terminal/commands/ls.tsx
```diff
@@ -139,10 +139,12 @@ export function ls(
...
+
+    const rowSplit = row.split("~");
+    let rowSplitArray = rowSplit.map((x) => [x.trim(), x.replace(x.trim(), "")]);
+    console.log(rowSplitArray);
+    rowSplitArray = rowSplitArray.filter((x) => !!x);
```
ApamNapat, line 147: When would this filter anything? `rowSplitArray`'s elements are two element arrays, so they are always truthy, aren't they?

ApamNapat (changes requested):

src/Terminal/commands/ls.tsx
Master-Guy: That should be removed, will do.

src/Terminal/commands/ls.tsx
Master-Guy: It was already in there, so I left it, but didn't change it to look at the first element of the array.

phyzical (approved):
lgtm

PR #3176, opened 2022-03-19
Fixed: purchaseSleeveAug checks shock value

incubusnb:
Edit: cancelled old PR and added this one because it was trying to add multiple commits and I just wanted one commit.

Edited Sleeve.ts to throw an error if a netscript function call tried to buy an augment when the sleeve's shock was higher than 0.

To test that this works I had a simple script try to buy an augment and checked the multipliers to see if any of them changed.


hydroflame:
Lint your file

incubusnb:
Oops, sorry, saw the error and assumed it was the Test that failed before. It's fixed.

TheRealMaxion:
This throws the error even when the sleeve shock is 0 in prod now and I'm not sure why.

TheRealMaxion:
Oh right, it's because 0 shock is 100 shock in the internal files.

TheRealMaxion:
And has already been fixed in an issue/pr so never mind.

PR #3192, opened 2022-03-19
Check before unlocking corp upgrade

gianfun:
# Bug fix

Can be tested by trying to call 'ns.corporation.unlockUpgrade' twice:

Ex:
```
ns.corporation.unlockUpgrade("Smart Supply")
ns.corporation.unlockUpgrade("Smart Supply")
```

This used to remove funds twice. Now it throws an error in the second try, similar to how `NewIndustry` works.

Tested by running locally.

Was unsure about whether to put this in 'Actions.ts' (where we have some checks) or in 'Corporation.tsx', do you have a preference?


phyzical (approved):
lgtm

phyzical:
i think the action is the right place as the ui uses it too, even though thats probably checking correctly

PR #3193, opened 2022-03-19
Implement getBonusTime in Corporation API

t-wolfeadam:
This simply implements a `getBonusTime` method for the Corporation API, similar to the ones for the bladeburner and gang APIs.

Usage example:

```javascript
	while (true) {
		let btime = ns.corporation.getBonusTime();
		console.log(ns.nFormat(btime, "00:00:00"));
		await ns.sleep(50);
	}
}
```

Resulting in:

bb-corp-api-demo-optimz


phyzical (approved):
lgtm

PR #3195, opened 2022-03-19
Fix #3194: include all division employees in small town achievement calc

### LINKED ISSUE #3194: Small town achievement (Corp 3000 employees) calculation is incorrect
The achievement says `Have a division with 3000 employees`, but the calculation doesn't sum up all the employees in the division, it just looks at one individual city.  I found out because I have 3,240 employees in a division (540 employees in each city: 540*6 = 3,240) but the achievement wasn't popping.

Should be a pretty easy fix, just sum the employee count for all cities within a division.
image

BrianLDev:
fixes #3194 

Quick and easy fix to this achievement calc that wasn't including all employees in the division in the calc.  Now it sums up the combined division employee count.

### DIFF
--- a/src/Achievements/Achievements.ts
+++ b/src/Achievements/Achievements.ts
@@ -458,10 +458,12 @@ export const achievements: IMap<Achievement> = {
     Condition: (): boolean => {
       if (Player.corporation === null) return false;
       for (const d of Player.corporation.divisions) {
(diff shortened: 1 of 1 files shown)


phyzical (approved):
lgtm

hydroflame:
Well ..... the achievement is wrong.

PR #3197, opened 2022-03-20
Initial API Wrapper Work

TheMas3212:
onlying wrapping Stanek api currently as test sample
had to change how it worked slightly as it the initial plan didnt quite work
every function in stanek (and a selection of other functions as well) has been tested and confirmed as still functional

### DIFF
--- a/src/Netscript/APIWrapper.ts
+++ b/src/Netscript/APIWrapper.ts
@@ -10,9 +10,7 @@ type ExternalAPI = {
   [string: string]: ExternalAPI | ExternalFunction;
 };
 
-type InternalFunction<F extends (...args: unknown[]) => unknown> = (
-  ctx: NetscriptContext,
-) => (...args: unknown[]) => ReturnType<F>;
+type InternalFunction<F extends (...args: unknown[]) => unknown> = (ctx: NetscriptContext) => F;
 export type InternalAPI<API> = {
   [Property in keyof API]: API[Property] extends ExternalFunction
     ? InternalFunction<API[Property]>
@@ -23,7 +21,7 @@ export type InternalAPI<API> = {
 
 type WrappedNetscriptFunction = (...args: unknown[]) => unknown;
 type WrappedNetscriptAPI = {
-  [string: string]: WrappedNetscriptAPI | WrappedNetscriptFunction;
+  readonly [string: string]: WrappedNetscriptAPI | WrappedNetscriptFunction;
 };
 
 export type NetscriptContext = {
@@ -47,7 +45,6 @@ type NetscriptHelpers = {
 };
 
 type WrappedNetscriptHelpers = {
-  updateDynamicRam: (ramCost: number) => void;
   makeRuntimeErrorMsg: (msg: string) => string;
   string: (argName: string, v: unknown) => string;
   number: (argName: string, v: unknown) => number;
@@ -80,7 +77,6 @@ function wrapFunction(
     workerScript,
     function: functionName,
     helper: {
-      updateDynamicRam: (ramCost: number) => helpers.updateDynamicRam(functionName, ramCost),
       makeRuntimeErrorMsg: (msg: string) => helpers.makeRuntimeErrorMsg(functionPath, msg),
       string: (argName: string, v: unknown) => helpers.string(functionPath, argName, v),
       number: (argName: string, v: unknown) => helpers.number(functionPath, argName, v),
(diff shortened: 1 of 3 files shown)


hydroflame:
PR made before mega fmt

src/Netscript/APIWrapper.ts
```diff
@@ -0,0 +1,157 @@
...
+import { Player } from "../Player";
+
+type ExternalFunction = (...args: any[]) => any;
+type ExternalAPI = {
+  [string: string]: ExternalAPI | ExternalFunction;
+};
+
+type InternalFunction<F extends (...args: unknown[]) => unknown> = (
+  ctx: NetscriptContext,
+) => (...args: unknown[]) => ReturnType<F>;
```
hydroflame, line 15: nitpick `...args: Parameters<F>` is more robust I think.

hydroflame: heck this entire declaration can be simplified as 
```ts
type InternalFunction<F extends (...args: unknown[]) => unknown> = (
  ctx: NetscriptContext,
) => F;
```

src/Netscript/APIWrapper.ts
```diff
@@ -0,0 +1,157 @@
...
+    ? InternalAPI<API[Property]>
+    : never;
+};
+
+type WrappedNetscriptFunction = (...args: unknown[]) => unknown;
+type WrappedNetscriptAPI = {
+  [string: string]: WrappedNetscriptAPI | WrappedNetscriptFunction;
+};
+
+export type NetscriptContext = {
```
hydroflame, line 29: nitpick, some of these field should be marked as readonly

src/Netscript/APIWrapper.ts
```diff
@@ -0,0 +1,157 @@
...
+    makeRuntimeErrorMsg: (message: string) => {
+      return helpers.makeRuntimeErrorMsg(functionPath, message);
+    },
+    log: (message: () => string) => {
+      workerScript.log(functionPath, message);
+    },
+    workerScript,
+    function: functionName,
+    helper: {
+      updateDynamicRam: (ramCost: number) => helpers.updateDynamicRam(functionName, ramCost),
```
hydroflame, line 83: I think this should be `updateDynamicRam(functionPath` ?

hydroflame: oh nvm that is not how the current implementation works ... but should it be?

src/Netscript/APIWrapper.ts
```diff
@@ -0,0 +1,157 @@
...
+    },
+  };
+  function wrappedFunction(...args: unknown[]): unknown {
+    helpers.updateDynamicRam(ctx.function, getRamCost(Player, ...tree, ctx.function));
+    return func(ctx)(...args);
+  }
+  const parent = getNestedProperty(wrappedAPI, ...tree);
+  Object.defineProperty(parent, functionName, {
+    value: wrappedFunction,
+    writable: true,
```
hydroflame, line 101: false probably, I didn't know about defineProperty, it's cool.

src/Netscript/APIWrapper.ts
```diff
@@ -0,0 +1,157 @@
...
+      case "function": {
+        wrapFunction(helpers, wrappedAPI, workerScript, namespace[property], ...tree, property);
+        break;
+      }
+      case "object": {
+        wrapAPI(helpers, wrappedAPI, workerScript, namespace[property], ...tree, property);
+        break;
+      }
+      default: {
+        setNestedProperty(wrappedAPI, namespace[property], ...tree, property);
```
hydroflame, line 126: This sends 4 arguments but the function declares 3 arguments

src/Netscript/APIWrapper.ts
TheMas3212: `writable` has to be true actually or the other wrapper fails

src/Netscript/APIWrapper.ts
TheMas3212: last arguent in setNestedProperty is a rest argument, so it takes all of tree and property
ie tree = ['a','b','c']; property = 'd';
inside the setNestedProperty call the tree argument contains ['a','b','c','d']

src/Netscript/APIWrapper.ts
TheMas3212: thinking about it since the wrapper calls updateDynamicRam itself theres no reason to include it in the helpers in context
so il actually just remove that

src/Netscript/APIWrapper.ts
TheMas3212: adding a commit in a moment

src/Netscript/APIWrapper.ts
TheMas3212: adding a commit in a moment

PR #3199, opened 2022-03-20
purchaseSleeveAug

phyzical:
used the getSleeveStats as it seems to be a miss connection between whats displayed and whats used

i.e sleeve.shock of 100 is actually a shock value of 0 for the use


smmalis37:
Fixes #3202 lol

PR #3201, opened 2022-03-20
Fix #3198: give player code a copy of Division.upgrades instead of the live object

### LINKED ISSUE #3198: Corporation API leaks live Division.upgrades object
`getSafeDivision` (in https://github.com/danielyxie/bitburner/blob/dev/src/NetscriptFunctions/Corporation.ts#L277) passes through the live `Division.upgrades` field, which is an array. Which means player code can edit the number of Coffee and AdVert.Inc that have been bought.

AdVert.Inc is a problem, as it's price scales based on number of times it's been bought, but it's effect is permanently applied to other stats when it's bought.

Sample exploit to repeatedly buy 1b cost AdVert.Inc:
```js
/** @param {NS} ns **/
	const divisionName = ns.args[0];
	const division = ns.corporation.getCorporation().divisions.find(division => division.name == divisionName);

	division.upgrades[1] = 0; // Reset AdVert.Inc to initial 1b cost

	ns.corporation.hireAdVert(divisionName); // Buy one AdVert.Inc, this should raise cost

	division.upgrades[1] = 0; // Reset again
	if (ns.corporation.getHireAdVertCost(divisionName) != 1e9) // Check if exploit still worked
		throw new Error("Exploit: fixed");
}
```

Ornedan:
Fixed by making a `slice()` copy, rather than giving the live game state object to player code. Sample exploit from #3198 no longer works:

```js
/** @param {NS} ns **/
	const divisionName = ns.args[0];
	const division = ns.corporation.getCorporation().divisions.find(division => division.name == divisionName);

	division.upgrades[1] = 0; // Reset AdVert.Inc to initial 1b cost

	ns.corporation.hireAdVert(divisionName); // Buy one AdVert.Inc, this should raise cost

	division.upgrades[1] = 0; // Reset again
	if (ns.corporation.getHireAdVertCost(divisionName) != 1e9) // Check if exploit still worked
		throw new Error("Exploit: fixed");
}
```

localhost_8000__noscripts (3)

### DIFF
--- a/src/NetscriptFunctions/Corporation.ts
+++ b/src/NetscriptFunctions/Corporation.ts
@@ -291,7 +291,7 @@ export function NetscriptCorporation(
       lastCycleExpenses: division.lastCycleExpenses,
       thisCycleRevenue: division.thisCycleRevenue,
       thisCycleExpenses: division.thisCycleExpenses,
-      upgrades: division.upgrades,
+      upgrades: division.upgrades.slice(),
       cities: cities,
       products: division.products === undefined ? [] : Object.keys(division.products),
     };


phyzical (approved):

Ornedan:
On further thought, should `Division.upgrades` even be exposed to player code? The contents are unexplained, and require a code dive to figure out. Coffee buy count is not really useful, and advert buy count is exposed separately via `ns.corporation.getHireAdVertCount`.

So a better solution might be to just remove the field from the Netscript-visible Division object altogether?

hydroflame:
Well for now lets at least prevent an easy exploit

PR #3218, opened 2022-03-21
Nerf corporation investment fraud

### LINKED ISSUE #2350: Corporation tuning needed.
As per discord: 

There are currently corps are in a very unbalanced state, with gameplay that trivializes the need for any hacking, scripting, or other mechanics. You are able to get massive evaluations by gaming your income to be able to hyper-scale your corp and beat almost any node in less than 6 hours with zero other mechanics used. 

Steps to reproduce 

1. Start software
2. Buy smart supply
3. get all 6 cities with warehouses for 6x production multi
4. get 900 storage in each city
5. horde AI limiters (set sell to 0 at MP so you hold them all)
6. place works as 1 operation / 1 engineer / 1 manager / 3 research
7. when storage fills, quickly flip all AI limiters to sell MAX at MP
8. when income skyrockets, get a high evaluation.
repeat this process (max storage, stockpile, sell off in 1 market tick) to get a second huge bump of money. After that you can play normally (boost averts/production/worker counts). Since you get a huge influx of cash even keeping 10% of your total company becomes viable with how fast you scale.

My suggestion is setting the investor to take an average of your income over the last 1-5 minutes to help level out this kind of play, but there are probably other ways that would work and not completely destroy the clever use of mechanics.

rj45:
Corporation valuations now look at the average profit over the last `ValuationProfitHistoryCycles` cycles. Currently this is set at 10 which leaves investment fraud still viable, just around 10 times less profitable.

This constant could be tweaked to make the nerf more or less severe. This mainly affects how quickly corps progress, it shouldn't affect dividends or stock price. 

Fixes #2350 

Currently play testing this to see how it compares to before. I can report some numbers here when the results come in.

### DIFF
--- a/src/Corporation/Corporation.tsx
+++ b/src/Corporation/Corporation.tsx
@@ -170,7 +170,10 @@ export class Corporation {
     let val;
 
     // take the average over the last ValuationProfitHistoryCycles
-    let profit = this.profitHistory.reduce((sum, val) => sum+val, 0) / this.profitHistory.length;
+    let profit = 0;
+    if (this.profitHistory.length > 0) {
+      profit = this.profitHistory.reduce((sum, val) => sum+val, 0) / this.profitHistory.length;
+    }
 
     if (this.public) {
       // Account for dividends
--- a/src/Corporation/Corporation.tsx
+++ b/src/Corporation/Corporation.tsx
@@ -110,7 +110,9 @@ export class Corporation {
         const profit = this.revenue - this.expenses;
 
         this.profitHistory.push(profit);
-        if (this.profitHistory.length > 60) this.profitHistory.shift();
+        if (this.profitHistory.length > CorporationConstants.ValuationProfitHistoryCycles) {
+          this.profitHistory.shift();
+        }
 
         const cycleProfit = profit * (marketCycles * CorporationConstants.SecsPerMarketCycle);
         if (isNaN(this.funds) || this.funds === Infinity || this.funds === -Infinity) {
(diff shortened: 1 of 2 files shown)


src/Corporation/Corporation.tsx
```diff
@@ -159,8 +165,18 @@ export class Corporation {
   }
 
   determineValuation(): number {
-    let val,
-      profit = this.revenue - this.expenses;
+    let val;
+
+    // take the average over the last 60 cycles (10 min)
+    const avgProfit = this.profitHistory.reduce((sum, val) => sum+val, 0) / this.profitHistory.length;
```
phyzical, line 171: maybe a check to avoid divide by 0 even though it would be very rare

phyzical:
seems so sensible but wont approve simply due to not really knowing if its the correct way to approach the problem

rj45:
So I modified this PR a bit after talking in discord a bit about it. I put the amount of history it looks at as constant so it can be tweaked to tweak balance. I set it to 10, but maybe a better initial value would be 3 or 5. This will keep investment fraud as a viable mechanic, just not as over powered as it was.

@phyzical should I try to remove the formatting updates from this PR? I didn't notice it did that, should be easy enough to revert if wanted.

rj45:
@phyzical I just reverted the formatting updates anyway, and I fixed the divide by zero condition

src/Corporation/Corporation.tsx
```diff
@@ -159,8 +167,14 @@ export class Corporation {
...
 
   determineValuation(): number {
-    let val,
-      profit = this.revenue - this.expenses;
+    let val;
+
+    // take the average over the last ValuationProfitHistoryCycles
+    let profit = 0;
+    if (this.profitHistory.length > 0) {
+      profit = this.profitHistory.reduce((sum, val) => sum+val, 0) / this.profitHistory.length;
```
ApamNapat, line 175:
```suggestion
      profit = this.profitHistory.reduce((sum, next) => sum + next, 0) / this.profitHistory.length;
```

hydroflame:
Closed because implemented slightly differently

PR #3219, opened 2022-03-21
fix to log manager take 2

### LINKED ISSUE #1775: Tail window won't pick up new run when running the same script from terminal.

phyzical:
added logic to pick up log when its closed via event emitter

closes #1775
Mar-21-2022 19-59-28

### DIFF
--- a/src/ui/React/LogBoxManager.tsx
+++ b/src/ui/React/LogBoxManager.tsx
@@ -18,6 +18,7 @@ import { Theme } from "@mui/material";
 import { findRunningScript } from "../../Script/ScriptHelpers";
 import { Player } from "../../Player";
 import { debounce } from "lodash";
+import { WorkerScriptStartStopEventEmitter } from "../../Netscript/WorkerScriptStartStopEventEmitter";
 
 let layerCounter = 0;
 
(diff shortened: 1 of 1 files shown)


src/ui/React/LogBoxManager.tsx
```diff
@@ -128,6 +129,23 @@ function LogWindow(props: IProps): React.ReactElement {
     setRerender((old) => !old);
   }
 
+  useEffect(
+    () =>
+      WorkerScriptStartStopEventEmitter.subscribe(() => {
+        setTimeout(() => {
```
hydroflame, line 135: Why the timeout?

phyzical: for some reason without it it didnt update i think it might be a race condition between the script starting up and being found as a "running script"

GhostCrab:
This change was merged in commit 6905651b226a4a09941cfe98a08b2bb45993a16a to the dev branch, and after that, I'm getting a bug where the game locks up when I tail a script. I'm still working on simple steps to reproduce, but it looks like the script starts over and over again or logs are repeated constantly until the game stops running. When I check out the commit before this (4c64863314352d7dc6ee9a8b22de5efd92783ca8) and build, I do not see this error.

PR #3220, opened 2022-03-21
add similar sort to the faction joined list

### LINKED ISSUE #3214: Regression with sorted faction display
- Web version: Bitburner v1.5.0 (0171fc23)
- Steps to reproduce: Join some factions

The ordering of factions has changed. It looks reversed.

image

phyzical:
closes #3214
Mar-21-2022 21-23-15

### DIFF
--- a/src/Faction/ui/FactionsRoot.tsx
+++ b/src/Faction/ui/FactionsRoot.tsx
@@ -82,7 +82,8 @@ export function FactionsRoot(props: IProps): React.ReactElement {
   }
 
   const allFactions = Object.values(FactionNames).map(faction => faction as string)
-  const allJoinedFactions = props.player.factions.sort((a, b) =>
+  const allJoinedFactions = props.player.factions.slice(0);
+  allJoinedFactions.sort((a, b) =>
     allFactions.indexOf(a) - allFactions.indexOf(b));
 
   return (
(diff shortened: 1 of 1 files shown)


src/Faction/ui/FactionsRoot.tsx
```diff
@@ -79,6 +79,9 @@ export function FactionsRoot(props: IProps): React.ReactElement {
       (augmentation: string) => !player.hasAugmentation(augmentation)
     ).length;
   }
+  let allJoinedFactions = props.player.factions
+  allJoinedFactions = allJoinedFactions.sort((a, b) =>
+    allJoinedFactions.indexOf(a) - allJoinedFactions.indexOf(b));
```
iTitus, line 84: the sorting works in-place: see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort
so the use of the return value is unnecessary.
it might also be good to copy the array instead of using `props.player.factions` directly

did you want to use `allFactionsThatExist.indexOf()` ?

iTitus: Is it intended that you change the `props.player.factions` array directly from UI code instead of working with a copy?

phyzical: hmm but by calling sort dont we get a new copy returned?

src/Faction/ui/FactionsRoot.tsx
```diff
@@ -80,6 +81,10 @@ export function FactionsRoot(props: IProps): React.ReactElement {
     ).length;
   }
 
+  const allFactions = Object.values(FactionNames).map(faction => faction as string)
+  const allJoinedFactions = props.player.factions.sort((a, b) =>
```
phyzical, line 85:
```suggestion
  const allJoinedFactions = props.player.factions.slice(0)
  allJoinedFactions.sort((a, b) =>
```

PR #3221, opened 2022-03-21
Move main bundle output to ./dist subfolder

MartinFournier:
Excludes the index.html so that the github page can still work as is.
This will help ensure contributors don't commit the built bundles.

It's the same PR content as #2825 that's been merged before, but reverted. I think it was reverted by mistake?

If merged, once built you'll need to manually delete the root's bundles for the first time like so:
ConEmu64_jNPa4cri0q

- Tested while using `npm run start:dev`
- Tested with `npm run build && npm run start`
- Tested with `npm run build && npm run electron && .build/bitburner-win32-x64/bitburner.exe`

### DIFF
--- a/package.json
+++ b/package.json
@@ -123,7 +123,7 @@
     "test:watch": "jest --watch",
     "watch": "webpack --watch --mode production",
     "watch:dev": "webpack --watch --mode development",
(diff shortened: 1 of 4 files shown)


phyzical (approved):
lgtm

PR #3232, opened 2022-03-22
Add `npm version` support to build releases

MartinFournier:
Replaces the `npm run allbuild` script with the `npm version` command.
It creates a new commit with a standardized message & includes the built
app. 

Usage: `npm version [patch | minor | version | 10.0.0 | etc]` ([documentation](https://docs.npmjs.com/cli/v6/commands/npm-version))

If you want to include a changelog and/or increment the save game version for migrations, you need to modify the files before running the command. You'll have to add the force flag to npm version ie: `npm version patch --force` since it'll complain about an unclean repository otherwise.

Will update the package.json & perform a bunch of steps in a row:

- Runs `npm install` in:
  - The root directory
  - The ./electron/ directory
  - The ./tools/bump-version/ directory
- Runs `npm run test`
- Update the version string in:
  - ./electron/package.json (version)
  - ./src/Constants.ts (VersionString)
  - ./doc/source/conf.py (version & release)
- Runs `npm run doc` (build the markdown documentation)
- Runs `npm run build` (production build)
- Runs `npm run electron` (electron build)
- Runs `git add --all`
- Runs `git push -u origin dev && git push --tags`

---

### Testing

Result commit of `npm version major` : https://github.com/MartinFournier/bitburner/commit/2d44fe9a82103afc76adafa14a013e694b2340d0

### DIFF
--- a/package.json
+++ b/package.json
@@ -129,6 +129,9 @@
     "electron:packager-win": "electron-packager .package bitburner --platform win32 --arch x64 --out .build --overwrite --icon .package/icon.png",
     "electron:packager-mac": "electron-packager .package bitburner --platform darwin --arch x64 --out .build --overwrite --icon .package/icon.png",
     "electron:packager-linux": "electron-packager .package bitburner --platform linux --arch x64 --out .build --overwrite --icon .package/icon.png",
-    "allbuild": "npm run build && npm run electron && git add --all && git commit -m \"allbuild commit $(git rev-parse --short HEAD)\" && git push -f -u origin dev"
+    "allbuild": "npm run build && npm run electron && git add --all && git commit -m \"allbuild commit $(git rev-parse --short HEAD)\" && git push -f -u origin dev",
+    "preversion": "npm install && npm run test",
+    "version": "sh ./tools/build-release.sh && git add --all",
+    "postversion": "git push -u origin dev && git push --tags"
   }
 }
--- a/tools/build-release.sh
+++ b/tools/build-release.sh
@@ -1,0 +1,46 @@
+#!/bin/sh
+
+# See https://gist.github.com/mohanpedala/1e2ff5661761d3abd0385e8223e16425
+set -euxo pipefail
+
+APP_VERSION=$(npm pkg get version | sed 's/"//g')
+
+echo ''
+echo '|============================================|'
+echo "  Preparing build v$APP_VERSION"
+echo '  Taken from the root ./package.json file.'
+echo '|============================================|'
+echo ''
+sleep 1
+echo 'Do you need to increment the save file version?'
+echo 'If you do, cancel/exit this script now.'
+echo 'Update the following file before running this script: '
+echo '> ./src/Constants.ts'
+echo ''
+sleep 2
+echo 'Do you want to include a changelog with this release?'
+echo 'If you do, cancel/exit this script now.'
+echo 'Update the following file before running this script: '
+echo '> ./src/Constants.ts'
+echo '> ./doc/source/changelog.rst'
+echo ''
+sleep 2
+
+cd ./tools/bump-version
+npm ci
+cd ../../
+node ./tools/bump-version/basic.js --version $APP_VERSION
+echo ''
+
+echo 'Building bundles in 2 seconds...'
+sleep 2
+
+echo ''
+echo 'Starting build...'
+npm run doc && \
+  npm run build && \
+  npm run electron
+
+echo ''
+echo 'Completed!'
+echo ''
--- a/tools/bump-version/basic.js
+++ b/tools/bump-version/basic.js
@@ -1,0 +1,57 @@
+import commandLineArgs from "command-line-args";
+
+import fs from "fs/promises";
+import path from "path";
+import { fileURLToPath } from "url";
+
+// https://github.com/75lb/command-line-args
+const optionDefinitions = [
+  {
+    name: "version",
+    alias: "v",
+    type: String,
+    required: true,
+  },
+];
+
+const cliArgs = commandLineArgs(optionDefinitions);
+
+const root = path.join(path.dirname(fileURLToPath(import.meta.url)), "../..");
+const appPaths = {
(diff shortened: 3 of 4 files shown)


package.json
```diff
@@ -129,6 +129,8 @@
     "electron:packager-win": "electron-packager .package bitburner --platform win32 --arch x64 --out .build --overwrite --icon .package/icon.png",
     "electron:packager-mac": "electron-packager .package bitburner --platform darwin --arch x64 --out .build --overwrite --icon .package/icon.png",
     "electron:packager-linux": "electron-packager .package bitburner --platform linux --arch x64 --out .build --overwrite --icon .package/icon.png",
-    "allbuild": "npm run build && npm run electron && git add --all && git commit -m \"allbuild commit $(which git)\" && git push -f -u origin dev"
+    "preversion": "npm install && npm run test",
+    "version": "sh ./tools/build-release.sh && git add --all",
```
ApamNapat, line 133: Why not something like `build` or even keeping the old name? I'd expect `version` to print the current version or something like that

package.json
```diff
@@ -129,6 +129,8 @@
...
+    "postversion": "git push -u origin dev && git push --tags"
```
ApamNapat, line 134: IMHO `git add --all` from the command above should be here

tools/build-release.sh
```diff
@@ -0,0 +1,43 @@
...
+echo '> ./src/Constants.ts'
+echo ''
+sleep 2
+echo 'Do you want to include a changelog with this release?'
+echo 'If you do, cancel/exit this script now.'
+echo 'Update the following file before running this script: '
+echo '> ./src/Constants.ts'
+echo '> ./doc/source/changelog.rst'
+echo ''
+sleep 2
```
ApamNapat, line 24: Why waste 5 seconds every single time this is run?

tools/build-release.sh
```diff
@@ -0,0 +1,43 @@
+#!/bin/sh
```
ApamNapat, line 1: Consider adding  `set -e, -u, -x, -o pipefail` as explained here https://gist.github.com/mohanpedala/1e2ff5661761d3abd0385e8223e16425

tools/build-release.sh
ApamNapat: Actually 7, as there is one more `sleep` below

tools/bump-version/basic.js
```diff
@@ -0,0 +1,47 @@
+// import { Octokit } from "@octokit/rest";
```
ApamNapat, line 1: Can't this be removed?

tools/bump-version/basic.js
```diff
@@ -0,0 +1,47 @@
...
+  let modifiedSphinxConfig = (await fs.readFile(appPaths.sphinxConf, 'utf8')).
+    replace(/(^version = ')(.*)'$/m, `$1${shortVersion}'`);
+  modifiedSphinxConfig = modifiedSphinxConfig.
+    replace(/(^release = ')(.*)'$/m, `$1${version}'`);
+  await fs.writeFile(appPaths.sphinxConf, modifiedSphinxConfig);
+  console.log(`> Modified ${appPaths.sphinxConf}`);
+}
+
+main(cliArgs.version).then(() => console.log('Done'));
+
```
ApamNapat, line 47: Consider running prettier on your code to ease its integration

tools/bump-version/basic.js
```diff
@@ -0,0 +1,47 @@
...
+  let modifiedSphinxConfig = (await fs.readFile(appPaths.sphinxConf, 'utf8')).
+    replace(/(^version = ')(.*)'$/m, `$1${shortVersion}'`);
+  modifiedSphinxConfig = modifiedSphinxConfig.
+    replace(/(^release = ')(.*)'$/m, `$1${version}'`);
+  await fs.writeFile(appPaths.sphinxConf, modifiedSphinxConfig);
+  console.log(`> Modified ${appPaths.sphinxConf}`);
+}
+
+main(cliArgs.version).then(() => console.log('Done'));
```
ApamNapat, line 46: I'd add a handler for failure

tools/bump-version/basic.js
```diff
@@ -0,0 +1,47 @@
...
+
+// https://github.com/75lb/command-line-args
+const optionDefinitions = [
+  { name: 'version', alias: 'v', type: String, required: true },
+];
+
+const cliArgs = commandLineArgs(optionDefinitions);
+
+const appPaths = {};
+appPaths.root = path.join(path.dirname(fileURLToPath(import.meta.url)), "../..");
```
ApamNapat, line 16: Why do it like this instead of constructing the object in one go `{"root": ..., "electronPackage": ...}`

package.json
MartinFournier: `npm version` is a built in npm command with its preversion & postversion hooks. See the [npmjs doc](https://docs.npmjs.com/cli/v6/commands/npm-version). If we'd keep `allbuild` or `build` we would need to forward command's args.

tools/build-release.sh
MartinFournier: Since this is done non-frequently, the sleep allows the publisher to cancel out of the build if he notices he should update the changelog or save migration number before running the command. IMO waiting 7 secs for something done once a day is fine.

tools/bump-version/basic.js
MartinFournier: Hmm, I think we need failure to throw an exception so that node returns a non-zero status code which will interrupt the `npm version` flow & not run the postversion hook.

package.json
MartinFournier: npm creates the commit right at the end version, but before running postversion. 
So the git add --all needs to be in the "version" hook otherwise the commit won't contain the files.

phyzical (approved):
looks like build scripts to me

package.json
ApamNapat: You learn something new every day

PR #3242, opened 2022-03-23
Add 0xDEADBEEF exploit (SF-1)

MartinFournier:
Adds a new exploit that requires the player to bruteforce a passphrase through the terminal
(or most likely read the source to reverse the seed method since bruteforce would take forever).

Adds a new `unlock` terminal command that adds a new server to the network once a
passphrase was successfully matched. This needs to be done from the
terminal, no direct ns script access.

Once created, the server is added as a hidden server which hides it from
scan, scan-analyze, autocomplete, ns.scan. This server is still able to be
connected to given the proper route.

Since the hostname is hex, it trips the connect method and
requires the player to wrap its hostname in quotes.
Once that server is backdoored, the exploit will be awarded and
the achievement unlocked.

---

### Unlocking the server

firefox_wQcyZMtA5i

### Accessing the server

firefox_oJ8n9UqMmy

### Backdooring the server
firefox_daJfpB6AcF

### Server not visible in scans

firefox_FcjYCfqhh1

### Direct connect works

firefox_WExl1wtZmL

### Backdooring the server

firefox_rVyUO4P3f6

### Current passphrase in dev menu

firefox_hVC9N6GIAs

### Achievement
firefox_b0T9ulf5A9

### DIFF
--- a/src/Achievements/Achievements.ts
+++ b/src/Achievements/Achievements.ts
@@ -746,6 +746,16 @@ export const achievements: IMap<Achievement> = {
     Secret: true,
     Condition: () => Player.exploits.includes(Exploit.TrueRecursion),
   },
+  DEADBEEF: {
+    // Replace the ID, Name & Description by the line below once the achievement is added to Steam.
+    // ...achievementData["DEADBEEF"],
+    ID: "DEADBEEF",
+    Name: "Exploit: 0xDEADBEEF",
+    Description: "Overpowered the obfuscated.",
+    Icon: "SF-1",
+    Secret: true,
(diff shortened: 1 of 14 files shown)


phyzical:
aw why did you close? was a cool idea i thought

MartinFournier:
Meh, it felt a bit convoluted and all over the place. Also messing with servers could possibly cause eisenbigsissues

phyzical:
yeah fair enough, but at the same time i wonder if having a "hidden" server prop would be useful in the future for other shenanigans @hydroflame ?

MartinFournier:
Alright sure, I'll keep the branch up & reopen this later once the backlog is cleared.

PR #3246, opened 2022-03-24
slice out event key constant refactors

phyzical:
a slice out of https://github.com/danielyxie/bitburner/pull/3241


src/Infiltration/utils.ts
```diff
@@ -1,21 +1,46 @@
+import { KEY } from "../utils/helpers/keyCodes";
+
 export function random(min: number, max: number): number {
   return Math.random() * (max - min) + min;
 }
 
+export const upArrowSymbol = "↑";
+export const downArrowSymbol = "↑";
+export const leftArrowSymbol = "↑";
+export const rightArrowSymbol = "↑";
```
phyzical, line 10:
```suggestion
export const upArrowSymbol = "↑";
export const downArrowSymbol = "↓";
export const leftArrowSymbol = "←";
export const rightArrowSymbol = "→";
```

PR #3247, opened 2022-03-24
Accept valid JSON arrays in coding contracts

Savlik:
I did not test it, but the code copied from another contract type, so hopefully, it will work 🤞


Ornedan:
What problem are you trying to fix with this change?

Savlik:
Well I want to submit the resulting string using `JSON.stringify` but it is not accepted because it doesn't sanitize quotes. So for example `[()()(), (())()]` is accepted but `["()()()", "(())()"]` is not.

PR #3249, opened 2022-03-24
Log info for field analysis now displays actual rank gained

ApamNapat:


src/Bladeburner/Bladeburner.tsx
```diff
@@ -1466,20 +1466,19 @@ export class Bladeburner implements IBladeburner {
...
+        this.changeRank(player, rankGain);
         this.getCurrentCity().improvePopulationEstimateByPercentage(eff * this.skillMultipliers.successChanceEstimate);
         if (this.logging.general) {
           this.log(
-            "Field analysis completed. Gained 0.1 rank, " +
-              formatNumber(hackingExpGain, 1) +
-              " hacking exp, and " +
-              formatNumber(charismaExpGain, 1) +
-              " charisma exp",
+            `Field analysis completed. Gained ${formatNumber(rankGain, 2)} rank, ` +
```
phyzical, line 1479: should we be keeping the same amount of decimals? as the other two being displayed?

phyzical (approved):

src/Bladeburner/Bladeburner.tsx
ApamNapat: For exp the hundredths are not really relevant but as for rank we could get something smaller than `0.1` so displaying one more digit makes sense

PR #3250, opened 2022-03-25
Add a warning triggered while auto-saves are off

MartinFournier:
## Add a warning triggered while auto-saves are off

The check auto-save engine loop will show a warning toast if the auto-saves are disabled (at 0s) and not suppressed. This warning includes a button to re-enable them.

Adds a user setting to suppress those warnings.
Adds information to save button tooltip.

### Testing

Tested by disabling and re-enabling the setting.
Tested by using the dev menu to throw to the recovery screen & making sure there's no toast.
Tested by going to the import save comparison & making sure there's no toast.

---

### Warning Toast with enable
firefox_02nqCiHKLo

### Save button tooltip
firefox_HY3L5By8Xy

### Warning Toast message
firefox_OJF9grnLKP

### Suppress option
firefox_yj6Xwh0LAE


phyzical (approved):
LGTM

MartinFournier:
I had forgotten to check the suppress user setting:
```ts
  // If the player has suppressed those warnings let's just exit right away.
  if (Settings.SuppressAutosaveDisabledWarnings) return;
```

PR #3253, opened 2022-03-25
purchaseTor() should returns true if player already has Tor

DavidGrinberg:
Change `purchaseTor` to return `true` if the player already has purchased Tor. Previously it would return `false`

Changing this to true puts the behavior inline with the already existing behavior of `purchaseProgram`, which returns true if you have already purchased the program. Additionally this lets us call `purcahseTor` with the same logic that we use to call `purchaseProgram`

### DIFF
--- a/src/NetscriptFunctions/Singularity.ts
+++ b/src/NetscriptFunctions/Singularity.ts
@@ -499,7 +499,7 @@ export function NetscriptSingularity(
 
       if (player.hasTorRouter()) {
         workerScript.log("purchaseTor", () => "You already have a TOR router!");
-        return false;
+        return true;
       }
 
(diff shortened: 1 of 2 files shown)


DavidGrinberg:
This is technically a breaking change since it changes behavior, but I think its for the better.

phyzical:
https://github.com/danielyxie/bitburner/blob/dev/src/ScriptEditor/NetscriptDefinitions.d.ts#L1600

you will need to update the netscript definition accordingly

DavidGrinberg:
> https://github.com/danielyxie/bitburner/blob/dev/src/ScriptEditor/NetscriptDefinitions.d.ts#L1600
> 
> you will need to update the netscript definition accordingly

Updated

phyzical (approved):

PR #3283, opened 2022-03-30
Fix logic of gang territory reaching limits

TheRealMaxion:
Fix the mixup of this and other that seems to have occurred sometime in the past of this code.

### DIFF
--- a/src/Gang/Gang.ts
+++ b/src/Gang/Gang.ts
@@ -223,7 +223,7 @@ export class Gang implements IGang {
 
         if (Math.random() < thisChance) {
           if (AllGangs[otherGang].territory <= 0) return;
-          const territoryGain = calculateTerritoryGain(thisGang, otherGang);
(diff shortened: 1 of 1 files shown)


TheRealMaxion:
It seems none of the failed tests have anything to do with this, and may be related to some bugs in those systems or tests?
Namely, ram calculation, and number formatting.

phyzical:
correct failing tests are in master

phyzical:
there is another issue hydro wants this to move towards ints instead of 0 to 1

TheRealMaxion:
Ah right. This is just to fix the issue that has been in the code for a long time, where one of the lines even got removed recently.

phyzical:
only problem is i think it only makes more territory lost in the meantime is all, dont think it would help right?

TheRealMaxion:
Fair point. I've fixed it to ensure no territory disappears now.

hydroflame:
Fixed some other way.

PR #3302, opened 2022-03-30
Support setFocus() for augmentation crafting

jjclark1982:
# Bug fix

The `setFocus()` singularity function should support augmentation grafting, as there is a "do something else" button in the UI.


phyzical:
@jjclark1982 mind adding a gif of this in action? generally there is more to it, the api will need the same logic then the message in the ui ( see [my pr on adding focus for crime](https://github.com/danielyxie/bitburner/pull/2844))

hydroflame:
Merged separately because it's harder to fix the MC than just copy.

Issue #3281, opened 2022-03-30
ns.goToLocation RAM cost is 2GB (actual) vs 5GB (documented)


alainbryden:
Is listed as 5
![image](https://user-images.githubusercontent.com/2285037/160733495-2c75c065-1e26-402f-8dd6-d72a4cd70ba4.png)

Is 2
![image](https://user-images.githubusercontent.com/2285037/160733579-736296c4-f2af-4293-b1cd-f5d4ad4514f6.png)

PR #3307, opened 2022-03-31
Fix 3281 - ns.goToLocation is expecting the wrong minimum RAM

steven-r:
Fix `ns.goToLocation`'s RAM expectations (#3281).

Approach: 
* Updated the expected value in `RamCostGenerator.ts`
* Added a test to check correct RAM expectation


phyzical (approved):
lgtm the d.ts suggest 5gb also

steven-r:
closes #3281

PR #3308, opened 2022-03-31
Fix tests

phyzical:
* moved updateRam to first call to avoid lack of mocks for tests
* added wrapper funcs to simplify updateRam
* added static parser tests and skiped what isnt worker still for next person to try and fix


Ornedan:
Dynamic RAM changes look OK.

Not sure about the number format test change, however. Is it known that the number formats were intentionally changed to always produce decimal parts, even if those would be trailing zeros? And the author of that change just forgot to update tests to match?
Or were the always-decimals not an intended result? In which case those tests are alerting to a real problem and should not be silenced.

phyzical:
@Ornedan im assuming they should contain .000 as there was another issue where it would cause number raves, going from no decimals to decimals. so i assume thats what broke that test

hydroflame:
PR made before mega fmt

PR #3331, opened 2022-04-02
New coding contract: Shortest Path in a Grid

Ornedan:
Add a new pathfinding type coding contract: "Shortest Path in a Grid". Uses similar obstacle'd grid terrain as "Unique Paths in a Grid II", but permits movement in all cardinal directions.

localhost_8000__noscripts

### DIFF
--- a/src/data/codingcontracttypes.ts
+++ b/src/data/codingcontracttypes.ts
@@ -864,12 +864,12 @@ export const codingContractTypesMetadata: ICodingContractTypeMetadata[] = [
         //prev[y] = new Array(width).fill(undefined) as [undefined];
       }
 
-      function validPosition(y: number, x: number) {
+      function validPosition(y: number, x: number): boolean {
         return y >= 0 && y < height && x >= 0 && x < width && data[y][x] == 0;
       }
 
       // List in-bounds and passable neighbors
-      function* neighbors(y: number, x: number) {
+      function* neighbors(y: number, x: number): Generator<[number, number]> {
         if(validPosition(y - 1, x)) yield [y - 1, x]; // Up
         if(validPosition(y + 1, x)) yield [y + 1, x]; // Down
         if(validPosition(y, x - 1)) yield [y, x - 1]; // Left
--- a/src/data/codingcontracttypes.ts
+++ b/src/data/codingcontracttypes.ts
@@ -1,4 +1,5 @@
 import { getRandomInt } from "../utils/helpers/getRandomInt";
+import { MinHeap } from "../utils/Heap";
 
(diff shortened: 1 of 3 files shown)


src/data/codingcontracttypes.ts
```diff
@@ -794,6 +795,135 @@ export const codingContractTypesMetadata: ICodingContractTypeMetadata[] = [
...
+      const dstX = width - 1;
+      const minPathLength = dstY + dstX; // Math.abs(dstY - srcY) + Math.abs(dstX - srcX)
+
+      const grid: number[][] = new Array(height);
+      for(let y = 0; y < height; y++)
+        grid[y] = new Array(width).fill(0);
+
+      for(let y = 0; y < height; y++) {
+        for(let x = 0; x < width; x++) {
+          if(y == 0 && x == 0) continue; // Don't block start
```
phyzical, line 838: we could consider allowing the start and end being blockable and just expect "" also, more edges for the dev to handle and come across at a random time

phyzical (approved):
looks resonable to me, though i didnt deep dive into the heap. but the solvers looks legit, check length and check answer doesn't hit an obstacle and make sure it ends at the end 👍

src/data/codingcontracttypes.ts
Ornedan: The problem starts with the "character" standing at the (0,0) grid location, which requires that location to not be blocked.

Blocking the end could be OK, but if it is blockable then the obstacle generation weights would also need to be adjusted. Just removing the "Don't block destination" special case drops solvable puzzles ratio to ~66%.

phyzical: no worries :)

Hedrauta:
🤔 what if a pattern allows multiple answers?
Maybe something like:
```
[
[0,1,0,0,0,0],
[0,0,0,0,1,0]]
```` 
Then, DRRRURRD or DRRURRRD  would be correct 🤔 
will the solver fetch all possible answers and chose the shortest ones for checking the user's answer, or does it only allow one?

phyzical:
when i read the source i think we are enforcing is it the same length as the first found answer, is what they provided a valid answer and does it end in the last corner.

so if there is multiple shortest paths it will allow any

Ornedan:
Yes. Solution checker process:
- find **a** shortest path
- require that the player-given answer is not longer than the found path
- require that the player-given path is a valid path (stays in bounds, does not attempt to move into obstacles)
- require that following the player-given path ends up at the target coordinates 

If there are multiple shortest paths, any one will be accepted because they all fill all requirements.

PR #3342, opened 2022-04-03
two new CodingContracts ( HammingCode encode/decode )

Hedrauta:
1. Encoding a decimal value into a binary string using the "HammingCode"-rule
    Ranked with difficulty 5, because you can "encode" the binary value with your bare hand, but its length is dynamic

2. Check ( and fix ) a ( possible 1-Bit-altered ) encoded HammingCode-String and decode + parse it back into a decimal integer
    Ranked difficulty 8, because the probability of an altered Bit with dynamic length encoded string

Moved solvers for this into a separate file, because I need the encoders' solver-function for the decode-generator 😜 

About 10 of each coding contracts are tested locally in dev-build, no issues detected at all. 
These contracts are working fine.

Took me 2 weeks of research (RL sucks 😑 ), 4 days creating a "fast" math for the sum of parity Bits  needed, plus another 6 days to write an encoder/decoder. 

It will be a pretty unique way to look into other functions of JavaScript, that may never be used.

PS: I commented almost every line in the encoder/decoder, for better understanding of the functions 😉


Hedrauta:
Reopening the PR with a single commit for easier commit-search

PR #3350, opened 2022-04-04
Fix for install augmentation stale objects bug

### LINKED ISSUE #3348: Bug: calling installAugs() while focused on a task gives an error screen
**Steps to reproduce:**

1. Start some work (e.g. hacking contract) and don't focus on it
2. Run this script:
```
  await ns.sleep(10_000)
  ns.installAugmentations()
}
```
3. Focus on the work
4. Wait ~20 seconds for augs to be installed and game to reset

**Expected result:** You should be taken to a fresh terminal screen

**Actual result:** You get to a page like this: 
image

v1.6.3 (6c8a46d3)

phyzical:
* move route to terminal to the inside of the common fucntion
* tested without bandaid still works correctly
* fix default script with installAugmentations
closes #3348 

just routes the user to terminal like the ui does

Apr-04-2022 18-26-27


hydroflame:
PR made before mega fmt

PR #3351, opened 2022-04-04
changes static arrays to functions to force new creation when called

phyzical:
fixes the cave message popup
fixes the issues related to reqs not reseting until refresh.

tldr i derped and should have made them functions all along


hydroflame:
To be fair they should be static data but they're not.

Issue #3355, opened 2022-04-04
[Feature Request] Add id of the active script to LogBoxManager html or change electron url resolution for `window.open`


josh-hemphill:
When trying to add some simple UI elements tied to an active script, either reusing the scripts log box or creating a popup are convenient for making sure the custom UI gets removed when it's no longer needed.
Currently, in order to get a reference to a scripts own log box, there's quite a lot of finagling to get it using `querySelector`, guessing if there's padding spaces, and traversing up and back down. It seems `id`s have been added to the other important elements, so this would be a nice addition.
Alternatively, whatever settings there are for electron's resolution of `window.open` currently assumes all URLs are for a local app or file, so there's no way to create a popup window and set its contents.

PR #3357, opened 2022-04-04
Add an id to script LogWindow instance

josh-hemphill:
mostly solves #3355

Added an id to both the top element of the LogWindow instance and the parent element of the actual log entries, since that's most like to what's targeted.

I also slightly changed how the IDs and titles are concatenated so that they're a little more consistent. e.g. Titles with no args would still have a space at the end, and IDs didn't have a hyphen after the script name. Happy to revert that bit if there's a reason they're like that.


hydroflame:
The UI is not officially supported. This is guaranteed to end with me changing the popups and someone complaining I broke their scripts.

PR #3372, opened 2022-04-07
[Feature] Add special Augmentation for Grafting

nickofolas:
as discussed in Discord, adds a special Augmentation which can only be obtained via Grafting

image

### Testing
- [x] i tested Grafting the Augmentation, and noted that Entropy is successfully reset and disabled past that point

### DIFF
--- a/src/PersonObjects/Grafting/GraftableAugmentation.ts
+++ b/src/PersonObjects/Grafting/GraftableAugmentation.ts
@@ -1,7 +1,6 @@
 import { sum } from "lodash";
 
 import { Augmentation } from "../../Augmentation/Augmentation";
-import { AugmentationNames } from "../../Augmentation/data/AugmentationNames";
 import { CONSTANTS } from "../../Constants";
 
 export interface IConstructorParams {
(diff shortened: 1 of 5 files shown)


phyzical:
should the nullifying of the entropy effect be as simple as 15 mins? (i cant remember how a user gets it tbh ) just sounds op :D

phyzical (approved):

nickofolas:
> should the nullifying of the entropy effect be as simple as 15 mins? (i cant remember how a user gets it tbh ) just sounds op :D 

i don't think the 15m craft time is too big a deal, after all, it is twice the price of qlink :p

src/PersonObjects/Grafting/GraftableAugmentation.ts
```diff
@@ -18,6 +19,11 @@ export class GraftableAugmentation {
   }
 
   get cost(): number {
+    // Sell the Grafting-exclusive Aug at its starting price only
+    if (this.augmentation.name === AugmentationNames.CongruityImplant) {
+      return this.augmentation.startingCost;
```
hydroflame, line 24: I'd rather you set the cost internally to 50t and let the formula handle the rest.

PR #3380, opened 2022-04-08
Charisma is now spelled consistently with other mults

ApamNapat:


phyzical:
it might be worth taking this a step further and defining constants for the stat names and updating all usages

phyzical (approved):

PR #3381, opened 2022-04-09
Adds SnackbarVariant enum

HeinousTugboat:
This PR converts `ns.toast` and friends to use a `SnackbarVariant` enum. 

Ran test script:

```js
/** @param {NS} ns */
	ns.toast('message', 'success');
	ns.toast('message', 'warning');
	ns.toast('message', 'error');
	ns.toast('message', 'info');
	ns.toast('message', 'foo');
}
```

Result:

image

### DIFF
--- a/src/ScriptEditor/NetscriptDefinitions.d.ts
+++ b/src/ScriptEditor/NetscriptDefinitions.d.ts
@@ -6183,7 +6183,7 @@ export interface NS extends Singularity {
    * @param variant - Type of toast, must be one of success, info, warning, error. Defaults to success.
    * @param duration - Duration of toast in ms. Can also be `null` to create a persistent toast. Defaults to 2000
    */
-  toast(msg: any, variant?: string, duration?: number | null): void;
+  toast(msg: any, variant?: "success" | "info" | "warning" | "error", duration?: number | null): void;
 
   /**
    * Download a file from the internet.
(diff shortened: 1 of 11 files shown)


src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -6183,7 +6183,7 @@ export interface NS extends Singularity {
    * @param variant - Type of toast, must be one of success, info, warning, error. Defaults to success.
    * @param duration - Duration of toast in ms. Can also be `null` to create a persistent toast. Defaults to 2000
    */
-  toast(msg: any, variant?: string, duration?: number | null): void;
+  toast(msg: any, variant?: "success" | "info" | "warning" | "error", duration?: number | null): void;
```
phyzical, line 6186: wouldn't this just be type SnackBarVariant? or is it just to make it a bit easier to use from the players perspective?

phyzical (approved):

src/ScriptEditor/NetscriptDefinitions.d.ts
HeinousTugboat: That's exactly my thinking, yeah. Seeing `SnackBarVariant` as the arg type isn't useful if you don't actually have the values of it.

hydroflame:
Ok so yes, but the goal behind this was to also feed the enum to the ns object so players don't have to hardcode magic string.

PR #3385, opened 2022-04-09
New coding contract type: Array Jumping Game II

anatoly-kussul:
Screen Shot 2022-04-10 at 2 41 00 AM


Hedrauta (approved):
Awesome Contract.... it took me a while for the preview generated one to realize, that I need at least 2 Jump 😄  can't wait for trying to write my own solver on that one 😉

Issue #3388, opened 2022-04-10
Corporation division tab length element can become too wide, stretches corporation interface forcing side scroll


ChrissiQ:
The corporations tab list appears to be limited to 65% of the screen and that surely is the intention based on the code specifying a `maxWidth` of 65% for this container.  However, a long tab list length still stretches the container, thereby stretching the interface beyond the window width, because the tabs element isn't bounded as expected.  This results in the entire interface being stretched.

Video demonstrates the problem occurring.

https://user-images.githubusercontent.com/1127719/162637427-aa7317ef-1a1d-4b32-bd10-036e5a3b380f.mp4

The appearance of the tabs for corporation divisions is limited to 65% of the screen via `max-width`, but the container still treats it as though it were not limited.  I think the chromium rendering engine is getting confused trying to calculate the container width.

Most people probably don't normally create enough corporation divisions to trigger this issue.  However, the number of divisions needed to trigger this issue decreases as 1) zoom is increased and 2) division name length increases  My partner who plays at a high zoom level is encountering this problem, so this problem may disproportionately affect users who require higher zoom levels in order to play, so I think it's also an accessibility issue.

I have a simple fix for this which I'll submit as a PR shortly.

PR #3389, opened 2022-04-10
♿️  (Corp UI Accessibility) Limit tabs width to avoid stretching screen.

ChrissiQ:
Fixes https://github.com/danielyxie/bitburner/issues/3388

The rendering engine doesn't seem to respond properly to a limit of 65% for the tabs
area. Though the tab area visually appears to be limited, the parent container is
still using the full extended width of the tabs, causing page stretch. Using `65vw`
instead of `65%` for `maxWidth` causes the calculation to be based on the actual page
width so no stretching occurs.

Video demonstrating fix:

https://user-images.githubusercontent.com/1127719/162638062-0fc2f9b5-41ad-47be-a947-5843177f1c3b.mp4

### DIFF
--- a/src/Corporation/ui/CityTabs.tsx
+++ b/src/Corporation/ui/CityTabs.tsx
@@ -30,7 +30,7 @@ export function CityTabs(props: IProps): React.ReactElement {
   }
   return (
     <>
-      <Tabs variant="fullWidth" value={city} onChange={handleChange} sx={{ maxWidth: "65%" }}>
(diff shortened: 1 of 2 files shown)


phyzical:
how does this start looking on a large screen incase the 65vw does something weird in that direction?

phyzical (approved):

ChrissiQ:
Good question, I didn't check, just on a regular screen - but assumed it would be the same as before since it's only max-width, not width, and it was previously 65% anyway.  So I don't think 65vw would be any different on larger screens.

PR #3404, opened 2022-04-12
Fixed hashCost() example code

### LINKED ISSUE #3057: Documentation: code in hashCost() incorrect
Minor issue, but the code examples in hacknet.hashCost() reference "upgName", but they should be "upgradeName" (the var / const the example creates at the beginning of the code example)

Chris380:
fixes #3057 

Example code fixed


Chris380:
My bad... won't change `bitburner.d.ts` in the future

PR #3424, opened 2022-04-13
Reduced icarus.msg spam and added new truthgazer.msg

Undeemiss:
Also fixes a weird outstanding bug where message contents were saved to the player's save file for no good reason. 
TS WILL THROW AN ERROR WITH THIS CODE, but by my testing it still runs just fine. I'm only posting a PR with the error so that if it's super easy to fix you can pull it before I'm available if you want.

Tested by using the newly added Check Messages button on the dev menu with and without TRP and at different hacking levels. It was very handy.

### DIFF
--- a/src/Message/MessageHelpers.ts
+++ b/src/Message/MessageHelpers.ts
@@ -5,23 +5,23 @@ import { Programs } from "../Programs/Programs";
 import { Player } from "../Player";
 import { Page } from "../ui/Router";
 import { GetServer } from "../Server/AllServers";
+import { SpecialServers } from "../Server/data/SpecialServers";
 import { Settings } from "../Settings/Settings";
 import { dialogBoxCreate } from "../ui/React/DialogBox";
-import { Reviver } from "../utils/JSONReviver";
 import { FactionNames } from "../Faction/data/FactionNames";
+import { Server } from "../Server/Server";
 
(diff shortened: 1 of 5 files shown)


src/Message/MessageHelpers.ts
```diff
@@ -34,21 +34,27 @@ function showMessage(name: string): void {
...
-    }
+  const server = GetServer("home");
+  if (server == null) {
+    throw new Error("The home server doesn't exist. You done goofed.");
   }
   server.messages.push(msg.filename);
 }
 
+//Returns whether the given message has already been received
+function recvd(msg: Message): boolean {
```
phyzical, line 50: try to use full words just makes it more readable :)

phyzical (changes requested):
looks to be failing to build,

also should we not be resetting the list of messages each time we go into a new bitnode?

from what i see what will happen is itll reset on game load, so say you got the first 5 messages then closed the game you will get served them again.

or inversely if you don't close the game and beat 3 bitnodes they will only be served once

src/Message/MessageHelpers.ts
Undeemiss: Calling it recvd was actually naming parity with the removed msg.recvd flag. I suppose there was no particular reason to keep the name.

PR #3436, opened 2022-04-13
Create program action no longer creates duplicates

### LINKED ISSUE #2798: [Bug?] By buying a item on darkweb you can get duplicate programs
When i let making relaySMTP.exe i saw that i had produced enought money to buy it, so i went and did it, but i saw that the creation process was not canceled and did it myself. 

By using de command ls to look for a file name i saw that there was a duplicate of the program.

- [ ] Steps to reproduce
> - Start program creation process
> - Press "Do something else simultaneously"
> - Go to terminal
> - Connect darkweb
> - Buy program
> - Stop creation process or let it finish
> - Once it is done "ls" on home to see files

- [ ] Version of the game, e.g. Bitburner v1.4.0 (49de4a28)
- [ ] Captures
bug

Undeemiss:
Prevents the create programs action from creating a second copy of a program if the program is already found at home. Still gives int exp if applicable.

Fixes #2798

### DIFF
--- a/src/PersonObjects/Player/PlayerObjectGeneralMethods.tsx
+++ b/src/PersonObjects/Player/PlayerObjectGeneralMethods.tsx
@@ -1315,18 +1315,19 @@ export function createProgramWork(this: IPlayer, numCycles: number): boolean {
 
 export function finishCreateProgramWork(this: IPlayer, cancelled: boolean): string {
   const programName = this.createProgramName;
-  if (cancelled === false) {
+  if (!cancelled) {
+    //Complete case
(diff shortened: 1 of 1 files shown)


Undeemiss:
Forgot to mention how it was tested. I checked that creating programs works normally and that buying a program from the darkweb while it is partway through being created does not create duplicates whether the creation action finishes or not.

hydroflame:
I think technically I fixed this yesterday by fixing something tangentially related. BaseServer have a pushProgram method that handles this scenario.

Issue #3417, opened 2022-04-13
Make stanek instruction demo board not stretch


hydroflame:
![Screen Shot 2022-04-12 at 8 22 46 PM](https://user-images.githubusercontent.com/79246165/163075397-d6bee43c-295c-499c-86af-056bc600de8e.png)

PR #3440, opened 2022-04-13
Hide broken Stanek info graphics

Undeemiss:
Because the ridiculously oversize Stanek info graphics aren't
particularly informative relative to the confusion they're likely
to cause, I believe it pertinent to have commented them out 
temporarily so as not to confuse players.

Intended to reduce the impact of #3417, but does not actually
solve the issue.


Undeemiss:
@nickofolas has written an actual fix for the issue right as I PRed, so I'm closing this to prevent it from accidentally being pulled.

PR #3445, opened 2022-04-14
Documentation updates

DavidGrinberg:
* Add `string` as an additional valid param type for first arg of `codingContract.attempt()` because the answer to the the new contracts (e.g. hamming encoded binary to int) is a string
* Update docs for `ns.exec()` to properly reflect that 0 threads is not valid and will throw an error
* Update port read/write to use `unknown` instead of `any` or `string | number` (you can technically stuff in arrays & objects, so I think unknown is best)
* Some JSDoc lint error fixes
* adding `.historty` to gitignore, useful for the Local History VSCode plugin

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 .DS_Store
+.history
 .vscode
 Changelog.txt
 Netburner.txt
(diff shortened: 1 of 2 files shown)


hydroflame:
The `unknown` part is definitely going to confused non-typescript users. So I copied the other changes just not the unknown part.

Issue #3444, opened 2022-04-14
ns.workForFacton() message could be better if working for a gang


RobLeachman:
This whole business about "workForFaction: Faction 'NiteSec' does not offer work at the moment." could be better. Apparently it's an FAQ from people writing their first Singularity stuff. 

While "This one's not a bug - you can't work for Gang factions while you're in a gang. NiteSec is a hack gang faction" is technically accurate, could you throw us a bone please and fix the error to be "You can't workForFaction for a gang faction while you're in a gang" or something? 

Anything to help us out would be great, thanks!

PR #3454, opened 2022-04-14
Fix inconsistancy with trying to work for gang factions while running a gang

TheMas3212:
quick fix to make workForFaction only fail if the faction is the faction you actually made a gang with
this is to match the faction ui which only fails for your gang

Tested via
```js
/** @param {NS} ns **/
    const faction = 'Slum Snakes';
    ns.singularity.workForFaction(faction, 'security');

    // const faction = 'The Dark Army';
    // ns.singularity.workForFaction(faction, 'hacking');
}
```


TheMas3212:
also closes #3444

PR #3459, opened 2022-04-14
SLEEVE: Sleeves can now take on bladeburner actions

rderfler:
Added

- Bladeburner actions to sleeves
-- Field Analysis, Recruitment, Diplomacy, Infiltrate synthoids, Support main sleeve, Take on Contracts

- Moved functionality from player into person when needed for sleeves
- Replaced ISleeveOrPlayer with IPerson
- Updated Bladeburner to be sleeve friendly
- Give sleeves a brain


src/Bladeburner/IBladeburner.ts
```diff
@@ -70,13 +72,8 @@ export interface IBladeburner {
...
   getSkillNamesNetscriptFn(): string[];
   startActionNetscriptFn(player: IPlayer, type: string, name: string, workerScript: WorkerScript): boolean;
-  getActionTimeNetscriptFn(player: IPlayer, type: string, name: string, workerScript: WorkerScript): number;
-  getActionEstimatedSuccessChanceNetscriptFn(
-    player: IPlayer,
-    type: string,
-    name: string,
-    workerScript: WorkerScript,
-  ): [number, number];
+  getActionTimeNetscriptFn(person: IPerson, type: string, name: string): number | string;
```
ApamNapat, line 75: Why did you change the return type? You specifically check that the returned value is not a `string` from what I see. Same for `getActionEstimatedSuccessChanceNetscriptFn`

src/Bladeburner/Bladeburner.tsx
```diff
@@ -1116,6 +1133,15 @@ export class Bladeburner implements IBladeburner {
       }
       const losses = getRandomInt(0, max);
       this.teamSize -= losses;
+      if (this.teamSize < this.sleeveSize) {
+        const sup = player.sleeves.filter((x) => x.bbAction == "Support main sleeve");
+        for (let i = 0; i > this.teamSize - this.sleeveSize; i--) {
```
ApamNapat, line 1138: Is this logic sound, isn't `teamSize` >= `sleeveSize`?

src/Bladeburner/IBladeburner.ts
rderfler: Changing the return type allows for the removal of the dependency on `WorkerScript` which was used for logging in the event of errors. This way the function is more functional allowing the caller to handle any side effects. This allows for sleeves and player to handle the errors in what is most appropriate for their respective situation.

alainbryden:
I didn't see whether any code accounted for the fact that Raid cannot be performed if the current city has 0 communities, even if the action `count` is greater than zero.

I believe in the UI the button is automatically disabled and the player is automatically set to idle when this happens. Perhaps the sleeve behaviour has to be the same (sleeve is set to idle when a contract count hits zero or becomes invalid) - at which point we probably have to add more logic to `possibleContracts` so that they aren't selectable in the first place.

Just spit-balling here, perhaps this is inherently accounted for somewhere and I just didn't see it.

One thing we should make sure **doesn't** happen is that the probably just gets set to 0 when there are no communities and the sleeve eats up all the player's "raid" contracts.

alainbryden:
While I like the name of the task "Infiltrate Synthroid Communities", it's non-intuitive that this would generate contracts, and it almost deprecates Incite Violence as a mechanic for this to happen without the usual chaos penalty.

I suggest simply replacing this with a sleeve "Incite Violence" that is identical to that of the player. It shouldn't be that problematic since we can now potentially greatly speed up Chaos reduction by having multiple sleeves do diplomacy at the same time (although unfortunately they cannot do so in different cities at the same time)

rderfler:
> While I like the name of the task "Infiltrate Synthroid Communities", it's non-intuitive that this would generate contracts, and it almost deprecates Incite Violence as a mechanic for this to happen without the usual chaos penalty.
> 
> I suggest simply replacing this with a sleeve "Incite Violence" that is identical to that of the player. It shouldn't be that problematic since we can now potentially greatly speed up Chaos reduction by having multiple sleeves do diplomacy at the same time (although unfortunately they cannot do so in different cities at the same time)

"Infiltrate Synthoid Communities" was part of the original design given by Hydro so I will leave keeping it or not up to him. I really like the idea, and dont find adding a more powerful option an issue. Especially since with all of your sleeves working on that the player will still eventually work through everything an need to incite violence

What could be added to make it more clear what that action does?

alainbryden:
> What could be added to make it more clear what that action does?

I can't think if any way if the oy place it appears is in a sleeve dropdown, especially since the sleeve ui has no place for providing context / documentation about each action. Sleeve actions historically only consist of exact duplicates of things the player can already do.

Perhaps a way to reconcile this is to add a player general action to the UI that ss also labeled "Infiltrate Synthoid Communities", but is permanently disabled? It can carry with it some lore that explains what it does like the other actions, but then a footnote like "but you are too infamous in the synthoid community to pull this off yourself, perhaps you can use your sleeves...".

We'd want some react magic of course to ensure it only appears once sleeves are unlocked.

alainbryden (changes requested):
Blocked right now as I'm getting errors while playtesting it

Using singularity to commit a crime:
image

Using the UI (nothing happens, stack trace in console):
image

alainbryden (changes requested):
Looks like we still need to add netscript API functions for the new sleeve actions, e.g. `ns.sleeve.setToBladeburnerAction` and corresponding interface, and docs as the other functions have.

alainbryden:
Sleeves appear to be progressing much faster than they should, did you leave in some debug code to speed it up?

alainbryden:
I just noted that sleeves taking on contracts reduces the player stamina when they complete. My guess is that wasn't intended. It effectively means the player can't do anything because their stamina is kept pegged at 0 by working sleeves.

alainbryden:
Need a touch of number formatting on this one
image

alainbryden:
Not sure what this is about:
image

It may or may not be related to me having a sleeve on "Support main sleeve" - seee the error in the console

alainbryden:
This is looking really good so far in my latest test.

One quick fix - `ns.sleeve.setToBladeburnerAction(` will currently return `true` and default to "Field Analysis" if it doesn't recognize the activity the user specified (e.g. due to a typo). The expected behaviour is to return `false` and not change the current activity.

PR #3460, opened 2022-04-14
STANEK: FIX #3277 Can no longer overlap rotated fragments

### LINKED ISSUE #3277: Stanek fragment placement is broken
image

Undeemiss:
Relaxed some overzealous short-circuiting to ensure fragments
on the Stanek's Gift board properly check for collisions.

Resolves #3277

The following recording was made while clicking on every invalid location: (There is no visual indication when I click in an invalid position, unfortunately)
Recording 2022-04-14 at 11 11 16


alainbryden:
Screentogif is a great tool (the one I used to make the gift in the bug report) and it shows mouse clicks and key presses, for future reference!

PR #3464, opened 2022-04-14
MISC: vi/nvim programs added (behave the same as vim)

Hoekstraa:


alainbryden:
Hmm, if these are just aliases for vi, I'd rather players that want this just use the "alias" command in their own game saves.

It makes it harder for new players to discover basic terminal functionality if the help listing is polluted with aliases.

hydroflame:
Agreed with alain

PR #3474, opened 2022-04-15
UI: FIX #3473 'mv' now says destination script is running instead of returning an error

### LINKED ISSUE #3473: [Bug] mv doesn't check if destination script is running.
## Save
[bitburnerSave_1650038815_BN1x0.zip](https://github.com/danielyxie/bitburner/files/8497272/bitburnerSave_1650038815_BN1x0.zip)

## Versions
Appears on all recent versions, including dev branch as of right now

## Error and reproduction
The error
```
Something went wrong...please contact game dev (probably a bug)
```
pops up when running
```
home; mv hack1.js scripts/hack1.js
```

## Thoughts
The issue gets resolved when running `killall`.
I'm looking at mv.ts right now, and I'm thinking it lacks a proper handler for when the destinationScript is running.

Hoekstraa:
# Bug fix
Fixes #3473 

Tested with same save file and procedure as mentioned in #3473

2022-04-15-194915_433x294_scrot

### DIFF
--- a/src/Terminal/commands/mv.ts
+++ b/src/Terminal/commands/mv.ts
@@ -66,6 +66,13 @@ export function mv(
 
       if (destFile != null) {
         // Already exists, will be overwritten, so we'll delete it
+
+        // Command doesnt work if script is running
+        if (server.isRunning(destPath)) {
+          terminal.error(`Cannot use 'mv' on a script that is running`);
+          return;
+        }
+
         const status = server.removeFile(destPath);
         if (!status.res) {
           terminal.error(`Something went wrong...please contact game dev (probably a bug)`);


phyzical:
looks good but would also need https://github.com/danielyxie/bitburner/blob/dev/src/NetscriptFunctions.ts

the api supported also

Hoekstraa:
> looks good but would also need https://github.com/danielyxie/bitburner/blob/dev/src/NetscriptFunctions.ts
> 
> the api supported also

The API behaves differently in that it successfully overwrites.
This seems (to me) to be in line with how `scp` behaves, the CLI gives an error when trying to do anything with running scripts, but the API function overwrites without any concern. Changing the API to block overwriting could cause script breakage too, since users might depend on the API version to overwrite indiscriminately.

For consistency sake, I would recommend to not have the API changed for this issue, since the current behaviour of `mv` is in line with `scp`. Let me know your thoughts on this @phyzical.

phyzical:
hmmmm the sounds weird AF but im not gonna question it :D i believe you

phyzical (approved):

PR #3480, opened 2022-04-16
CORPORATION: Limit shareholder priority on newly issued shares

Undeemiss:
Updated the computations for issuing new corp shares so that
private investors cannot purchase a greater percentage of the
company than they already own. This change is an indirect buff
to corporations that elect not to use as many private investors
as possible.

Also updated UI elements to reflect the change. I'm not sure about
the wording I chose; I think it could be improved upon, but that
was the best I came up with.

The change was tested by creating a corp with a high share price,
then checking that the results of issuing new shares were logical
when performed with 0-4 sets of private investors having been
used.

Fixes [this issue on discord](https://discord.com/channels/415207508303544321/415213413745164318/964904751546368010) that did not get a Github issue

### DIFF
--- a/src/Corporation/ui/modals/IssueNewSharesModal.tsx
+++ b/src/Corporation/ui/modals/IssueNewSharesModal.tsx
@@ -73,24 +73,31 @@ export function IssueNewSharesModal(props: IProps): React.ReactElement {
 
     const profit = newShares * newSharePrice;
     corp.issueNewSharesCooldown = CorporationConstants.IssueNewSharesCooldown;
-    corp.totalShares += newShares;
 
     // Determine how many are bought by private investors
-    // Private investors get up to 50% at most
-    // Round # of private shares to the nearest millionth
-    let privateShares = getRandomInt(0, Math.round(newShares / 2));
-    privateShares = Math.round(privateShares / 1e6) * 1e6;
+    // If private investors own n% of the company, private investors get up to 0.5n% at most
+    // Round # of private shares to the nearest million
+    const privateOwnedRatio = 1 - (corp.numShares + corp.issuedShares) / corp.totalShares;
+    const maxPrivateShares = Math.round((newShares / 2) * privateOwnedRatio);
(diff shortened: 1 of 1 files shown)


src/Corporation/ui/modals/IssueNewSharesModal.tsx
```diff
@@ -73,24 +73,31 @@ export function IssueNewSharesModal(props: IProps): React.ReactElement {
...
-    corp.totalShares += newShares;
 
     // Determine how many are bought by private investors
-    // Private investors get up to 50% at most
-    // Round # of private shares to the nearest millionth
-    let privateShares = getRandomInt(0, Math.round(newShares / 2));
-    privateShares = Math.round(privateShares / 1e6) * 1e6;
+    // If private investors own n% of the company, private investors get up to 0.5n% at most
+    // Round # of private shares to the nearest million
+    const privateOwnedRatio = 1 - (corp.numShares + corp.issuedShares) / corp.totalShares;
```
phyzical, line 80: is this correct though? as numShares is just how many they own it doesnt really have anything to do with if they received funding from private investors right?

phyzical: or did i get confused and "private shareholders are just the shares sold to market" and theres some logic to randomly buy those somewhere else?

Undeemiss: From what I understand, numshares is the number of shares owned by the player, and issuedshares is the number of publicly owned shares. As the only way I know of to reduce both of those numbers at once is private investors, I believe it should work. I don’t have an amazing understanding of corps, but I was referencing other code that needs to determine how much of the corp the player actually owns when I wrote that.

Undeemiss: Also, to be clear, I did spend at least like 10 minutes just testing this, and in the final version it seemed to work as I expected. So regardless of if my math makes sense, the results seemed reasonable to me :smile:

PR #3483, opened 2022-04-16
API: Fix leak of real Employee object in hireEmployee

TheMas3212:
Tested via
```js
/** @param {NS} ns */
    const corpName = 'Corp';
    const divisionName = 'Farms';
    const cityName = 'Sector-12';
    ns.corporation.createCorporation(corpName, false);
    ns.corporation.expandIndustry('Agriculture', divisionName);
    const e1 = ns.corporation.hireEmployee(divisionName, cityName);
    const e2 = ns.corporation.hireEmployee(divisionName, cityName);
    const e3 = ns.corporation.hireEmployee(divisionName, cityName);
    [e1, e2, e3].forEach((e) => {
        e.sal = -Infinity;
    })
    await ns.corporation.assignJob(divisionName, cityName, e1.name, 'Operations');
    await ns.corporation.assignJob(divisionName, cityName, e2.name, 'Engineer');
    await ns.corporation.assignJob(divisionName, cityName, e3.name, 'Business');
}
```

### DIFF
--- a/src/NetscriptFunctions/Corporation.ts
+++ b/src/NetscriptFunctions/Corporation.ts
@@ -730,7 +730,22 @@ export function NetscriptCorporation(
       const divisionName = helper.string("hireEmployee", "divisionName", _divisionName);
       const cityName = helper.city("hireEmployee", "cityName", _cityName);
       const office = getOffice(divisionName, cityName);
-      return office.hireRandomEmployee();
+      const employee = office.hireRandomEmployee();
+      if (employee === undefined) return undefined;
+      return {
+        name: employee.name,
+        mor: employee.mor,
+        hap: employee.hap,
+        ene: employee.ene,
+        int: employee.int,
+        cha: employee.cha,
+        exp: employee.exp,
+        cre: employee.cre,
+        eff: employee.eff,
+        sal: employee.sal,
+        loc: employee.loc,
+        pos: employee.pos,
+      };
     },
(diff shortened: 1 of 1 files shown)


src/NetscriptFunctions/Corporation.ts
```diff
@@ -730,7 +730,22 @@ export function NetscriptCorporation(
       const divisionName = helper.string("hireEmployee", "divisionName", _divisionName);
       const cityName = helper.city("hireEmployee", "cityName", _cityName);
       const office = getOffice(divisionName, cityName);
-      return office.hireRandomEmployee();
+      const employee = office.hireRandomEmployee();
+      if (employee === undefined) return undefined;
+      return {
```
phyzical, line 735: how about we do this inside the hireRandomEmployee?

TheMas3212: the return value of hireRandomEmployee is used on line 55 of src/Corporation/OfficeSpace.ts
also it makes more sense to me to sanitize it at the edge between internal and userspace

Snarling: Should be able to just return `{...employee}` here instead of manually building a new object. This would also keep working if in the future there is a change to what characteristics an employee has.

TheMas3212: probably.
In this case im just mimicing what getEmployee does elsewhere in the file

Snarling (approved):
Looks good to me.

PR #3490, opened 2022-04-18
Singularity: Fix #3489 Disable checkTixApiAccess for purchase4SMarketData

### LINKED ISSUE #3489: [Bug] You cannot purchase 4s market data via singularity before TIX API access
https://github.com/danielyxie/bitburner/blob/3bccfb1de672cd6e9a65155051ed9c07f37981bc/src/NetscriptFunctions/StockMarket.ts#L359-L361

I'm presuming this is unintentional as you can purchase 4s market data via UI before getting TIX API access.

DavidGrinberg:
Closes #3489 
This restriction doesn't apply on UI. It doesn't make sense to apply it to singularity api.

### DIFF
--- a/src/NetscriptFunctions/StockMarket.ts
+++ b/src/NetscriptFunctions/StockMarket.ts
@@ -358,7 +358,6 @@ export function NetscriptStockMarket(player: IPlayer, workerScript: WorkerScript
     },
     purchase4SMarketData: function (): boolean {
       updateRam("purchase4SMarketData");
-      checkTixApiAccess("purchase4SMarketData");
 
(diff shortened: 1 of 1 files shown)


phyzical:
oh actually isnt this due to needing api access to use it in an api context? though it probably isnt needed

phyzical (approved):

DavidGrinberg:
> oh actually isnt this due to needing api access to use it in an api context? though it probably isnt needed

There is a separate purcahse for 4S API, and there it makes sense to do TIX API check. This one is really just for UI.

PR #3494, opened 2022-04-18
FREEZE until after development restarts

### LINKED ISSUE #3295: Set up proper CD pipeline
Per hydro, adding an issue for this: we should add a proper pipeline so we can have both a beta and prod release, allowing us to release features to a limited set of users before more broadly pushing them out to the full userbase. Additionally, we should use envvars to track the deployed sha so we can avoid the chicken-and-egg issue of not knowing the _current_ sha of the deployed code.

phyzical:
closes #3295

Okay so ill do my best to explain what this is doing

This new workflow will start 
* on dev branch pushes IF the CI workflow is successful
OR
* on a release being published  

>${{ github.event.workflow_run.conclusion == 'success' }} || ${{ github.event.workflow_run.conclusion == false }}

This is just the check to find if the CI was successful OR its the release trigger

The workflow then builds the app as production, installs wine then builds the electorn app for windows,linux and mac

Then zips the assets for smaller download size

Then 
* if the the workflow was triggered via a dev commit it will upload the builds as artifacts see example (https://github.com/phyzical/bitburner/actions/runs/2184141670)
* if its a release then it will find the release that triggered the workflow and attach the build assets to those see example (https://github.com/phyzical/bitburner/releases/tag/v1.22)

Also if its a release it will take the results of the dist folder and push them to a branch named `gh-pages` so it should just be a matter of updating the github pages to use this branch going forward

Given the above we should have auto build for both development and releases (i.e after using the existing bump-version workflow)

The outstanding things in the workflow ATM are
* auto deploy for docs NFI how this is currently handled but can update when i get some info
* auto deploy to steam, have left the 3rd part action commented out until we can get access to secrets for when that time comes

lmk if ive not explained something well enough

### DIFF
--- a/.github/workflows/bump-version.yml
+++ b/.github/workflows/bump-version.yml
@@ -12,20 +12,21 @@ on:
       changelog:
         description: "Changelog (url that points to RAW markdown)"
         default: ""
+        required: false
       buildApp:
         description: "Include Application Build"
         type: boolean
-        default: "true"
+        default: true
         required: true
       buildDoc:
         description: "Include Documentation Build"
         type: boolean
-        default: "true"
+        default: true
         required: true
       prepareRelease:
         description: "Prepare Draft Release"
         type: boolean
-        default: "true"
+        default: true
         required: true
 
 jobs:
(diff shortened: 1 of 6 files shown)


Hoekstraa (approved):
Looks good to me :+1:

PR #3513, opened 2022-04-19
STANEK: FIX #3282 Added NS function stanek.acceptGift

### LINKED ISSUE #3282: Missing Stanek API function for getting the initial gift
There is a full API for managing the gift once you have it, but I see no current way to get the initial gift through the API without some heavy DOM manipulation.
If adding such a function, make sure it has the same limitations as when doing it manually.

Undeemiss:
Added a new NS function stanek.acceptGift to the game, with a ram
cost of 2gb. The function attempts to join the CotMG and install
the gift, then returns true iff the player is in the CotMG and has
the gift installed.

Tested by checking the ram cost of the function is handled properly,
that the function properly joins the church and installs the aug when
the player has access to it, and that the function properly fails when
the player does not have access to the church. Also tested the
impossible to legitimately reach edge case of being in the church
without having the aug installed.

Resolves #3282


src/NetscriptFunctions/Stanek.ts
```diff
@@ -109,5 +113,28 @@ export function NetscriptStanek(
         checkStanekAPIAccess("removeFragment");
         return staneksGift.delete(rootX, rootY);
       },
+    acceptGift: (_ctx: NetscriptContext) =>
+      function (): boolean {
+        //Check if the player has access to the church
+        if (player.canAccessCotMG()) {
+          //Attempt to join CotMG
+          const faction = Factions[FactionNames.ChurchOfTheMachineGod];
+          if (!player.factions.includes(FactionNames.ChurchOfTheMachineGod)) {
```
hydroflame, line 122: joinFaction already has that check.

src/NetscriptFunctions/Stanek.ts
```diff
@@ -109,5 +113,28 @@ export function NetscriptStanek(
...
+          }
+          //Attempt to install the first Stanek aug
+          if (
+            !player.hasAugmentation(AugmentationNames.StaneksGift1) &&
+            !player.queuedAugmentations.some((a) => a.name === AugmentationNames.StaneksGift1)
+          ) {
+            applyAugmentation({ name: AugmentationNames.StaneksGift1, level: 1 });
+          }
+        }
+        //Return true iff the player is in CotMG and has the first Stanek aug installed
```
hydroflame, line 133: Missing the check that there's no other aug beside NFG

PR #3518, opened 2022-04-19
MISC: fix typing conflict between jest and cypress

taralx:
Cypress and Jest both define "expect", but they come from different
libraries (Cypress uses Chai, Jest uses its own thing). So we can't
include both of them in the tsconfig.json. Conveniently, however, we
don't need any of the test code to be part of the main project, so
it suffices to give cypress its own tsconfig.json file.

That being done, adding "jest" to the global types lets us remove all
those imports.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -4,13 +4,15 @@
     "esModuleInterop": true,
     "isolatedModules": true,
     "jsx": "react",
-    "lib": ["es2016", "dom", "es2017.object", "es2019"],
     "module": "commonjs",
-    "target": "es6",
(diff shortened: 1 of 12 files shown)


tsconfig.json
```diff
@@ -4,13 +4,15 @@
     "esModuleInterop": true,
     "isolatedModules": true,
     "jsx": "react",
-    "lib": ["es2016", "dom", "es2017.object", "es2019"],
```
hydroflame, line 7: Why is this gone?

taralx:
Rebased.

PR #3525, opened 2022-04-20
TERMINAL: FIX #3492 Allow cd .. even when destination directory is empty

### LINKED ISSUE #3492: [Bug] Inconsistency: deleting a folder on any other server than home has different folder deletion behaviour than home
## Description
When deleting the last file in a folder, while currently being located within the folder, the player is 'cut off' from other folders, unable to use relative pathing to return to the base folder ( ~/ ).

Above behaviour is prevalent in all servers but home, which handles the scenario as one might expect, allowing the player to use `cd ..` for example to return to a folder that is in existence.

## Version
Tested on 1.6.4 (fbb99c17)

## Reproduction
Home (showing no error)
behaviour on home
Other server (showing error)
behaviour on faction server

Snarling:
Fixes #3492 
`cd ..` would previously fail when the destination was empty (see issue). Player should always be able to travel up the directory tree, unless already at `/`.

### DIFF
--- a/src/Terminal/commands/cd.ts
+++ b/src/Terminal/commands/cd.ts
@@ -30,11 +30,11 @@ export function cd(
         terminal.error("Invalid path. Failed to change directories");
         return;
       }
-      if (terminal.cwd().length>1 && dir === ".."){
+      if (terminal.cwd().length > 1 && dir === "..") {
         terminal.setcwd(evaledDir);
         return;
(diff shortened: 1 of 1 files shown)


Snarling:
Prettier check failed due to some weird failure to download cypress, not related to this PR. Format and lint were both ran prior to PR.

PR #3537, opened 2022-04-22
CODINGCONTRACT: FIX #3484 BREAKING Fixed capitalization in contract name

### LINKED ISSUE #3484: "HammingCodes: Integer to encoded Binary" should have "encoded" capitalized to match other coding contracts
The coding contract "HammingCodes: Integer to encoded Binary" should have "encoded" capitalized to match other coding contracts:
HammingCodes: Integer to Encoded Binary

Changing this may mess with people's existing coding contract scripts, if they've already hard coded this in, but it's a new enough contract that it may not be a problem.

phyizical mentioned on discord: "the correct solution to this would be to make an enum for all the coding contract identifiers so that theres a way for users to pull this information, that way changing the front facing stuff shouldn't break their scripts, tldr less magic strings"

Undeemiss:
Renamed "HammingCodes: Integer to encoded Binary" to
"HammingCodes: Integer to Encoded Binary" for consistency.
This will break any/all solvers that have hardcoded this name.

Resolves #3484


Undeemiss:
Ah, yep, that does in fact seem necessary. I'll see if I can figure that out tonight.

Undeemiss:
> This needs a save file migration or people will crash.

Done. Tested by loading an old save file with the renamed contract and ensuring
that opening the migrated contract works as intended in the new version.

PR #3539, opened 2022-04-22
API: Fix inconsistent return value in `ns.grafting.getAugmentationGraftTime`

nickofolas:
fixes `ns.grafting.getAugmentationGraftTime` to return a value that's consistent with the value displayed in the UI and used in calculations for actual completion

in doing so, adds a couple helpers functions to provide a reusable access point for the values


phyzical (approved):
lgtm

PR #3540, opened 2022-04-22
MISC: add better typing to Electron.tsx

taralx:
There was a lot of abuse of `any` in here. This change adds actual typing for the electron bridge via ambient declarations (`declare interface`). There's still one any left in the receive function since the types of the parameters are variable.


phyzical:
can approve with a quick gif of the electron app built and running 👍

taralx:
image

phyzical (approved):

PR #3544, opened 2022-04-23
REFACTOR:  augmentation cost, rep cost and level to be calculated in place

phyzical:
* rename augmentation to static augmentation
* remove level, base cost and rep requirement from base augmentation obj
* added in place calculations for cost and level of augmentations given player context
* removed redundant logic for dynamically calculating cost,rep and level against the augmentation object
* replaced references to startingCost to baseCost as it is now always whatever it started at
* freeze the cost to stop sneeky sneekers

Apr-15-2022 11-09-22
Apr-15-2022 11-08-34


nickofolas (approved):
`src/Faction/ui/PurchaseableAugmentation.tsx` can be removed as it's unused now. Other than that, looks great!

PR #3548, opened 2022-04-23
UI: Improve tail window buttons

Snarling:
Tail window buttons have been moved properly inside of the tail window titlebar, previously the buttons clipped out, and would behave oddly if the titlebar was shrunk too small in the X direction. See video.

https://user-images.githubusercontent.com/84951833/164945581-a0100aea-9c6f-4db9-b400-249912f623b3.mp4

### DIFF
--- a/src/ui/React/LogBoxManager.tsx
+++ b/src/ui/React/LogBoxManager.tsx
@@ -96,6 +96,9 @@ const useStyles = makeStyles((theme: Theme) =>
       display: "flex",
       flexDirection: "column-reverse",
     },
+    titleButton: {
+      padding: "1px 6px",
+    },
     success: {
       color: theme.colors.success,
     },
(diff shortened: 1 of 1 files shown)


Snarling:
Added an additional commit  that places constraints on minimum size of tail resizer. See previous video for the unconstrained resize.

https://user-images.githubusercontent.com/84951833/164949901-bb26c299-60e9-46b6-98fb-11247f80b048.mp4

src/ui/React/LogBoxManager.tsx
```diff
@@ -260,35 +263,34 @@ function LogWindow(props: IProps): React.ReactElement {
...
-                  </Button>
-                )}
-                {workerScripts.has(script.pid) && (
-                  <Button onClick={kill} onTouchEnd={kill}>
-                    Kill
-                  </Button>
-                )}
-                <Button onClick={minimize} onTouchEnd={minimize}>
-                  {minimized ? "\u{1F5D6}" : "\u{1F5D5}"}
+              {!workerScripts.has(script.pid) && (
```
phyzical, line 270: nit: could || these two buttons based on the condition

phyzical (approved):
lgtm

src/ui/React/LogBoxManager.tsx
Snarling: Yeah no need to do the same check twice. Changed it to a ternary.

PR #3549, opened 2022-04-23
INFILTRATION: Fix minigame cycle

nickofolas:
fixes infiltration so that games besides wire cutting are utilized


Snarling (approved):

phyzical:
was that me or hydro big oppsie 😆

phyzical (approved):

PR #3552, opened 2022-04-24
COMPANY: Fix #3551 Applying for a new  job will not change active employer if player is performing company work

### LINKED ISSUE #3551: Applying to a new job while in the middle of working transfers ongoing reputation to new job.
From discord. https://discord.com/channels/415207508303544321/415213413745164318/967469637983940608

Snarling:
Also streamlined some logic, reducing applyForWork by ~30 lines.

If the player is currently working for company X and applies to a different company, they can now acquire or change jobs at the new employer while continuing their current shift of work at company X. If this player applies to a different position at company X, they also continue their shift and their position will just change in the middle of the shift.

Fixes #3551

### DIFF
--- a/src/PersonObjects/Player/PlayerObjectGeneralMethods.tsx
+++ b/src/PersonObjects/Player/PlayerObjectGeneralMethods.tsx
@@ -569,7 +569,7 @@ export function processWorkEarnings(this: IPlayer, numCycles = 1): void {
 
 /* Working for Company */
 export function startWork(this: IPlayer, companyName: string): void {
-  this.resetWorkStatus(CONSTANTS.WorkTypeCompany, companyName);
+  if (this.isWorking) this.resetWorkStatus(CONSTANTS.WorkTypeCompany, companyName);
   this.isWorking = true;
   this.companyName = companyName;
   this.workType = CONSTANTS.WorkTypeCompany;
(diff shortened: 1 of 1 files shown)


src/PersonObjects/Player/PlayerObjectGeneralMethods.tsx
```diff
@@ -1764,66 +1756,44 @@ export function applyForJob(this: IPlayer, entryPosType: CompanyPosition, sing =
...
-        return false;
+        const reqText = getJobRequirementText(company, nextPos);
+        dialogBoxCreate("Unfortunately, you do not qualify for a promotion<br>" + reqText);
       }
     }
+    return false;
   }
 
   this.jobs[company.name] = pos.name;
-  if (!this.focus && this.isWorking && this.companyName !== this.location) this.resetWorkStatus();
```
phyzical, line 1822: hmm but if we dont call this wont we run into issues with the workness of the user? or is that intentional as in carry over time worked into the new job? imo shouldn't we finish their work and start a new work session

Snarling: If player is currently working a shift at company X and applies for a new job at company Y, there's no need to reset work status. We can just add the newly applied for job without interrupting the current work shift.

We do have to reset the work status when the player decides to start actually _working_ at company Y, but that's handled by a separate function. This function is just for applying for a new job which has no reason to reset work status.

It used to need to do this when the player could only have one job - if the player applied for a new job we would need to cancel any in progress job at a different company because the player's employer was being replaced. But now with multiple jobs, nothing is being replaced when the player applies at a new company.

PR #3565, opened 2022-04-26
AUGMENTATIONS: Fix `isSpecial` filter in helper (Removes NeuroFlux, Stanek's Gift, etc from gangs)

nickofolas:
Fixes the operator in the filter so that augs must meet both criteria


phyzical (approved):
ah shoot opps, good spot

PR #3570, opened 2022-04-27
CORPORATIONS: Expose makeProducts on NSDivision interface

DavidGrinberg:
`corp.makeProduct()` barfs if you try to make a product on an industry that does not make products (such as Agriculture). The game does not provide systematic access to what industries do and don't make products, so users are left to manually make maps, which sucks. 

Exposing this value fixes that.

### DIFF
--- a/src/NetscriptFunctions/Corporation.ts
+++ b/src/NetscriptFunctions/Corporation.ts
@@ -295,6 +295,7 @@ export function NetscriptCorporation(
       if (office === 0) continue;
       cities.push(office.loc);
     }
+
     return {
       name: division.name,
       type: division.type,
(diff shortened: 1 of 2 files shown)


src/NetscriptFunctions/Corporation.ts
```diff
@@ -309,6 +310,7 @@ export function NetscriptCorporation(
       upgrades: division.upgrades.slice(),
       cities: cities,
       products: division.products === undefined ? [] : Object.keys(division.products),
+      makesProducts: getDivision(division.name).makesProducts,
```
phyzical, line 313: we shouldn't need to get the division again 

```suggestion
      makesProducts: division.makesProducts,
```

should suffice

DavidGrinberg: Its a different type of division. `IIndustry` vs `NSDivision`

PR #3577, opened 2022-04-28
FEATURE: added logic to allow quitJob to be called from singularity

### LINKED ISSUE #3567: missing singularity function: quitCompanyJob
there should be a function to quit a job at a company.

Use case may be that one want to make sure that no contract gets wasted by producing rep for this company.

phyzical:
closes #3567

### DIFF
--- a/src/PersonObjects/IPlayer.ts
+++ b/src/PersonObjects/IPlayer.ts
@@ -248,7 +248,7 @@ export interface IPlayer {
   queryStatFromString(str: string): number;
   getIntelligenceBonus(weight: number): number;
   getCasinoWinnings(): number;
-  quitJob(company: string): void;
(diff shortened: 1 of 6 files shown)


src/PersonObjects/IPlayer.ts
```diff
@@ -248,7 +248,7 @@ export interface IPlayer {
   queryStatFromString(str: string): number;
   getIntelligenceBonus(weight: number): number;
   getCasinoWinnings(): number;
-  quitJob(company: string): void;
+  quitJob(company: string, sing?: boolean): void;
```
DavidGrinberg, line 251: Whats the `sing` param do?

phyzical: nothing atm, its more of a "just incase" basically if the func ever needed to do logging it would be needed

PR #3591, opened 2022-05-01
Update bitburner.singularity.getdarkwebprograms.md

Rick-Bird:
fixed examples

### DIFF
--- a/markdown/bitburner.singularity.getdarkwebprograms.md
+++ b/markdown/bitburner.singularity.getdarkwebprograms.md
@@ -28,7 +28,7 @@ This function allows the player to get a list of programs available for purchase
 
 ```ts
 // NS1
-getDarkwebProgramsAvailable();
+getDarkwebPrograms();
 // returns ['BruteSSH.exe', 'FTPCrack.exe'...etc]
 ```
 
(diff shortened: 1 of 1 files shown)


phyzical (approved):

TheMas3212:
is this editing the right file?
it looks like its targeting the generated markdown, not the typedefs

Per https://github.com/danielyxie/bitburner/blob/dev/doc/CONTRIBUTING.md#as-a-documentor

phyzical:
ah good point, @Rick-Bird so yeah the mds are auto generated via the d.ts files as pointed out by the provided doc thanks @TheMas3212

TheMas3212 (changes requested):
Per https://github.com/danielyxie/bitburner/blob/dev/doc/CONTRIBUTING.md#as-a-documentor
markdown changes should be done by changing the typescript definitions used to generate them not buy editing the markdown files themself. as the markdown files are autogenerated
specfically in this case you need to edit
https://github.com/danielyxie/bitburner/blob/9c805dabb8f6be1e9158b715f83a5df63be087f1/src/ScriptEditor/NetscriptDefinitions.d.ts#L2340-L2351

Issue #3596, opened 2022-05-02
Autocomplete/argument parsing does not work correctly with quoted arguments/escaped spaces


RevanProdigalKnight:
Game version: `Bitburner v1.6.4 (00628a8c)`

This is really 2 issues, autocomplete not working properly on quoted options and terminal argument parsing not properly handling escaped spaces.

When trying to include autocomplete for single arguments which contain spaces by wrapping them in quotation marks (single or double), the autocomplete breaks and doesn't suggest anything.

```js
// autocomplete-cities.js
export function autocomplete(data, args) {
  return ['Sector-12', 'Aevum', 'Volhaven', 'Chongqing', '"New Tokyo"', 'Ishima'];
}

export async function main(ns) {
  ns.tprintf('City: %s', ns.args[0]);
}
```

```sh
# in the terminal
$ run autocomplete-cities.js S\t # Autocomplete works, fills in with Sector-12. Same for other single-word cities
$ run autocomplete-cities.js N\t # No autocomplete occurs, no popup shows
$ run autocomplete-cities.js "N\t # same as above
```

Returning arguments with spaces that are *not* wrapped in quotations does provide autocomplete options but the autocomplete does not auto-wrap them in quotes

```js
// autocomplete-cities-2.js
export function autocomplete(data, args) {
  return ['Sector-12', 'Aevum', 'Volhaven', 'Chongqing', 'New Tokyo', 'Ishima'];
}

export async function main(ns) {
  ns.tprintf('City: %s', ns.args[0]);
}
```

```sh
# in the terminal
$ run autocomplete-cities-2.js N\t # autofill provides `New Tokyo` (without quotes)
Running script with 1 thread(s), pid 1 and args: ["New", "Tokyo"]
City: New
```

Autocomplete works if spaces are escaped with `\\`, but then still ends up being 2 separate arguments when run:

```js
// autocomplete-cities-3.js
export function autocomplete(data, args) {
  return ['Sector-12', 'Aevum', 'Volhaven', 'Chongqing', 'New\\ Tokyo', 'Ishima'];
}

export async function main(ns) {
  ns.tprintf('City: %s', ns.args[0]);
}
```

```sh
# in the terminal
$ run autocomplete-cities-3.js N\t # autofill provides `New\ Tokyo` (without quotes)
Running script with 1 thread(s), pid 1 and args: ["New\\", "Tokyo"]
City: New
```

[...]

PR #3606, opened 2022-05-04
MISC: #3596 Enhanced terminal command parsing

RevanProdigalKnight:
This is a partial fix for #3596 in that it does not address the autocomplete issues identified, but it does address terminal command parsing in order to allow joining what would be split arguments before into single arguments if they were separated by an escaped space.

```sh
# before
$ run script.js single\ argument
Running script with 1 thread(s), pid 1 and args: ["single\\", "argument"].

# after
$ run script.js single\ argument
Running script with 1 thread(s), pid 1 and args: ["single argument"]
```

Other command parsing improvements included in this PR:

- Additional JS numeric literal formats are now supported by using the Number constructor instead of parseFloat, so hexadecimal, octal, and binary number representations can be used as arguments
- Nested quoting in quoted arguments is handled properly, e.g.: `"quoted arg 'with nested \"quoting\" gets parsed as a single arg'"`
  - also does not break on apostrophes inside of arguments

I attempted to create a unit test suite in order to verify functionality beyond my testing in the browser while I was refactoring the `ParseCommand` function, but compilation of the test failed while trying to resolve an import in a file I didn't touch. I've included the error I got below, if anyone has any suggestions on how to get around that I'd be more than happy to add a unit test suite.

```
$ npm run test Terminal/Parser

> bitburner@1.6.4 test
> jest "Terminal/Parser"

 FAIL  test/jest/Terminal/Parser.test.ts
  ● Test suite failed to run

    Cannot find module '!!raw-loader!../NetscriptDefinitions.d.ts' from 'src/ScriptEditor/ui/ScriptEditorRoot.tsx'

    Require stack:
      src/ScriptEditor/ui/ScriptEditorRoot.tsx
      src/ui/GameRoot.tsx
      src/NetscriptFunctions/Grafting.ts
      src/NetscriptFunctions.ts
      src/NetscriptWorker.ts
      src/Prestige.ts
      src/Augmentation/AugmentationHelpers.tsx
      src/Faction/FactionHelpers.tsx
      src/Bladeburner/Bladeburner.tsx
      src/PersonObjects/Player/PlayerObjectBladeburnerMethods.ts
      src/PersonObjects/Player/PlayerObject.ts
      src/Player.ts
      src/DarkWeb/DarkWeb.tsx
      src/Terminal/Terminal.ts
      src/Terminal.ts
      src/Alias.ts
      src/Terminal/Parser.ts
      test/jest/Terminal/Parser.test.ts

      43 | import { Modal } from "../../ui/React/Modal";
      44 |
    > 45 | import libSource from "!!raw-loader!../NetscriptDefinitions.d.ts";
         | ^
      46 | import { TextField, Tooltip } from "@mui/material";
      47 |
      48 | interface IProps {

      at Resolver.resolveModule (node_modules/jest-resolve/build/resolver.js:322:11)
      at Object.<anonymous> (src/ScriptEditor/ui/ScriptEditorRoot.tsx:45:1)

Test Suites: 1 failed, 1 total
Tests:       0 total
Snapshots:   0 total
Time:        17.149 s
Ran all test suites matching /Terminal\\Parser/i.
```

### DIFF
--- a/src/Terminal/Parser.ts
+++ b/src/Terminal/Parser.ts
@@ -64,7 +64,7 @@ export function ParseCommand(command: string): (string | number | boolean)[] {
         lastQuote = c;
         // Otherwise if we're in a string argument, add the current character to it
       } else {
-      	arg += c;
+        arg += c;
       }
       // If the current character is a space and we are not inside a string, parse the current argument
       // and start a new one
--- a/src/Terminal/Parser.ts
+++ b/src/Terminal/Parser.ts
@@ -1,17 +1,17 @@
 import { KEY } from "../utils/helpers/keyCodes";
 import { substituteAliases } from "../Alias";
 // Helper function to parse individual arguments into number/boolean/string as appropriate
-function parseArg(arg) {
+function parseArg(arg: string): string | number | boolean {
   // Handles all numbers including hexadecimal, octal, and binary representations, returning NaN on an unparseable string
   const asNumber = Number(arg);
   if (!isNaN(asNumber)) {
     return asNumber;
   }
-  
+
   if (arg === 'true' || arg === 'false') {
     return arg === 'true';
   }
-  
+
   // Strip quotation marks from strings that begin/end with the same mark
   return arg.replace(/^"(.*?)"$/g, '$1').replace(/^'(.*?)'$/g, '$1');
 }
@@ -44,14 +44,14 @@ export function ParseCommands(commands: string): string[] {
 export function ParseCommand(command: string): (string | number | boolean)[] {
   let idx = 0;
   const args = [];
-  
+
   // Track depth of quoted strings, e.g.: "the're 'going away' rather 'quickly \"and awkwardly\"'" should be parsed as a single string
-  const quotes = [];
-  
+  const quotes: string[] = [];
+
   let arg = '';
   while (idx < command.length) {
     const c = command.charAt(idx);
-    
+
     // If the current character is a backslash, add the next character verbatim to the argument
     if (c === '\\') {
       arg += command.charAt(++idx);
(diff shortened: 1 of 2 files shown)


phyzical (approved):
looks resonable

src/Terminal/Parser.ts
```diff
@@ -1,12 +1,21 @@
...
-function isNumber(str: string): boolean {
-  if (typeof str != "string") {
-    return false;
-  } // Only process strings
-  return !isNaN(str as unknown as number) && !isNaN(parseFloat(str));
+// Helper function to parse individual arguments into number/boolean/string as appropriate
+function parseArg(arg: string): string | number | boolean {
+  // Handles all numbers including hexadecimal, octal, and binary representations, returning NaN on an unparseable string
+  const asNumber = Number(arg);
+  if (!isNaN(asNumber)) {
```
taralx, line 7: Prefer isFinite unless you want to allow Infinity.

src/Terminal/Parser.ts
```diff
@@ -33,98 +42,62 @@ export function ParseCommands(commands: string): string[] {
...
-          continue;
-        }
-      } else if (inQuote === ``) {
-        inQuote = KEY.QUOTE;
-      } else if (inQuote === KEY.QUOTE) {
-        inQuote = ``;
-      }
-    } else if (c === KEY.SPACE && inQuote === ``) {
-      const arg = command.substr(start, i - start);
+  // Track depth of quoted strings, e.g.: "the're 'going away' rather 'quickly \"and awkwardly\"'" should be parsed as a single string
```
taralx, line 48: That's going to be weird. Generally speaking you don't match quotes inside quotes, otherwise `"they're"` has an unbalanced quote.

taralx (changes requested):

src/Terminal/Parser.ts
RevanProdigalKnight: I don't see any reason to explicitly forbid Infinity as a numeric argument when all other valid JavaScript numeric literals are now allowed.

src/Terminal/Parser.ts
RevanProdigalKnight: Hmm, I guess I did get a bit overly fancy there. I can cut it down to simple matching of the start quote, though the current processing defines valid start patterns of ` "` and `="` (and also with single quotes), so it currently doesn't break on an apostrophe if it's unescaped. Thinking on it further, there are a few other edge cases which would potentially break the current codeso I'll go ahead and reduce it to matching the start quote of the string.

PR #3613, opened 2022-05-05
AUGMENTATIONS: Fix NeuroFlux being applied improperly and migrate broken saves

nickofolas:
fixes a regression introduced in a recent refactor wherein NeuroFlux Governor was not properly being applied, and instead accumulating on the player object as pictured

image

this PR fixes the issue, and includes a save file migration to fix saves affected


src/Augmentation/AugmentationHelpers.tsx
```diff
@@ -85,9 +85,17 @@ function applyAugmentation(aug: IPlayerOwnedAugmentation, reapply = false): void
     Player.applyEntropy(Player.entropy);
   }
 
+  // Special logic for NeuroFlux Governor
+  const ownedNfg = Player.augmentations.find((pAug) => pAug.name === AugmentationNames.NeuroFluxGovernor);
+  if (aug.name === AugmentationNames.NeuroFluxGovernor && !reapply && ownedNfg) {
```
phyzical, line 90: i wonder if this should be `(!reapply || ownedNfg)`

phyzical (approved):

src/Augmentation/AugmentationHelpers.tsx
phyzical: lol nvm that wouldnt work

nickofolas: i don't think so - when reapplied, the level from the save object is what's applied, so that's not a concern. if `ownedNfg` is undefined, then the special logic will be skipped and the player's first NFG should be applied

PR #3618, opened 2022-05-06
API: Terminal screen can now be cleared from within scripts with ns.ui.clearTerminal()

Hoekstraa:
# API addition

## Based on
User request from Discord:
Screenshot of user request

## Tests

- Verified to work when in other screens than Terminal
- Log result tested
- RAM usage of script verified
- `npm run format` and `npm run lint` were ran

### DIFF
--- a/src/ScriptEditor/NetscriptDefinitions.d.ts
+++ b/src/ScriptEditor/NetscriptDefinitions.d.ts
@@ -4395,7 +4395,7 @@ interface UserInterface {
   /**
    * Clear the Terminal window, as if the player ran `clear` in the terminal
    * @remarks
-   * RAM cost: 0 GB
+   * RAM cost: 0.2 GB
    */
   clearTerminal(): void;
 }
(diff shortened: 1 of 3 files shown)


src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -4391,6 +4391,13 @@ interface UserInterface {
    * RAM cost: 0 GB
    */
   getGameInfo(): GameInfo;
+
+  /**
+   * Clear the Terminal window, as if the player ran `clear` in the terminal
+   * @remarks
+   * RAM cost: 0 GB
```
nickofolas, line 4398: Should this be `* RAM cost: 0.2 GB`? The constant RAM cost is set to `0.2`, which I believe should match

Hoekstraa: Changed. Was apparently missed due to me going back and forth on RAM usage a few times. Woops!

phyzical (approved):

PR #3624, opened 2022-05-07
MISC: Implemented infinite loop safety net.

hydroflame:
Screen Shot 2022-05-07 at 5 44 16 PM
Screen Shot 2022-05-07 at 5 44 04 PM


taralx (approved):

Harg2001:
Hello,
I think there is something wrong, because the safety net kill some of my simple script despite there is a sleep within the loop.
Did you count the 1000ns as global instead of per script ? because i run 1000 scripts simultaneously (obviously).

PR #3627, opened 2022-05-07
MISC: Wrap most of the API in the new api wrapper

hydroflame:


src/NetscriptFunctions.ts
```diff
@@ -360,7 +367,7 @@ export function NetscriptFunctions(workerScript: WorkerScript): NS {
     }
   };
 
-  const hack = function (
+  const hack = async function (
```
taralx, line 370: Not clear on why this was changed.

taralx (commented):
It seems like there's a lot of work here to pass context as a parameter, but each of these is in a generative function. What's the advantage of this function-returning-a-function over just having them close over the context?

PR #3628, opened 2022-05-08
UI: Add a quick re-buy option on ascension modal

borisflagell:
Add a button to the Gang Ascension modal allowing for quick re-purchasing of upgrade owned by the gang member.

Also, some old territory-related console.log() was running in the background. => deleted it.

image

Known issue : 
the button Tooltip is overshadowed by Modal => already handled by nickofolas in PR#3620

Testing : 
Vanilla Ascend button still work as intended.
Re-purchasing works as intended. No uprade is given for free, no negative money shenanigan, etc...

### DIFF
--- a/src/Gang/ui/AscensionModal.tsx
+++ b/src/Gang/ui/AscensionModal.tsx
@@ -118,7 +118,7 @@ export function AscensionModal(props: IProps): React.ReactElement {
             confirm(true);
           }}
         >
-          Ascend + Re-Equip
+          Ascend and re-purchase upgrade.
         </Button>
       </Tooltip>
     </Modal>
--- a/src/Gang/ui/AscensionModal.tsx
+++ b/src/Gang/ui/AscensionModal.tsx
@@ -33,14 +33,17 @@ export function AscensionModal(props: IProps): React.ReactElement {
 
   function confirm(reEquip = false): void {
     props.onAscend();
-    const arrEquipement = Object.assign([], reEquip ? props.member.upgrades : null);
+    const arrEquipment = Object.assign([], reEquip ? props.member.upgrades : null);
     const res = gang.ascendMember(props.member);
-    arrEquipement?.forEach((equipement) => {
-      const upg = GangMemberUpgrades[equipement];
-      if (player.money >= gang.getUpgradeCost(upg)) {
-        props.member.buyUpgrade(upg, player, gang);
-      }
-    });
+
+    if (arrEquipment) {
+      arrEquipment.forEach((equipment) => {
+        const upg = GangMemberUpgrades[equipment];
+        if (player.money >= gang.getUpgradeCost(upg)) {
+          props.member.buyUpgrade(upg, player, gang);
+        }
+      });
+    }
 
     dialogBoxCreate(
       <>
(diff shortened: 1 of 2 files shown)


nickofolas:
i'm not entirely sure i agree with this - it feels like something players should need to implement themselves via scripts, rather than have with a button.

just my opinion though

borisflagell:
>  it feels like something players should need to implement themselves via scripts, rather than have with a button.

It crossed my mind too.
So I pulled my punch up a little and cut down on some initially planned feature.

It won't tell you how much money you need to buy back your upgrade.
It won't autobuy new upgrade if you can afford them.
It won't tell you anything when you cannot afford something.

Hopefully, it should leave enough room for improvement; so that player will still feel the need to come up with their own automation.


I see it as another kind of "buy MAX" button : it will save you the trouble to click hundreds of times like a madman; but it won't take decision for you.

src/Gang/ui/AscensionModal.tsx
```diff
@@ -92,7 +104,23 @@ export function AscensionModal(props: IProps): React.ReactElement {
...
+      >
+        Ascend
+      </Button>
+      <Tooltip title={<Typography>Will attempt to buy back all current upgrade after ascending.</Typography>}>
+        <Button
+          onClick={() => {
+            confirm(true);
+          }}
+        >
+          Ascend and re-purchase upgrade.
```
ApamNapat, line 121: Shouldn't this be `upgrades`?

ApamNapat: Same above

src/Gang/ui/AscensionModal.tsx
```diff
@@ -20,16 +23,25 @@ interface IProps {
...
 
   useEffect(() => {
     const id = setInterval(() => setRerender((old) => !old), 1000);
     return () => clearInterval(id);
   }, []);
 
-  function confirm(): void {
+  function confirm(reEquip = false): void {
     props.onAscend();
+    const arrEquipement = Object.assign([], reEquip ? props.member.upgrades : null);
```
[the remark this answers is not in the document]
ApamNapat: And the new logic could be inside an `if` guarded by the new variable, instead of being mixed into the generic ascension logic

ApamNapat (changes requested):

src/Gang/ui/AscensionModal.tsx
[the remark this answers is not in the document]
borisflagell:
```
if (arrEquipment) {
      arrEquipment.forEach((equipment) => {
        const upg = GangMemberUpgrades[equipment];
        if (player.money >= gang.getUpgradeCost(upg)) {
          props.member.buyUpgrade(upg, player, gang);
        }
      });
    }
```
Is that what you had in mind ?

For my culture : 
What's the idea behind that ? Is that some kind of future-proofing good practice ? Or just for clarity's sake ?
Functionnally, it's not changing anything. There was an implicit _if_ already in the previous syntax.

src/Gang/ui/AscensionModal.tsx
borisflagell: Applied.

Issue #3566, opened 2022-04-26
Request: easier Sleeve Task interface


vladtepesch:
Restoring a sleeves task currently is tedious and thanks to https://github.com/danielyxie/bitburner/issues/2884 impossible for some tasks.

May be an function could be added that takes a `SleeveTask` object.

then the following workflow would be possible

```
let oldTask = ns.sleeve.getTask(someId);
// set to some other tasks
ns.sleeve.setTask(someId, oldTask);
```

borisflagell:
#2884 will be fixed by PR #3631.

Concerning a SetTask function. The current consensus is that it would go against the game philosophy. It's up to the player to write something to make the process less tedious.

PR #3631, opened 2022-05-08
API: fix to a bunch of sleeve issue.

### LINKED ISSUE #2700: Feature: Add function to get base price of augment rather than current price(including the multiplier) or a function to sleeve that returns the price of their augments
This is more of a readability thing, but I like my scripts to output what the cost of something is when they are waiting to buy it. At the moment there isn't a way to get the price of an augment for a sleeve, unless I am missing it which would be ideal. For since sleeve augs dont increase in price, they are only the same price as normal augs when you don't have any queued.

### LINKED ISSUE #2884: Sleeve 'getTask' SleeveTask is missing className
https://github.com/danielyxie/bitburner/blob/master/src/NetscriptFunctions/Sleeve.ts#L166
sl.className is required to determine which class the sleeve is currently enrolled in, similar to gymStatType

SleeveTask interface - https://github.com/danielyxie/bitburner/blob/master/src/ScriptEditor/NetscriptDefinitions.d.ts#L1025

### LINKED ISSUE #3132: sleeve.setToCompanyWork behaviour different from UI
You can place all sleeves to work for the same company when placing through the UI. When trying the same through the API, it's only possible to place one sleeve per company:
Screenshot 2022-03-13 083805
Screenshot 2022-03-13 083951
Screenshot 2022-03-13 084043

Also the error message is wrong.

https://github.com/danielyxie/bitburner/blob/bb433d4629d436ad18baf1338a93d2306a85011f/src/NetscriptFunctions/Sleeve.ts#L82

### LINKED ISSUE #3366: Sleeve UI not updating for gym stat when assigned via script
When assigning a sleeve to workout at the gym via script like this:
`ns.sleeve.setToGymWorkout(2, c.GymName.powerhouse, 'def');`
... the UI always shows "Train Strength" even though a different stat is being trained.  

See screenshot below where Defense Exp is increasing but the UI shows "Train Strength".
image

### LINKED ISSUE #3394: Assigned 2 sleeves to work for faction but only one faction allowed workers
# DELETE THIS AFTER READING

# include (where relevant)

- [ ] Save file
- [1. Joined 1 faction that allowed sleeves to work for them. [Aevum was the faction I used]
-  2. selected 'work for Faction' on 2 different sleeves.
-  3. started first sleeve while other sleeve was still on 'work for faction'] Steps to reproduce
- [1.6.3 ] Version of the game, e.g. Bitburner v1.3.0 (216bf616)

### LINKED ISSUE #3556: Abrupt crash while setting task for sleeve
# DELETE THIS AFTER READING

# include (where relevant)

- [ ] Save file
- [ ] Steps to reproduce
- [ ] Version of the game, e.g. Bitburner v1.3.0 (216bf616)

### LINKED ISSUE #3661: `ns.sleeve.getSleeveStats()` should include sleeve memory
Doesn't appear to be any in-game way to programmatically check on sleeve memory. Could trivially be added to a property of SleeveStats

borisflagell:
(Updated for 1.7.0, re-tested all fix, all seems fine)

Working & Clean : 
fix #2884 - Add the missing className property to NetScript SleeveTask object.
fix #3394 - Same issue as 3132
fix #3132 - Fixed the Sleeve UI TaskSelector to make sure no other sleeve already works in the same company.
fix #3661 - add the missing memory property to sleeve.getStat() function return value.
fix #3366 - Sleeve.gymStatType will now be constrainded to the following values by Sleeve.workoutAtGym() :
"none" , "str", "def", "dex", "agi". (could previously be anything inputted by player through API)
Sleeve.updateGainTaskRate() and TaskSelector.tasks's initializer block will now use a switch, expecting one of those value, instead of reevaluating the string.
fix #2700 -  Tweaked  existing function : ( by adding a returnBaseCost boolean. )
- singularity.GetAugmentationPrice
- singularity.GetAugmentationRepReq
Added new functions :
- sleeve.GetSleeveAugmentationPrice
- sleeve.GetSleeveAugmentationRepReq

Discarded : 
put on hold 3187 fix - Player's ApplyForJob() and QuitJob() function now check for sleeve currently working in the impacted company. If one is found, its current job will be accordingly terminated or updated. CANCELLED. Was an akward way to solve the issue. Similar to the above issue. Can wait for the whole system overhaul.


phyzical:
i think the fact that sleeves can work for the same company is actually a ui bug, and shouldn't be possible

phyzical:
also idk about allowing an ns object to take in an object, it feels like something that should be user defined logic to call the appropriate functions. but ill let others chime in

borisflagell:
I agree, 2284 is the only legitimate change.

3566 is arguably a QoL. Arguably. 

3394 is probably a step in the wrong direction.

borisflagell:
Regarding company work :
As its turn out, Hydroflame as previously stated that ns function behaved correctly, and that UI was not in #3132.
I will commit my code accordingly ASAP.

src/PersonObjects/Sleeve/ui/TaskSelector.tsx
```diff
@@ -226,8 +227,18 @@ function getABC(sleeve: Sleeve): [string, string, string] {
...
+      if (sanitizedStat.includes("str")) {
+        return ["Workout at Gym", "Train Strength", sleeve.currentTaskLocation];
+      } else if (sanitizedStat.includes("def")) {
+        return ["Workout at Gym", "Train Defense", sleeve.currentTaskLocation];
+      } else if (sanitizedStat.includes("dex")) {
+        return ["Workout at Gym", "Train Dexterity", sleeve.currentTaskLocation];
+      } else if (sanitizedStat.includes("agi")) {
+        return ["Workout at Gym", "Train Agility", sleeve.currentTaskLocation];
+      }
+    }
```
Ornedan, line 241: While this does kind of work, it doesn't solve the real problem, which is that `Sleeve.workoutAtGym()` stores arbitrary user input in `Sleeve.gymStatType` field that then needs to be identified over and over again wherever it is used.

The correct solution is to normalise the user input to one of an enumerated list of strings (eg. `"", "str", "def", "dex", "agi"`), and then other code that uses `Sleeve.gymStatType` can rely on it being one of those known values.

src/NetscriptFunctions/Sleeve.ts
```diff
@@ -174,6 +172,84 @@ export function NetscriptSleeve(player: IPlayer, workerScript: WorkerScript, hel
 
       return player.sleeves[sleeveNumber].workoutAtGym(player, gymName, stat);
     },
+    setToWork: function (_sleeveNumber: number, _sleeveTask: SleeveTask): boolean {
```
Ornedan, line 175: Should be named `setToTask`, as the functionality covers more than just working for companies/factions.

But also, I'm not sure if this is even an appropriate addition as it's redundant with all the individual `sleeve.setTo*` functions. The requested feature in #3566 is something a player can implement themselves on top of the existing Sleeve API, and in my opinion should have to if they want to have a helper function that works in that specific way.

src/PersonObjects/Sleeve/ui/TaskSelector.tsx
borisflagell: Agreed. I'll get back to it if I have some time to spare.

If not... well it is working, and resolve an actual issue. It will be shipped as is.

src/PersonObjects/Sleeve/ui/TaskSelector.tsx
```diff
@@ -45,13 +45,14 @@ function possibleJobs(player: IPlayer, sleeve: Sleeve): string[] {
...
-    }
-  }
 
-  return allJobs;
+  return allJobs.filter((company) => {
+    if (forbiddenCompanies.includes(company)) {
+      return false;
+    } else {
+      return true;
+    }
```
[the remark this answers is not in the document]
borisflagell: Nice catch! I'll use that.

src/NetscriptFunctions/Sleeve.ts
borisflagell: That's a sound argument. Phyzical had a similar reaction. And in fact, I think so too. (more than  half the reason I did that coding in the first place was to familiarize myself with netscript)

I will revert these change. #3566 will be discarded. (well, at least the underlying issue should be fixed by patching the SleeveTask interface)

src/NetscriptFunctions/Singularity.ts
```diff
@@ -133,18 +133,18 @@ export function NetscriptSingularity(player: IPlayer, workerScript: WorkerScript
         return aug.prereqs.slice();
       },
     getAugmentationPrice: (_ctx: NetscriptContext) =>
-      function (_augName: unknown): number {
+      function (_augName: unknown, _basePrice: unknown = false): number {
         _ctx.helper.checkSingularityAccess();
         const augName = _ctx.helper.string("augName", _augName);
         const aug = getAugmentation(_ctx, augName);
-        return aug.getCost(player).moneyCost;
+        return !_basePrice ? aug.getCost(player).moneyCost : aug.baseCost;
```
phyzical, line 140: dont forget to bump the bool through the helper, although i wonder how useful the base price would be considering the sleeve function exists

`const returnBasePrice = _ctx.helper.boolean("returnBasePrice", _returnBasePrice);`

phyzical: imo it makes sense to me to keep it clean and separate. the basePrice outside of sleeves is essentially source code at this point imo

src/PersonObjects/Player/PlayerObjectGeneralMethods.tsx
```diff
@@ -1850,6 +1861,14 @@ export function quitJob(this: IPlayer, company: string, _sing = false): void {
...
     this.companyName = this.hasJob() ? Object.keys(this.jobs)[0] : "";
   }
+
+  if (this.sleeves !== null) {
+    for (const sleeve of this.sleeves) {
+      if (sleeve.currentTask === SleeveTaskType.Company && sleeve.currentTaskLocation === company) {
+        sleeve.finishTask(this);
+      }
+    }
+  }
```
phyzical, line 1871: it feels like this should be handled by the sleeves "quitJob" 

as wont this just quit your sleeves job when you quit as the player at company X?

src/PersonObjects/Player/PlayerObjectGeneralMethods.tsx
```diff
@@ -1794,6 +1795,16 @@ export function applyForJob(this: IPlayer, entryPosType: CompanyPosition, sing =
...
   this.jobs[company.name] = pos.name;
+
+  if (this.sleeves !== null) {
+    for (const sleeve of this.sleeves) {
+      if (sleeve.currentTask === SleeveTaskType.Company && sleeve.currentTaskLocation === company.name) {
+        sleeve.finishTask(this);
+        sleeve.workForCompany(this, company.name);
+      }
+    }
+  }
```
phyzical, line 1806: and this feels out of place, like again if the player is applying to work for X wont this just assume you want 1 of the sleeves to work for this same company X also?

src/PersonObjects/Player/PlayerObjectGeneralMethods.tsx
borisflagell: Yeah, I'm aware. The whole working system is a mess right now.
It does works; but it's spaghetti coding.

I'm considering discarding theses changes for the time being, and wait for the working system rework (It's already on hydro's TO-DO).

src/PersonObjects/Player/PlayerObjectGeneralMethods.tsx
borisflagell: I think not, it's only doing anything if one of your sleeves _already_ works in the concerned company. 
Idle sleeve, and sleeve working for another company won't enter the if block.

We should be fine.

But yeah, nonetheless, PlayerObjectGeneralMethod interacting directly with sleeve is awkward.

borisflagell: Reverted that change.

src/PersonObjects/Player/PlayerObjectGeneralMethods.tsx
borisflagell: Reverted that change.

src/PersonObjects/Sleeve/ui/TaskSelector.tsx
borisflagell: Done !

phyzical (approved):

borisflagell:
Will be split into indiviual PR/issue for an easier merger.

PR #3632, opened 2022-05-08
Fix infiltration rep BN mult calculation

trambelus:


phyzical (approved):
swear we fixed this already ahwell 👍

PR #3633, opened 2022-05-08
UI: FIX #2829 Remove defeated NPC gangs from territory page

### LINKED ISSUE #2829: Remove defeated NPC gangs from territory page
AFAICT, once an NPC gang is reduced to 0 territory, it will never regain any nor will it clash with you again. However, these defeated gangs continue to be listed on the 'Territory' tab, slowly regaining power that will never affect anything. The interface would be cleaner if these gangs were removed from the list entirely - possibly instead being added to a separate list on the same page of 'defeated opponents' (without, of course, showing their now-irrelevant power or win chance).

Version: a3c99e6

Dane-Horn:
fixes #2829 
Filters non player gangs with 0 territory out of the territory list, screenshots below:
0-gangs
1-gang
2-gangs
3-gangs
4-gangs
5-gangs
6-gangs

### DIFF
--- a/src/Gang/ui/TerritorySubpage.tsx
+++ b/src/Gang/ui/TerritorySubpage.tsx
@@ -86,7 +86,11 @@ export function TerritorySubpage(): React.ReactElement {
       </Box>
       <Box sx={{ display: "grid", gridTemplateColumns: "repeat(3, 1fr)" }}>
         {gangNames
-          .filter((name) => AllGangs[name].territory > 0)
+          .sort((a, b) => {
+            if (AllGangs[a].territory <= 0 && AllGangs[b].territory > 0) return 1;
+            if (AllGangs[a].territory > 0 && AllGangs[b].territory <= 0) return -1;
+            return 0;
+          })
           .map((name) => (
             <OtherGangTerritory key={name} name={name} />
           ))}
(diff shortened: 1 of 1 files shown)


src/Gang/ui/TerritorySubpage.tsx
```diff
@@ -85,9 +85,11 @@ export function TerritorySubpage(): React.ReactElement {
         </Typography>
       </Box>
       <Box sx={{ display: "grid", gridTemplateColumns: "repeat(3, 1fr)" }}>
-        {gangNames.map((name) => (
-          <OtherGangTerritory key={name} name={name} />
-        ))}
+        {gangNames
+          .filter((name) => AllGangs[name].territory > 0)
```
nickofolas, line 89: I think it might be a bit unpleasant from a UI perspective to remove defeated gangs entirely.

I might suggest something like sorting defeated gangs to the back, and lowering their element opacity (like how it's done in the `Create Program` page

```suggestion
          .sort((a, b) => {
            if (AllGangs[a].territory <= 0) return 1;
            if (AllGangs[b].territory <= 0) return -1;
            return 0;
          })
```

And then reduce the opacity by like 25% for enemy gangs with 0 territory

image

Dane-Horn:
Updated. Reorders 0 territory gangs to the back without affecting those with territory and reduces their opacity to 0.75:
image

nickofolas (approved):

PR #3637, opened 2022-05-09
STANEK: Fix #3196 Charging booster fragments throws an error

### LINKED ISSUE #3196: Charging boosters should throw error

Undeemiss:
Added a check to ns.stanek.chargeFragment(x, y) to throw an error if the selected fragment is a booster. Technically a breaking change in some sense, because it now throws an error in a case where previously it would simply accomplish nothing. If a user script relies on this fringe behavior, it will no longer work. 
image
Tested by attempting an invalid charge on an empty space, a valid charge on a chargeable fragment, and an invalid charge on a booster fragment. All three functioned as expected.
Closes #3196

### DIFF
--- a/src/NetscriptFunctions/Stanek.ts
+++ b/src/NetscriptFunctions/Stanek.ts
@@ -5,6 +5,7 @@ import { netscriptDelay } from "../NetscriptEvaluator";
 
 import { staneksGift } from "../CotMG/Helper";
 import { Fragments, FragmentById } from "../CotMG/Fragment";
+import { FragmentType } from "../CotMG/FragmentType";
 
 import {
   Fragment as IFragment,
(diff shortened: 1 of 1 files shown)


src/NetscriptFunctions/Stanek.ts
```diff
@@ -42,11 +43,19 @@ export function NetscriptStanek(
       },
     chargeFragment: (_ctx: NetscriptContext) =>
       function (_rootX: unknown, _rootY: unknown): Promise<void> {
+        //Get the fragment object using the given coordinates
         const rootX = _ctx.helper.number("rootX", _rootX);
         const rootY = _ctx.helper.number("rootY", _rootY);
         checkStanekAPIAccess("chargeFragment");
         const fragment = staneksGift.findFragment(rootX, rootY);
+        //Check whether the selected fragment can ge charged
```
phyzical, line 51:
```suggestion
        //Check whether the selected fragment can get charged
```

phyzical (approved):

PR #3638, opened 2022-05-09
MISC: Fix #3125 BREAKING Renamed BN mult CorporationSoftCap to CorporationSoftcap

### LINKED ISSUE #3125: CorporationSoftCap has different capitalization from other Softcaps
https://github.com/danielyxie/bitburner/blob/bb433d4629d436ad18baf1338a93d2306a85011f/src/BitNode/BitNodeMultipliers.ts#L292

I just find it a bit inconsistent that the BitNode multiplier CorporationSoftCap has different capitalization than the other two Softcaps which have a small 'c' in 'cap' (Gang and PurchasedServer). Though PurchasedServerSoftcap goes even further and has a large 'C' in the documentation (.md) but a small 'c' in the code.

Undeemiss:
Replaced all instances of "CorporationSoftCap" in src/ with "CorporationSoftcap" for parity with other instances of "Softcap". By my testing, this does not seem to require a save file migration except for any user scripts that have hardcoded the name of this property, which is obviously not reasonable to migrate automatically.

Tested by loading a BN8 save from the current Steam build and ensuring that the CorporationSoftCap multiplier was properly replaced by CorporationSoftcap.

Resolves #3125

### DIFF
--- a/src/BitNode/BitNode.tsx
+++ b/src/BitNode/BitNode.tsx
@@ -488,7 +488,7 @@ export const defaultMultipliers: IBitNodeMultipliers = {
   FourSigmaMarketDataApiCost: 1,
 
   CorporationValuation: 1,
-  CorporationSoftCap: 1,
+  CorporationSoftcap: 1,
 
   BladeburnerRank: 1,
   BladeburnerSkillCost: 1,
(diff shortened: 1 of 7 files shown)


Undeemiss:
I see this didn't build properly. Investigating now.

phyzical (commented):
hmm doesnt this need a save file migration? 
or is it fine since we moved away to static multiplier calculator functions

Undeemiss:
I don’t *think* it needs one? It didn’t seem that way to me based on a bit of testing, but I didn’t dig too deep once it seemed to work.

phyzical (approved):

PR #3642, opened 2022-05-10
OPTIONS: Fix sliders not sliding correctly

nickofolas:
fixes options sliders so that they don't freeze up every time their value is changed

firefox_1ozrr8IkMe

the slider value is now computed within the slider element, and is only released to the user `Settings` once the slider is no longer being used (this can be observed in the above GIF as the save icon on the character overview only changes state once the slider thumb is released)

### Testing
- [x] I have confirmed that the sliders can be slid

### DIFF
--- a/src/GameOptions/ui/OptionsSlider.tsx
+++ b/src/GameOptions/ui/OptionsSlider.tsx
@@ -1,8 +1,8 @@
 import { Slider, Tooltip, Typography, Box } from "@mui/material";
-import React from "react";
+import React, { useState } from "react";
 
 interface IProps {
-  value: any;
+  initialValue: any;
   callback: (event: any, newValue: number | number[]) => void;
   step: number;
   min: number;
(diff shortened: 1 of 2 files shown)


src/GameOptions/ui/OptionsSlider.tsx
```diff
@@ -13,14 +13,21 @@ interface IProps {
...
     <Box>
       <Tooltip title={<Typography>{props.tooltip}</Typography>}>
         <Typography>{props.label}</Typography>
       </Tooltip>
       <Slider
-        value={props.value}
-        onChange={props.callback}
+        value={value}
+        onChange={onChange}
+        onChangeCommitted={props.callback}
```
phyzical, line 30: nit:
```suggestion
        onChangeEmitted={props.callback}
```

phyzical (approved):

src/GameOptions/ui/OptionsSlider.tsx
nickofolas: `onChangeEmitted` isn't a part of MUI's slider API

phyzical: oh derp thought it was a custom component 👍

PR #3653, opened 2022-05-14
TERMINAL: FIX #3651 Make directory name regex more flexible

### LINKED ISSUE #3651: Valid folder names are not valid in game
Version 1.6.4
Run `nano testv1.0/test.js` in the in-game terminal: 
image
Same folder name is valid in VS Code, for example:
image

Additionally, the regex used to validate directory names is out of sync with the comments above it:
```ts
export function isValidDirectoryName(name: string): boolean {
  // Allows alphanumerics, hyphens, underscores, and percentage signs.
  // Name can begin with a single period, but otherwise cannot have any
  const regex = /^.?[a-zA-Z0-9_-]+$/;

  // match() returns null if no match is found
  return name.match(regex) != null;
}
```
Suggestion to adjust this regex to match more flexible folder names that are still legal and won't interfere with the directory system.

Dane-Horn:
Fixes #3651  

Initial change to expand the set of allowed directory names. Draft PR while investigating the potential consequences of this.

### DIFF
--- a/src/Terminal/DirectoryHelpers.ts
+++ b/src/Terminal/DirectoryHelpers.ts
@@ -51,7 +51,7 @@ export function isValidFilename(filename: string): boolean {
 export function isValidDirectoryName(name: string): boolean {
   // Allows alphanumerics, hyphens, underscores, and percentage signs.
   // Name can begin with a single period, but otherwise cannot have any
-  const regex = /^.?[a-zA-Z0-9_-]+$/;
+  const regex = /^\.?(?:\w[_\-.%]*)+$/;
 
   // match() returns null if no match is found
   return name.match(regex) != null;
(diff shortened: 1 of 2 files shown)


fdellwing:
I have to say, I find that list of valid characters a bit random? What is the goal?

Being like linux, than you have to allow nearly every single character besides `/`, because well `f<o>o|b:a&r\b aßz` would be a valid directory name.

Just being some sane base? Than including `%` seems strange.

Dane-Horn:
The new regex is mostly built up based on the previous comments provided in that function, hence allowing `%`. The only real extension added so far is the ability to use `.` anywhere in the name besides the end as an easy catch for the `.` and `..` cases.
I'm by no means an expert on what should be allowed here, so open to suggestions.

phyzical:
@Dane-Horn looking good, 
maybe just a suggestion add a few more complicated directory names in the tests, like the ones that prompted you to make the fix in the first place

phyzical (approved):

PR #3657, opened 2022-05-15
STANEK: Fix  #3623 Ugly fix to ns.stanek functions engine leak

### LINKED ISSUE #3623: Stanek API leaks engine classes
It is possible to access engine internal classes and their methods via the objects returned from `ns.stanek.fragmentDefinitions()` and `ns.stanek.getFragment()`:

```js
/** @param {NS} ns */
	const def0 =  ns.stanek.fragmentDefinitions()[0];

	ns.stanek.clearGift();
	ns.stanek.placeFragment(0, 0, 0, def0.id);
	const placed = ns.stanek.getFragment(0, 0);
	ns.stanek.clearGift();

	// Grab class objects
	const Fragment = def0.__proto__;
	const ActiveFragment = placed.__proto__;

	console.log(Fragment, ActiveFragment);
}
```

Prototype editing is possible, but the most effective exploit is to jump from `ActiveFragment.fragment()` to editing the fragment definitions. `power` and `shape` letting a player set arbitrary multipliers to any stat modifiable via Stanek.

```js
/** @param {NS} ns */
	// Clear the board
	ns.stanek.clearGift();

	// Modify the fragments: boost power and reshape to 1x1
	const statFrags = ns.stanek.fragmentDefinitions().filter(frag => frag.type != 18);
	for (const def of statFrags) {
		ns.stanek.placeFragment(0, 0, 0, def.id);
		const realDef = ns.stanek.getFragment(0, 0).fragment();
		ns.stanek.removeFragment(0, 0);

		realDef.power = 1e6; // PHENOMENAL COSMIC POWER
		realDef.shape = [[true]]; // itty bitty living space
	}

	// Place all fragments
	for (let y = 0; y < ns.stanek.giftHeight(); y++) {
		for (let x = 0; x < ns.stanek.giftWidth(); x++) {
			const def = statFrags.pop();
			if (def)
				ns.stanek.placeFragment(x, y, 0, def.id);
		}
	}

	// Charge all fragments
	for (const afrag of ns.stanek.activeFragments())
		await ns.stanek.chargeFragment(afrag.x, afrag.y);
}
```

borisflagell:
Reimplement the Fragment and ActiveFragment class, without any method attached. Uses them as a return value rather than the original class.

Feels likes too much code for too few added-value. 
There's probably a cleaner way to do it.

### DIFF
--- a/src/NetscriptFunctions/Stanek.ts
+++ b/src/NetscriptFunctions/Stanek.ts
@@ -18,6 +18,25 @@ import { FactionNames } from "../Faction/data/FactionNames";
 import { joinFaction } from "../Faction/FactionHelpers";
 import { Factions } from "../Faction/Factions";
 
+class APIActiveFragment implements IActiveFragment {
+  id: number;
+  highestCharge: number;
+  numCharge: number;
+  rotation: number;
+  x: number;
+  y: number;
+  
+  constructor (src:IActiveFragment) {
+    this.id = src.id;
+    this.highestCharge = src.highestCharge;
+    this.numCharge = src.numCharge;
+    this.rotation = src.rotation;
+    this.x = src.x;
+    this.y = src.y;
+  }
+}
+
+
 export function NetscriptStanek(
   player: IPlayer,
   workerScript: WorkerScript,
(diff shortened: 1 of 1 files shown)


hydroflame:
JSON.stringify

hydroflame:
JSON.stringify

Issue #3585, opened 2022-04-29
Ability to specify <x,y> coordinates as optional parameters to ns.tail()


dleblond312:
It would help a lot with augment installs if various log output screens could be opened at various locations instead of stacked on each other in the middle of the screen.

JohnnyQuazar:
Great idea! I thought I opened a similar issue but apparently not, will ride your coattails :)

We also need (NEED) a way to declare width and height.

MatiasCardullo:
Yes please

OriginalRobotWizard:
@hydroflame To answer the question: Optional parameters.

BUT: I highly suggest to change this into 2 new functions, because it's much more versatile:
- moveTail(x, y)
- resizeTail(w, h)

The cherry on the cake would be 
- closeTail()

OriginalRobotWizard:
Right, tail has the possibility to just open any log, not just the one of the own script. Interesting, but I never used it that way, so I did not know.

Anyway, then the 3 new functions can solve the needs?

vladtepesch:
maybe `tail` could return a handle that has some member functions like `move`, `close`, `resize`, `clear`
This would have the advantage to not clutter the ns-namespace.

PR #3666, opened 2022-05-17
MISC: Added NS function closeTail to close tail windows

Undeemiss:
Old, and now inaccurate description of this PR, kept for posterity (See later comment for new description):

Attempted to implement `ns.closeTail()`, `ns.closeTail(pid)`, and `ns.closeTail(fn, hostname, ...args)`. Currently, `ns.closeTail()` works, `ns.closeTail(pid)` works if the given pid points to a script that's still running, and `ns.closeTail(fn, hostname, ...args)` works. It has not been tested extensively, and has not been formatted/linted because it's a WIP.

Related to #3585

### DIFF
--- a/src/NetscriptFunctions.ts
+++ b/src/NetscriptFunctions.ts
@@ -193,7 +193,7 @@ export function NetscriptFunctions(workerScript: WorkerScript): NS {
       throw makeRuntimeRejectMsg(
         workerScript,
         `Invalid scriptArgs argument passed into getRunningScript() from ${callingFnName}(). ` +
-        `This is probably a bug. Please report to game developer`,
+          `This is probably a bug. Please report to game developer`,
       );
     }
 
@@ -826,7 +826,8 @@ export function NetscriptFunctions(workerScript: WorkerScript): NS {
         workerScript.log(
           "weaken",
           () =>
-            `'${server.hostname}' security level weakened to ${server.hackDifficulty
+            `'${server.hostname}' security level weakened to ${
+              server.hackDifficulty
             }. Gained ${numeralWrapper.formatExp(expGain)} hacking exp (t=${numeralWrapper.formatThreads(threads)})`,
         );
         workerScript.scriptRef.onlineExpGained += expGain;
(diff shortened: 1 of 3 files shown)


src/ui/React/LogBoxManager.tsx
```diff
@@ -131,6 +142,8 @@ function LogWindow(props: IProps): React.ReactElement {
     return () => clearInterval(id);
   }, []);
 
+  //TODO Not actually a todo I just want this file to show up as changed
+  //so I can find it easier later
```
hydroflame, line 146: rm

Undeemiss:
New description of PR:

Implemented NS function `closeTail(pid?: number)`, which closes any open tail window with the given pid. If no pid is given, closes any open tail window of the calling script. 

Tested by ensuring it properly handles the no input case, as well as pids corresponding to dead scripts, running scripts, pids that have never been used, and pids that correspond to tail windows that have run a new copy of the same script.
Notably, it does nothing, not even giving an error, when given a pid that does not correspond to a script associated with a tail window, and tail windows are only associated with the most recent pid (i.e. if you run some script test.js with pid 1, close it via a tail window, and rerun it with pid 2, the tail window will only correspond to pid 2)

Related to #3585, but does not satisfy it completely

src/NetscriptFunctions.ts
```diff
@@ -988,6 +988,15 @@ export function NetscriptFunctions(workerScript: WorkerScript): NS {
 
       LogBoxEvents.emit(runningScriptObj);
     },
+    closeTail: function (pid?: number): void {
```
hydroflame, line 991: `pid?: unknown = workerScript.pid`

then call `helper.number`

Undeemiss: Done, I think. It's not exactly what you suggested, but it matches the way another function right next to it works, and doing what you gave me exactly just gives a syntax error.

hydroflame: The question mark in my example should not have been there

PR #3667, opened 2022-05-17
MISC: Update package-lock.json

Undeemiss:
I don't actually know what this does, but it was done automatically by what I can only assume is NPM, and looks important (plus it's bugging me). The game works fine when this is updated.


hydroflame:
Yeah but see when I do it it reverts it back

PR #3674, opened 2022-05-19
MISC: FIX #3672 (TOR router can no longer be bypassed)

### LINKED ISSUE #3672: Bug: You can buy a personal server named "darkweb"
This allows you to circumvent buying the tor node separately, saving some cash.
It behaves relatively nicely, but should not be possible.

Steps to reproduce;

Prerequisite: Able to be done on any save that doesn't own darkweb at the moment of reproduction.

1. Buy a personal server called "darkweb"
2. ...
3. Profit.
4. use the darkweb like usual, without having bought it.

Exploitation of this bug is possible on dev build as of today.

xRalic:
# The Issue

Prior to this PR, the purchase of the TOR router could be bypassed by purchasing a personal server with the name ``darkweb``. 

The problematic code was the following:

```js
export function hasTorRouter(this: IPlayer): boolean {
  return !!GetServer(SpecialServers.DarkWeb);
}
```

If **any** server with the name ``darkweb`` exists, the function will return true, and the game will pretend that you've purchased TOR.

## Linked issues
fixes #3672 

## Bug fix

- ``darkweb`` now behaves like ``w0rld_d43m0n``, spawning at the start of the game without any connections in the network. By doing this, purchasing servers as ``darkweb`` will create ``darkweb-xx`` instead.
    - To achieve this, ``darkweb`` has been added to ``../Server/data/server.ts``.
    - The ``purchaseTorRouter()`` method no longer creates the ``darkweb``, it simply connects the existing ``darkweb`` to ``home``.
- The ``hasTorRouter()`` method in the Player Object has been refactored as a boolean, to be set as true when the TOR router is purchased.
    - In consequence, any calls to the method have been replaced with a check for the attribute.
- The ``hasTorRouter`` boolean has been added to the reset function, so that the player doesn't keep the router after soft resetting/fluming out. This could perhaps cause an issue in future BitNodes if the TOR router is added the BitNode reward.
- Some more refactoring was needed in places like the ``buy`` command, since the command checked for the existence of the DarkWeb instead of checking for TOR.

The bugfix has been tested by trying to ``buy`` and ``connect`` to the darkweb with and without the router, before and after soft and hard resets, with success.

### DIFF
--- a/src/PersonObjects/IPlayer.ts
+++ b/src/PersonObjects/IPlayer.ts
@@ -49,6 +49,7 @@ export interface IPlayer {
   has4SDataTixApi: boolean;
   hashManager: HashManager;
   hasTixApiAccess: boolean;
+  hasTorRouter: boolean; 
   hasWseAccount: boolean;
   hp: number;
   jobs: IMap<string>;
@@ -205,7 +206,6 @@ export interface IPlayer {
   hasAugmentation(aug: string | Augmentation, installed?: boolean): boolean;
   hasCorporation(): boolean;
   hasGangWith(facName: string): boolean;
-  hasTorRouter(): boolean;
   hasProgram(program: string): boolean;
   inBladeburner(): boolean;
   inGang(): boolean;
--- a/src/PersonObjects/IPlayer.ts
+++ b/src/PersonObjects/IPlayer.ts
@@ -49,7 +49,7 @@ export interface IPlayer {
   has4SDataTixApi: boolean;
   hashManager: HashManager;
   hasTixApiAccess: boolean;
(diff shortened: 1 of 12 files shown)


src/PersonObjects/IPlayer.ts
```diff
@@ -49,6 +49,7 @@ export interface IPlayer {
   has4SDataTixApi: boolean;
   hashManager: HashManager;
   hasTixApiAccess: boolean;
+  hasTorRouter: boolean;
```
hydroflame, line 52: no, no cache data

hydroflame:
Fixed

PR #3677, opened 2022-05-20
UI: Open tail window in external window

U1F984:
POC implementation for opening tail windows in external browser windows.

Improvement ideas:

* split LogBoxManager.tsx into three files (LogBoxContent + cross window message types, Impl for main window, Impl for foreign window)
* Have a separate page for log instead of using hash
* Reduce bundle size by only including log-related stuff for log window (possibly related to point 2)


src/ui/Log/LogBoxManager.tsx
```diff
@@ -80,11 +157,27 @@ export function LogBoxManager(): React.ReactElement {
...
+    rerender();
+  }
+
   return (
     <>
-      {logs.map((log) => (
-        <LogWindow key={log.id} script={log.script} id={log.id} onClose={() => close(log.id)} />
-      ))}
+      {logs
+        .filter((log) => !log.foreignWindow)
```
taralx, line 171: Your render cannot depend on mutable state or it will desync when that state is changed.

U1F984: The previous implementation also had this: logs itself was updated in place, then rerender was called. I can switch up the design, but there might have been a reason to do it manually?

PR #3684, opened 2022-05-21
MISC: FIX Update type in NetscriptDefinitions.d.ts for ns.args[0] being undefined

PathToLife:
Simple one liner PR.
Add undefined type to `ns.args[]` for better type checking

## Details
I've copied over this type file to vscode, and linter didn't warn that args was never undefined.
Hence I was caught out, and had the below error pop up in bitburner.

```js
const target = ns.args[0].toString()
```
image

It may be good to merge this PR so others won't be caught out by this.

For reference, nodejs `process.env` types uses `Dict<string>` which has `undefined` explicitly stated
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/75f98a0b191aa1458658b90b5b1329748cc3c082/types/node/v14/globals.d.ts#L739
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/75f98a0b191aa1458658b90b5b1329748cc3c082/types/node/v14/process.d.ts#L84

### DIFF
--- a/src/ScriptEditor/NetscriptDefinitions.d.ts
+++ b/src/ScriptEditor/NetscriptDefinitions.d.ts
@@ -4532,7 +4532,7 @@ export interface NS {
    * It is also possible to get the number of arguments that was passed into a script using: 'args.length'
    * WARNING: Do not try to modify the args array. This will break the game.
    */
-  readonly args: (string | number | boolean)[];
+  readonly args: (string | number | boolean | undefined)[];
 
   /**
    * Steal a servers money.


hydroflame:
That's incorrect. Your issue is not that the value of index 0 is undefined. It's that the expression `args[0]` is out of bound for the array and so you get undefined.

PathToLife:
I'm developing in typescript, this PR improves the UX by adding a catch for the below edge case

image

PathToLife:
Issue created #3716

PR #3695, opened 2022-05-21
UI: FIX #2962 Add a setting to display middle time unit in Time Elapsed String

### LINKED ISSUE #2962: Don't omit middle components in long-form times
When the GUI displays a duration, e.g. the run-time of a script on the active scripts tab, or the time since installing augs on the stats tab, it formats it with numeric values and written unit names, e.g. "6 hours 3 minutes 2 seconds".

Currently, any time element which is 0 is omitted entirely, e.g. "6 hours 12 seconds" instead of "6 hours 0 minutes 12 seconds". However, it makes such times less readable - it is easy to mistake "6 hours 12 seconds" for "6 hours 12 minutes" at a glance, because we *expect* the unit after 'hour' to be 'minute'. Similarly, if you have multiple dates with the same number of displayed components but different units, its easy to miss that difference in units, e.g. if you have "6 hours 12 minutes" followed by "6 minutes 40 seconds", it's easy to mistake that as two very similar times rather than two times which differ by an order of magnitude.

I propose that this be changed, so that only *leading* components are omitted when 0.

Version: a3c599e6

borisflagell:
Fix #2962 Add a setting to display middle time unit in Time Elapsed String

ex : "1 days 14 seconds" becomes "1 days 0 hours 0 minutes 14 seconds".
"13 hours 55 seconds" becomes "13 hours 0 minutes 55 seconds".

image

(Honestly, I'm not convinced of the added-value of this PR. But regardless of if it will be discarded or merged, at least, the issue will be adressed.)

Let me now if the toggleable settings is uneeded.

### DIFF
--- a/src/GameOptions/ui/CurrentOptionsPage.tsx
+++ b/src/GameOptions/ui/CurrentOptionsPage.tsx
@@ -220,6 +220,12 @@ export const CurrentOptionsPage = (props: IProps): React.ReactElement => {
             <>If this is set all references to memory will use GiB instead of GB, in accordance with IEC 60027-2.</>
           }
         />
+        <OptionSwitch
+          checked={Settings.ShowMiddleNullTimeUnit}
+          onChange={(newValue) => (Settings.ShowMiddleNullTimeUnit = newValue)}
+          text="Show all intermediary times unit, even when null."
+          tooltip={<>ex : 1 hours 13 seconds becomes 1 hours 0 minutes 13 seconds.</>}
+        />
         <Tooltip
           title={
             <Typography>
--- a/src/Settings/Settings.ts
+++ b/src/Settings/Settings.ts
@@ -155,8 +155,8 @@ interface IDefaultSettings {
   UseIEC60027_2: boolean;
 
   /*
-  * Display intermediary time unit when their value is null
-  */
+   * Display intermediary time unit when their value is null
+   */
   ShowMiddleNullTimeUnit: boolean;
 
   /*
--- a/src/Settings/Settings.ts
+++ b/src/Settings/Settings.ts
@@ -155,6 +155,11 @@ interface IDefaultSettings {
   UseIEC60027_2: boolean;
 
   /*
+  * Display intermediary time unit when their value is null
+  */
+  ShowMiddleNullTimeUnit: boolean;
+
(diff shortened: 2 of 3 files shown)


phyzical (approved):
looks fine to me but i wonder if this shouldn't be like show all units instead of just show middle?

borisflagell:
0 days 13 hours 0 minutes 45 seocnds ?

phyzical:
yeah, but i guess its more a question for those that use it.

im assuming that people just like units to be in the same place. feel free to disregard if you dissagree

borisflagell:
After giving it a second thought; I think you're right.
Fixing seconds will prevent the "yo-yo" effect at the 60th second mark.
And fixing the days will make a hypothetetical column of time_elapsed more legible.
I guess adding space or zero to single digit number would be a nice touch too.

WW days XX hours YY minutes ZZ seconds
WW days XX hours YY minutes ZZ seconds
WW days XX hours YY minutes ZZ seconds
WW days XX hours YY minutes ZZ seconds
WW days XX hours YY minutes ZZ seconds

is easier to the eye than 

WW days XX hours YY minutes Z seconds
W days X hours Y minutes
XX hours YY minutes ZZ seconds
WW days XX hours YY minutes ZZ seconds
XX hours YY minutes ZZ seconds

phyzical:
yeah that was what i was thinking too, but wasn't sure if i was making assumptions as to the user base

borisflagell:
It _is_ the kind of feature that will get a really mixed reception. Because it has a really subjective(?) basis behind it.

Should it be optional ? Should there be more option even ? Should the option be put into the user hand ? Or should it rather be left to the dev jugement ? Is it really worth our time to begin with?

In the end, nobody bloody _knows_. It's just personnal opinions... :-S 


Perhaps the best way to deal with it, is to not deal with it ? :-p 
Let's expose the function prototype to the player and let them redefine it to their leisure !

phyzical:
haha template may be the most robust but pain in the ass to make sure it covers all basis,

just go with what youve got or change to the verbose version and we can always come back an address later :)

hydroflame:
Copied to another PR because github wouldn't let me fix MC inside github. All credit to borisflagell.

PR #3698, opened 2022-05-21
BLADEBURNER: FIX #3685 Error being thrown when finishing a BlackOps

### LINKED ISSUE #3685: Bladeburner message popup
assassination

[bitburnerSave_1653117909_BN13x2.zip](https://github.com/danielyxie/bitburner/files/8747162/bitburnerSave_1653117909_BN13x2.zip)

-Latest steam version
-message pops up on bladeburner BlackOps completion. 
Everything else seems to be working properly. Nothing crushes, blackops finishes correctly and lets me do next one.

### LINKED ISSUE #3701: EXCEPTION: Bladeburner.process()
Bitburner 1.7.0 (b71418b0)

Had an exception thrown (screenshot attached) upon completion of the first bladeburner black op.

bladeburner_exception
[bitburnerSave_1653090265_BN10x1.zip](https://github.com/danielyxie/bitburner/files/8748605/bitburnerSave_1653090265_BN10x1.zip)

borisflagell:
fixes #3685 - Bladeburner.action property was reset before being used for the last time in Bladeburner.processAction().
fixes #3701 - duplicate of the issue

Moving
`const action = this.getActionObject(this.action);`
before the call to
`const retValue = this.completeAction(player, player, this.action);`

Thus, ensuring the action state stay available for whatever check needs to be performed below.

### DIFF
--- a/src/Bladeburner/Bladeburner.tsx
+++ b/src/Bladeburner/Bladeburner.tsx
@@ -1677,11 +1677,11 @@ export class Bladeburner implements IBladeburner {
     this.actionTimeOverflow = 0;
     if (this.actionTimeCurrent >= this.actionTimeToComplete) {
       this.actionTimeOverflow = this.actionTimeCurrent - this.actionTimeToComplete;
+      const action = this.getActionObject(this.action);
       const retValue = this.completeAction(player, player, this.action);
       player.gainMoney(retValue.money, "bladeburner");
       player.gainStats(retValue);
       // Operation Daedalus
-      const action = this.getActionObject(this.action);
       if (action == null) {
         throw new Error("Failed to get BlackOperation Object for: " + this.action.name);
       } else if (action.name === BlackOperationNames.OperationDaedalus && this.blackops[action.name]) {


phyzical:
would the action object update correctly? id assume it would hold the state from before the completion right?

borisflagell:
I'm fairly sure it does.
I was using console.log() to monitor variable content when testing.
const action retained the initial state even after the completeAction() call. Which is what we want it to do in this instance.

phyzical:
oh right its literally just used for the name, (any reason we cant just use `this.action.name`?

or is `this.action.name` front facing?

borisflagell:
Er... from what I gather : 
In theory, `this.action.name` cannot replace `getActionObject()` because some IAction implementation don't have a name property.

But in this instance, `processAction()` already handle that because of the following blob : 
```
if (!(this.action instanceof ActionIdentifier)) {
      throw new Error("Bladeburner.action is not an ActionIdentifier Object");
    }
```

So we could, in fact, use `this.action.name` in this instance. But, they're would not be lot of added-value to it anyway.

Because, in the end, `const action` is used to decide if we must re-run the action or not. Which is something that can only be done after we call `completeAction()`. 
So either we copy the action value to use it after `completeAction()`, either we evaluate it beforehand and store the result in a boolean for later. It doesn't make much difference from my perspective :-D

phyzical (approved):

PR #3702, opened 2022-05-22
BLADEBURNER: FIX #3508 Add GetActionCurrentTime() to Bitburner API

### LINKED ISSUE #3508: [Question/FeatureRequest] Bladeburner API: track progress
Is there a way to get the current progress on the bladeburner action? i'd like to make my script wait for the current action to complete once before checking for another action to run.

i.e.
```
const curr = ns.bladeburner.getCurrentAction();
const total = ns.bladeburner.getActionTime(curr.type, curr.name);

const timeSpent = ??;
const remaining = total  - timeSpent;

// best would be:
const [timeSpent , timeRemaining] = ns.bladeburner.getCurrentActionTime();  // or similar...
```

Side note: `getActionTime` docs are not sure what the method returns, text says seconds, `@returns` says milliseconds (cc @hydroflame). Checking the values, it seems to be millis.

borisflagell:
FIX #3508 - Add GetActionCurrentTime() to Bitburner API.
Slightly tweak BItburner.resetAction() to also reset time counters.
Fix GetActionTime documentation. 

I've some concerns : 

Not sure if it's the intended way to exploit Bladeburner API. Or if the player is expected to use ns.sleep in conjunction with getActionTime() instead.

Currently, the function return 0 when no Action is underway. Should it rather return NaN or throw an Error ?

### DIFF
--- a/src/ScriptEditor/NetscriptDefinitions.d.ts
+++ b/src/ScriptEditor/NetscriptDefinitions.d.ts
@@ -2897,7 +2897,7 @@ export interface Bladeburner {
    * @remarks
    * RAM cost: 4 GB
    *
-   * Returns the number of milliseconds it takes to complete the specified action
+   * Returns the number of milliseconds already spent on the current action.
    *
    * @returns Number of milliseconds already spent on the current action.
    */
--- a/src/ScriptEditor/NetscriptDefinitions.d.ts
+++ b/src/ScriptEditor/NetscriptDefinitions.d.ts
@@ -2884,13 +2884,24 @@ export interface Bladeburner {
    * @remarks
    * RAM cost: 4 GB
    *
-   * Returns the number of seconds it takes to complete the specified action
+   * Returns the number of milliseconds it takes to complete the specified action
    *
    * @param type - Type of action.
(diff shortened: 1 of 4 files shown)


phyzical:
i think 0 is fine, as the users code would probably do whatever, aslong as they check they are trying to do an action they shouldn't shoot themselves in the foot.

though i wonder what would be better time remaining or time spent?

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -2884,14 +2884,25 @@ export interface Bladeburner {
...
    * @returns Number of milliseconds it takes to complete the specified action.
    */
   getActionTime(type: string, name: string): number;
 
+  /**
+   * Get the time elapsed on current action.
+   * @remarks
+   * RAM cost: 4 GB
+   *
+   * Returns the number of milliseconds it takes to complete the specified action
```
phyzical, line 2900: Number of milliseconds already spent on the current action.

borisflagell: done.

PR #3728, opened 2022-05-23
UI: Apply the Typography tag to the share purchase cost text when buying back shares.

MonkeyManKing:
UI: Apply the Typography tag to the share purchase cost text when buying back shares. Otherwise text is incredibly dark on dark theme


borisflagell (approved):
You might want to use a variant of Typography to help separate the text : 

default typo : 
image


with `<Typography variant="caption">`
image

borisflagell (approved):
Seems ready to ship now 👍

PR #3729, opened 2022-05-23
API: Fix #2700 alter/add API function to get unmodified augmentation cost.

### LINKED ISSUE #2700: Feature: Add function to get base price of augment rather than current price(including the multiplier) or a function to sleeve that returns the price of their augments

borisflagell:
fixes #2700

Tweaked existing function : ( by adding a returnBaseCost boolean parameter. )
-    singularity.GetAugmentationPrice( )
-    singularity.GetAugmentationRepReq( )
 
Added new functions :
-    sleeve.GetSleeveAugmentationPrice( )
-    sleeve.GetSleeveAugmentationRepReq( )

image

### DIFF
--- a/src/NetscriptFunctions/Singularity.ts
+++ b/src/NetscriptFunctions/Singularity.ts
@@ -134,18 +134,20 @@ export function NetscriptSingularity(player: IPlayer, workerScript: WorkerScript
         return aug.prereqs.slice();
       },
     getAugmentationPrice: (_ctx: NetscriptContext) =>
-      function (_augName: unknown): number {
+      function (_augName: unknown, _returnBasePrice: unknown = false): number {
         _ctx.helper.checkSingularityAccess();
         const augName = _ctx.helper.string("augName", _augName);
+        const returnBasePrice = _ctx.helper.boolean(_returnBasePrice);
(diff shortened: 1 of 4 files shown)


src/NetscriptFunctions/Singularity.ts
```diff
@@ -134,18 +134,20 @@ export function NetscriptSingularity(player: IPlayer, workerScript: WorkerScript
         return aug.prereqs.slice();
       },
     getAugmentationPrice: (_ctx: NetscriptContext) =>
-      function (_augName: unknown): number {
+      function (_augName: unknown, _returnBasePrice: unknown = false): number {
```
phyzical, line 137: yeah i still dont think we need the base price on this function imo, like will it ever be useful?

or is there a usecase im missing?

hydroflame:
The sleeve functions were added manually and the base price functions were added in another patch

PR #3741, opened 2022-05-24
MISC: Convert Create program to current work

hydroflame:


src/NetscriptFunctions/Singularity.ts
```diff
@@ -526,8 +527,10 @@ export function NetscriptSingularity(player: IPlayer, workerScript: WorkerScript
 
         player.getHomeComputer().pushProgram(item.program);
         // Cancel if the program is in progress of writing
-        if (player.createProgramName === item.program) {
+        const w = player.currentWork;
```
phyzical, line 530:
```suggestion
        const work = player.currentWork;
```

src/PersonObjects/Work/Work.ts
```diff
@@ -0,0 +1,21 @@
+import { IPlayer } from "../IPlayer";
+
+export enum WorkType {
+  CreateProgram = 0,
```
phyzical, line 4: shouldn't we keep it as string enums, more reusability for user facing stuff? or reason to move to ints?

src/engine.tsx
```diff
@@ -291,15 +291,12 @@ const Engine: {
       Player.gainMoney(offlineHackingIncome, "hacking");
       // Process offline progress
       loadAllRunningScripts(Player); // This also takes care of offline production for those scripts
+      Player.focus = true;
```
phyzical, line 294: hmm wont this cause some unintended side effects? or is it set back to false all over the joint?

phyzical (approved):
looking good so far,

i think i understand why we so  / by the idle speed here as we multiple it by it on the other side,

But whats the motivation for the the doubling of exp OOC?

PR #3745, opened 2022-05-24
UI: Proof of concept - Make a Grid Component responsive.

borisflagell:
Improve the hacknet page behaviour on low-width context/browser/terminal.
That kind of comportement should be declinable in most situation.

chrome-capture-2022-4-24

image


borisflagell:
Github setup went FUBAR.

PR #3759, opened 2022-05-25
MISC Two new encryption themed contracts - caesar and vigenere

Markus-D-M:
Two new contracts:
- Encryption I: Caesar Cipher
- Encryption II: Vigenère Cipher

New pull, because i messed up the old one - sorry for that.


stalefishies:
I would specify how the data is given to you via the API: at the moment the description doesn't say it's giving you an array. Algorithmic Stock Trader IV is a good one to copy to match the style of the current contracts.

I'd also add the description to the table at the bottom of https://github.com/danielyxie/bitburner/blob/dev/doc/source/basicgameplay/codingcontracts.rst

Also, maybe Caesar should give you a shift in the range [1, 25] instead of [0, 25] to avoid the trivial case?

Cool contracts though!

xRalic:
It was fun coding the solutions for these. Definitely a welcome addition, and hopefully it'll get people interested in learning about encryption. 😃

xRalic (approved):

PR #3763, opened 2022-05-25
API: FIX#3630 Reduce Corporation API Ram usage at 64Go (was 1To)

### LINKED ISSUE #3630: Suggestion: Make the corporation API cost 0 ram in the Corporatocracy BitNode
This game rocks, I've sunk huge hours and it always comes back offering more. Every BitNode offers something new and exciting. The first time I noticed myself discouraged with continuing was in the Corporatocracy node (my 5th node type). The issue that caused discouragement was because I hit a wall of progression waiting for my wealth to accumulate to where I could start using the ns.corporation APIs by purchasing home ram over 1TB.

I know that manual manipulation via the UI could get me there faster - but that's the real shame, I try and interact with the UI purely to learn mechanics. The moment I have an understanding I want to exclusively work in scripts for all aspects of the game (and the game really really excels at letting me do this - except for this spot).

There is a similar issue in BitNode 6 + 7 where the API to code against BitNode 6's mechanic is in 7...which I don't like... However, it can be worked around. This one is really punishing if you don't manually setup a corp. It adds something like 72 hours per BitNode reset if you do not use the manual interface.

The huge ram requirement for corps makes sense to me, just not in context of the Corporatocracy BitNode since this is supposed to be an "easy" rated node, either set to 0 or set to a lower number like 32 or 64.

borisflagell:
closes #3630.

Reduce Corporation API Ram usage at 64Go (was 1To).

I brought the linked issue on discord once. It was met with some assentiment : 1To is a lot, and the reason behind it is weird. (apparently, it was some form of retaliation from a dev against ram usage scumming)

https://discord.com/channels/415207508303544321/415213435974975508/977879935131729931


phyzical:
imo instead we convert it to work like other apis and actually force ram usage per function rather than a flat fee

borisflagell:
> imo instead we convert it to work like other apis and actually force ram usage per function rather than a flat fee, this change is a little more complicated but i can take a crack/ help you out if your up for it

I've just pushed a commit to enforce your suggestion.

I might need an helping hand with the Jest test error now :/

stalefishies:
TBH I actually really like the APIs having a flat cost, especially with the corporation API being as...bitty and sprawling as it is. Charging RAM for something like getUnlockUpgradeCost just rewards people for grabbing raw numbers from the source code. There's also functions like getDivision, which is a helper function that just grabs a division from the object returned from getCorporation. Charging a flat fee doesn't punish players for using the nice-to-have functions - which is surely something we'd want to encourage.

Reducing the RAM cost is also effectively a buff to corps which it most certainly doesn't need. This is doubly true with separated RAM cost, since you can now avoid certain functions and save more RAM.

IMO the RAM cost for corps _should_ feel somewhat high. It gives the API a unique niche - high cost, but an automatic money printer is high reward - though perhaps it's fair enough to come down from 1TB. My gut says something like 128GB as a flat fee, but idk.

phyzical:
@stalefishies what if we were to make things like info cheap as chips and aim for the total cost of an avg script to be around 128 gb?

as for sniffing the source for raw numbers id say thats kind of encouraged tbh ;) i know i did it from the start to get all the enum values ect, goes along the whole if you don't understand something check the source good teaching moments

stalefishies:
@phyzical IMO there's a difference between encouraging people to read the source code to work out how the mechanics work (which I agree we should absolutely encourage) and then them just copy-pasting data from the source to hardcode it into a script.

For example, I know the costs for the corp upgrades are simple exponentials, and I learned that from reading the source. So I could easily calculate the costs of each level by myself, but I'd still rather get them from the API in case the costs get rebalanced or something.

But on the actual topic of the RAM costs: if our goal becomes to make the average cost 128GB, then why not just make it a flat 128GB cost? It's simpler to use and simpler to implement.

IDK what the real benefit is from making functions have different RAM costs. It's not like NS or Singularity where scripts are going to pick and choose which bits of the API they want to use - corp scripts are going to be pretty all or nothing about which API functions they use. So I don't see the benefit - I don't see there being a meaningful or interesting gameplay consequence from separating RAM costs. In that case, I'd argue the simplicity of a flat cost wins.

I'd like to hear other people's takes on this, since I'm obviously primarily basing that on how I play and think about the game.

borisflagell:
> But on the actual topic of the RAM costs: if our goal becomes to make the average cost 128GB, then why not just make it a flat 128GB cost? It's simpler to use and simpler to implement.

It's highly dependent of Player's coding style.
If you use a few big script to handle your whole Corporation usage, flat cost for namespace is best for you.
If you use a lot of small script, per-function cost basis will be best.

It's one those situation where you cannot please everyone.
Lowering Ram usage should from 1To -> 64Go. Should be well-received by most people.
Switching from per-namespace -> to per-function is more of a skub though.

I'm open to reverting it.

phyzical:
@stalefishies imo the game is funner (and encourages smarter scripts) when the user has to logically choose some parts to be run before they have enough for the full suite. but does allow for you to automate things sooner.

but again thats my opinion happy to just go with a flat fee if others prefer.

@borisflagell maybe pass by a few others in discord see what they think?

hydroflame:
With the API being all sync this pain point should be mostly eliminated.

PR #3776, opened 2022-05-27
INFIL: Fix getPossibleLocations

### LINKED ISSUE #3804: method ns.infiltration.getPossibleLocations() returns all names as "[object Object]"
It's currently not possible to load a list of infiltration locations. Method `ns.infiltration.getPossibleLocations()` returns a bunch of useless strings.

```
/** @param {NS} ns */
	ns.tprint(ns.infiltration.getPossibleLocations());
}
```
- [ ] Version of the game: v1.7.0 (00adc2ff)

- [ ] Possible reason:
I think the problem might be in [this line](https://github.com/danielyxie/bitburner/blob/69781359d5d48764419646754927e72581c81dbc/src/NetscriptFunctions/Infiltration.ts#L45)

phyzical:
* return city and location name instead of whatever its trying todo
closes #3804

### DIFF
--- a/src/NetscriptFunctions/Infiltration.ts
+++ b/src/NetscriptFunctions/Infiltration.ts
@@ -41,8 +41,8 @@ export function NetscriptInfiltration(player: IPlayer): InternalAPI<IInfiltratio
     };
   };
   return {
-    getPossibleLocations: () => (): string[] => {
-      return getLocationsWithInfiltrations.map((l) => l + "");
+    getPossibleLocations: () => (): any[] => {
+      return getLocationsWithInfiltrations.map((l) => ({ city: l.city, name: l.name }));
     },
     getInfiltration:
       (ctx: NetscriptContext) =>
--- a/src/NetscriptFunctions/Infiltration.ts
+++ b/src/NetscriptFunctions/Infiltration.ts
@@ -1,7 +1,7 @@
 import { IPlayer } from "../PersonObjects/IPlayer";
 
 import { Infiltration as IInfiltration, InfiltrationLocation } from "../ScriptEditor/NetscriptDefinitions";
-import { Location } from "../Locations/Location";
+import { Location, SimpleLocation } from "../Locations/Location";
 import { Locations } from "../Locations/Locations";
 import { calculateDifficulty, calculateReward } from "../Infiltration/formulas/game";
 import {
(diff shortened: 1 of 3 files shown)


borisflagell (approved):

src/NetscriptFunctions/Infiltration.ts
```diff
@@ -41,8 +41,8 @@ export function NetscriptInfiltration(player: IPlayer): InternalAPI<IInfiltratio
     };
   };
   return {
-    getPossibleLocations: () => (): string[] => {
-      return getLocationsWithInfiltrations.map((l) => l + "");
+    getPossibleLocations: () => (): any[] => {
+      return getLocationsWithInfiltrations.map((l) => ({ city: l.city, name: l.name }));
```
thefotios, line 45: (Long time player; first time commenter :wink:  )

I've been trying out this branch, and this method signature doesn't really help much in the editor (or VSCode when loading `bitburner.d.ts`).

At the very least could this (and other references) be changed to
```suggestion
    getPossibleLocations: () => (): {city: string, name: string}[] => {
      return getLocationsWithInfiltrations.map((l) => ({ city: l.city, name: l.name }));
```

Ideally, it could be `{city: Cityname | null, name: Location}`, but that would require exporting those `enum` values and I'm not sure what this project's stance is on that (it looks like a few stock related ones are exported, but no others)

phyzical: hmm i think i tried this and could not or maybe it was lost in a merge in the past ill try again.

phyzical:
ah right so the issue was that you cant import interfaces into the Netscript definitions.

But what ill do (cant promise it will make it in using this approach ) is just repeat the same interface to get it working for intellisense 

image
image

phyzical:
@thefotios if you get a chance mind giving it a whirl again and let me know if it does what your after

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -4325,11 +4325,20 @@ export interface InfiltrationReward {
  * @public
  */
 export interface InfiltrationLocation {
-  location: any;
+  location: SimpleLocation;
```
thefotios, line 4328: This looks like it should be `Location` instead based on the return signature from https://github.com/danielyxie/bitburner/blob/e85a66202a74f7bac458d4bb4438ad246f6aa017/src/NetscriptFunctions/Infiltration.ts#L19-L38
```suggestion
  location: Location;
```

thefotios (commented):
@phyzical The `SimpleLocation` stuff works great! I left 2 other comments that you might want to look into

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -4325,11 +4325,20 @@ export interface InfiltrationReward {
...
 }
 
+/**
+ * Infiltration API.
+ * @public
+ */
+interface SimpleLocation {
+  city: CityName | null;
+  name: LocationName;
+}
```
thefotios, line 4340: I'm not sure what the best practice is here. These need to be imported in order to be resolved [^1], but they're `enum` with string values, so do they belong in a `.d.ts` file?

[^1]: As does `Location` from my suggestion above

phyzical:
closed in favour of the accepted fix

PR #3787, opened 2022-05-29
BLADEBURNER: Added bulk upgrades to upgradeSkill NS API

Undeemiss:
This PR introduces a new optional parameter to `ns.bladeburner.upgradeSkill` and `ns.bladeburner.getSkillUpgradeCost` to allow multiple simultaneous upgrades.

BALANCE: Substitutes the formula for BB skill costs with a more-efficient version that deviates slightly from the existing formula in some cases. This may be undesirable.

A player earning quadrillions of BB skill points per operation was unable to spend them as fast as they were being earned due to IRL computer speed bottlenecks, so I attempted to implement bulk upgrading to alleviate this. Bulk upgrades use an optimized cost formula, collapsing a large summation at the cost of some computational accuracy. As such, small numbers still use the old formula, but large numbers use the optimized version, meaning the lost accuracy only comes into play when its effect on overall cost is negligible.
It is worth noting that the current implementation of the optimized formula is designed with accuracy for non-integers in mind, and reliably underestimates the cost by 0.5 skill points per level for upgrades where `skill.costInc * BitNodeMultipliers.BladeburnerSkillCost` is an integer.

Credit to my good friend [Es-tactic](https://lichess.org/@/Es-tactic) for helping me make the optimized formula more accurate when `skill.costInc * BitNodeMultipliers.BladeburnerSkillCost` is not an integer. (Yes that link goes to his page on a chess site lol)

The functions have been tested with many values of count, such as -Infinity, Infinity, NaN, 1.5, 1, 99, 100, 1e153, and -32980.2932920. The optimized cost function was tested in Desmos to ensure that for large count, the output values should be effectively indistinguishable from the recursive mode, but it is possible that floating-point errors or the like may render the values inaccurate. I tested to be sure that all of the outputs I got looked reasonable, but by no means did I use a calculator to calculate the actual effective error of every input value I tried ingame.

### DIFF
--- a/src/Bladeburner/Bladeburner.tsx
+++ b/src/Bladeburner/Bladeburner.tsx
@@ -2276,7 +2276,7 @@ export class Bladeburner implements IBladeburner {
     }
   }
 
-  getSkillUpgradeCostNetscriptFn(skillName: string, workerScript: WorkerScript): number {
+  getSkillUpgradeCostNetscriptFn(skillName: string, count: number, workerScript: WorkerScript): number {
     if (skillName === "" || !Skills.hasOwnProperty(skillName)) {
       workerScript.log("bladeburner.getSkillUpgradeCost", () => `Invalid skill: '${skillName}'`);
       return -1;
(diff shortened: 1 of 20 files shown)


Undeemiss:
It would appear one of my commits accidentally formatted files unrelated to the changes I was working on. Sorry about that.

PR #3798, opened 2022-05-30
MISC: Fix lint errors and reformat

Undeemiss:
Manually removed unused imports to fix lint errors, then reran `npm run format; npm run lint` in terminal. This should fix all the errors on dev.


Undeemiss:
I see this failed some checks; I have no idea what those are, given that when I run lint myself it works perfectly well with no errors or warnings.

Undeemiss:
I believe I found the problem; I'm going to make a new PR instead of trying to fix this one.

PR #3799, opened 2022-05-30
MISC: Fix formatting, lint errors, and a broken test case on dev

Undeemiss:
This PR fixes all of the failed PR tests that have been plaguing recent changes. Most notably, the test intended to check that the stock market was properly updating was broken and impossible to pass. I have fixed this bug, although I'm not convinced my fix is the cleanest solution available.


hydroflame:
Changes were added manually

PR #3800, opened 2022-05-30
UI: Fixed an issue causing BN12 stats to show incorrectly

Undeemiss:
Fixed an off-by-one error causing the stats screen to display BN mults for BN12.(n+1) when the player was actually in BN12.n


hydroflame:
Doesn't look to be an issue, I have SF12.1, I am on BN12.2
Screen Shot 2022-07-21 at 1 55 22 AM

It correctly displays 1.02^2

PR #3802, opened 2022-05-30
CORPORATION: (BREAKING) Dividend fixes and exposing dividends info via scripts

stalefishies:
Makes the corporation internals consistent in using a dividend rate (between 0 and 1) instead of mostly percentages (between 0 and 100), except sometimes it was between 0 and 1 but called a percentage. Renames `corporation.dividendPercentage` to `corporation.dividendRate` accordingly.

Replaces the unused `corporation.dividendTaxPercentage` with an actually-used `coropration.dividendTax`.

Exposes `dividendRate`, `dividendTax`, and `dividendEarnings` (which is shareholder earnings per second) to scripts via `ns.corporation.getCorporation`.

Fixes documentation for `ns.corporation.issueDividends` to clarify it expects a value between 0 and 1. Also fixes the error checking for that function which checks if it's between 0 and 100, and then passes it to a function which checks if it's between 0 and 1.

Fixes the value of the player's shareholder earnings on the corporation overview page which was a factor of 10 out; the value was per-cycle but was displayed as if it were per-second.

### DIFF
--- a/src/Corporation/Corporation.tsx
+++ b/src/Corporation/Corporation.tsx
@@ -35,8 +35,8 @@ export class Corporation {
   shareSalesUntilPriceUpdate = CorporationConstants.SHARESPERPRICEUPDATE;
   shareSaleCooldown = 0; // Game cycles until player can sell shares again
   issueNewSharesCooldown = 0; // Game cycles until player can issue shares again
-  dividendPercentage = 0;
-  dividendTaxPercentage = 50;
+  dividendRate = 0;
+  dividendTaxPercentage = 100 - 100 * BitNodeMultipliers.CorporationSoftcap;
   issuedShares = 0;
   sharePrice = 0;
   storedCycles = 0;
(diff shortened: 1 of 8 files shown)


src/Corporation/Corporation.tsx
```diff
@@ -277,10 +281,8 @@ export class Corporation {
     this.funds = this.funds - price;
 
     // Apply effects for one-time upgrades
-    if (upgN === 5) {
-      this.dividendTaxPercentage -= 5;
-    } else if (upgN === 6) {
-      this.dividendTaxPercentage -= 10;
+    if (upgN === 5 || upgN === 6) {
```
phyzical, line 284: nit: might not hurt to just always call it as the default seems safe 

otherwise might aswell not have the default case within updateDividentTax

src/Corporation/data/Constants.ts
```diff
@@ -61,7 +61,7 @@ export const CorporationConstants: {
 
   ProductProductionCostRatio: 5, //Ratio of material cost of a product to its production cost
 
-  DividendMaxPercentage: 1,
+  DividendMaxRate: 1,
```
phyzical, line 64: would this default case still be 1? not 0.01?

src/Corporation/ui/modals/IssueDividendsModal.tsx
```diff
@@ -19,7 +19,7 @@ export function IssueDividendsModal(props: IProps): React.ReactElement {
   const corp = useCorporation();
   const [percent, setPercent] = useState(0);
 
-  const canIssue = !isNaN(percent) && percent >= 0 && percent <= CorporationConstants.DividendMaxPercentage * 100;
+  const canIssue = !isNaN(percent) && percent >= 0 && percent <= CorporationConstants.DividendMaxRate * 100;
```
phyzical, line 22: (i think)
```suggestion
  const canIssue = !isNaN(percent) && percent >= 0 && percent <= CorporationConstants.DividendMaxRate;
```

phyzical: and obviously update to be rate not % or did you actualy want % here which is weird why it was being *100 to begin with

phyzical:
changes seem reasonable few nits/comments but id also mark this as a breaking change as its moving the api entry point from a % to a rate (50 => 0.5)

src/Corporation/data/Constants.ts
stalefishies: Nope, this is a case where the property was named percentage but was actually used by multiplying by 100 (e.g. https://github.com/danielyxie/bitburner/blob/4993a024211714c5af9a79fd73f8d9a37ed4025f/src/Corporation/Corporation.tsx#L129)

stalefishies:
I don't see the API break: both the script API in `ns.corporation.issueDividends` and the internal API in `IssueDividends` (in Actions.ts) both already took rates (0 - 1), even if their arguments / documentation said 'percent'. So this isn't a breaking change; instead it refits the internal corp code to match the existing API.

phyzical (approved):
while that may be true its still breaking in the sense that users scripts will not work as they did before but its just more of a update the pr title to mark as breaking change so users are more likely to see the change rather than just claiming its a new bug :)

PR #3841, opened 2022-06-13
BUGFIX: Fix #3803 Servers can no longer have duplicate IPs

### LINKED ISSUE #3803: defcomm and iron-gym share an ip address
EDIT: I've since learned that ips change on reset. But that would mean that during a reset the game is assigning two different servers the same ip address. Must have been a slim chance, but it happened, can happen.

image

Bitburner v1.7.0 (00adc2ff)
- [ ] I don't upload save files.
- [x] Script: 
```
/** @param {NS} ns */
	const irongym = ns.getServer("iron-gym");
	const ip = ns.getServer("61.7.8.3");
	const defcomm = ns.getServer("defcomm");
	console.log(irongym);
	console.log(ip);
	console.log(defcomm);
	ns.tprint(irongym);
	ns.tprint(ip);
	ns.tprint(defcomm);
}
```
- [x] Run this script and you will see that the ips conflict. I can't share the entire framework that discovered the bug, but this script will at least show you the results.
- [x] Bitburner v1.7.0 (00adc2ff)

crimsonhawk47:
# Issue 

There was a possibility that two servers could share an IP address. 

# Fix

The AllServers object was indexed by IP instead of hostname, and so it always returned `null`. This loops through the servers to see if any of the IPs match. 

# Linked issues

Closes #3803 

# Testing

I made `createRandomIp` return the same IP address every time, and I tested the game before and after. 

Before applying the fix, the game ran fine. Every server had the same IP, and you could buy servers fine. 

After applying the fix, the game would be in an infinite loop on launch, which is expected since `createUniqueRandomIp` will loop infinitely when the same IP is returned every time.

### DIFF
--- a/src/Server/AllServers.ts
+++ b/src/Server/AllServers.ts
@@ -79,10 +79,10 @@ export function DeleteServer(serverkey: string): void {
 export function ipExists(ip: string): boolean {
   for (const hostName in AllServers) {
     if (AllServers[hostName].ip === ip) {
-      return true
+      return true;
     }
   }
-  return false
+  return false;
 }
 
 export function createUniqueRandomIp(): string {
--- a/src/Server/AllServers.ts
+++ b/src/Server/AllServers.ts
@@ -77,7 +77,12 @@ export function DeleteServer(serverkey: string): void {
 }
 
 export function ipExists(ip: string): boolean {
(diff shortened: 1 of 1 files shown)


phyzical (approved):
Ah was wondering what caused this

PR #3847, opened 2022-06-16
COMPANY: FIX #3837, #3830 corrected backwards logic in quitJob function

Snarling:
Fixes #3830
Fixes #3837 (or at least seems to)

If you're currently doing company work, and quitting from the same company that is active, then it will finish your work shift before deleting the job. Previously, it would only attempt to finish the work shift if you _weren't_ doing company work, but were quitting from your active company.

Edit: Fixes #3851, Fixes #3855, Fixes #3826, Fixes #3809, Fixes #3769, Fixes #3857, Fixes #3856, Fixes #3877, Fixes #3878

PR #3874, opened 2022-06-29
JOB: FIX #3865 Finish the job before quitting a company

### LINKED ISSUE #3865: TypeError: t is undefined (at "Work")
### How did this happen?
To do this bug:
1. You have to be in a job (ingame).
2. Then work.
3. Press "Do something else simultaneously".
4. Quit the job (ingame)
5. Press the Focus button.

This will result in Recovery mode with this bug, or would i say, a TypeError.

Please fill this information with details if relevant.

- [ ] Save file
- [X] Steps to reproduce

### Environment

* Error: TypeError: t is undefined
* Page: Work
* Version: v1.7.0 (b60f3da8)
* Environment: Production
* Platform: Browser
* UserAgent: Mozilla/5.0 (Windows NT 6.1; Win64; x64; rv:101.0) Gecko/20100101 Firefox/101.0
* Features: lang=en cookiesEnabled=true doNotTrack=unspecified indexedDb=true
* Source: https://danielyxie.github.io/bitburner/dist/main.bundle.js

### Explanation
What it might be, is that the t variable is stored in shifts. And if you aren't in a shift let alone not being in a job (ingame), so it doesn't have access to the t variable, therefore causing a fatal error, which is only fatal if you don't have a save file before the triggered bug.

I found this out, on my own, while I was trying to find a job that has a lot per second, I forgot to press the "Stop working" button, and quitted out of the job (ingame), so I tried to see if it would still work even if i quitted it, it didn't.

ilkecan:
Fixes: #3865

### DIFF
--- a/src/PersonObjects/Player/PlayerObjectGeneralMethods.tsx
+++ b/src/PersonObjects/Player/PlayerObjectGeneralMethods.tsx
@@ -1854,7 +1854,11 @@ export function getNextCompanyPosition(
 }
 
 export function quitJob(this: IPlayer, company: string, _sing = false): void {
-  if (this.isWorking == true && this.workType !== WorkType.Company && this.companyName == company) {
+  if (
+    this.isWorking == true &&
+    [WorkType.Company, WorkType.CompanyPartTime].includes(this.workType) &&
+    this.companyName == company
+  ) {
     this.finishWork(true);
   }
   delete this.jobs[company];


phyzical:
I think this is already covered by https://github.com/danielyxie/bitburner/pull/3847

ilkecan:
Yeah, I didn't see it.

PR #3881, opened 2022-07-03
UI: Fix #3854 ctrl+c does not clear terminal input 

### LINKED ISSUE #3854: ctrl+c does not clear terminal input
- [ ] Steps to reproduce: have "enable bash hotkeys" set (or not set, doesnt matter) in options -> misc, try to use `ctrl+c` to clear terminal input. it does nothing, regardless of either hotkey option (disable hotkeys or enable bash hotkeys) setting.
- [ ] Version of the game: Bitburner v1.7.0 (00adc2ff)

its mentioned [in the docs](https://bitburner.readthedocs.io/en/latest/shortcuts.html#terminal-bash-shortcuts) to `nano .fconf` to enable the hotkeys, but that behavior is no longer supported. and i can't figure out how to allow ctrl+c to override the global hotkey since the disable hotkeys toggle doesnt allow it

evil-tim:
Fixes #3854

Implement ctrl+c bash hotkey. Clears current terminal input. If there is selected text, the hotkey is not triggered to allow copying.

1. Check if ctrl+c clears terminal input when no text is selected in terminal input
2. Check if ctrl+c copies text when text is selected in terminal input

bb-3854-s

### DIFF
--- a/src/Terminal/ui/TerminalInput.tsx
+++ b/src/Terminal/ui/TerminalInput.tsx
@@ -324,6 +324,7 @@ export function TerminalInput({ terminal, router, player }: IProps): React.React
     if (Settings.EnableBashHotkeys) {
       if (event.code === KEYCODE.C && event.ctrlKey && ref && ref.selectionStart === ref.selectionEnd) {
         event.preventDefault();
+        terminal.print(`[${player.getCurrentServer().hostname} ~${terminal.cwd()}]> ${value}`);
         modifyInput("clearall");
       }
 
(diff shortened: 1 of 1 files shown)


Hoekstraa (approved):

phyzical (approved):

evil-tim:
Update ctrl+c to print content before clearing to mimic real terminal behavior

1. Check if ctrl+c clears text from the input and prints that text to the terminal without executing it or adding it to history 
2. Check if ctrl+c copies selected text when text is selected in terminal input

bb-3854-s-3

GottZ:
how about also adding ctrl + u for that. (which on proper shells will clear the current line)

PR #3946, opened 2022-07-31
Remove "based" from positive adjectives in infil

faangbait:
This slang term has had three different meanings in my lifetime alone. It's far too regional and cohort-sensitive to be universally understood to be a positive word.

Note that I don't have the same objections to "cringe," introduced in the same commit, as it has obviously negative connotations that don't require understanding the slang version.


phyzical (approved):
meh

Aerophia:
i suggest replacing it with "poggers" and "redpilled"

hydroflame:
Actually cringe should be removed as well tbh

PR #3954, opened 2022-08-02
CONTRACTS: FIX #3755 change input handling for contract attempts

### LINKED ISSUE #3755: Coding contract "Proper 2-Coloring of a Graph" does not accept empty array as solution, insists on a literal string "[]"
When my contract solver returned an empty array as solution for a 2-colorung contract that didn't have a proper coloring, it failed. (My
debug output showed, that it supplied as solution the empty array.)

I then checked that there indeed was no solution (graphs contained [2,7], [2,5] and [5,7]), the answer was correct.

Then I tried it in the UI, entering nothing in the solution field (this usually also denotes the "empty" solution in other contracts), this didn't work, too. Finally I entered in the UI the string "[]" and the answer was accepted.

My expectation was that the empty array as a solution from a script should be accepted as correct answer.

Unfortunately I don't have a save file with that contract alive.

- [x] Version of the game, e.g. Bitburner v1.7.0 (837cceba)

Snarling:
Fixes #3755 

* For 2-coloring of a graph, moved sanitization of player input to before the "empty" check, so both `""` and `"[]"` strings will be accepted when there is no solution.
* Standardized handling of all inputs by the player. Previously, 2d arrays were handled differently than all other submitted answers, because `String(answer)` would destroy the 2d structure. Using `JSON.stringify` makes any array input match the way a player would have submitted the answer using a string.


src/NetscriptFunctions/CodingContract.ts
```diff
@@ -34,20 +33,11 @@ export function NetscriptCodingContract(player: IPlayer, workerScript: WorkerScr
...
-          const answerComponents = [];
-          for (let i = 0; i < answer.length; ++i) {
-            answerComponents.push(["[", String(answer[i]), "]"].join(""));
-          }
-
-          answerStr = answerComponents.join(",");
-        } else {
-          answerStr = String(answer);
-        }
+        if (typeof answer !== "number" && typeof answer !== "string" && !Array.isArray(answer))
```
Snarling, line 36: typechecking to actually enforce the input types allowed in documentation, and provide a helpful error message.

src/NetscriptFunctions/CodingContract.ts
```diff
@@ -34,20 +33,11 @@ export function NetscriptCodingContract(player: IPlayer, workerScript: WorkerScr
...
+          throw new Error("The answer provided was not a number, string, or array");
 
+        // Convert answer to string.
+        const answerStr = JSON.stringify(answer);
```
Snarling, line 40: JSON.stringify will always return a string for the types that are still allowed for answer, no need to throw another error for if JSON.stringify results in undefined.

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -3222,7 +3222,12 @@ export interface CodingContract {
    * @param opts - Optional parameters for configuring function behavior.
    * @returns True if the solution was correct, false otherwise. If the returnReward option is configured, then the function will instead return a string. If the contract is successfully solved, the string will contain a description of the contract’s reward. Otherwise, it will be an empty string.
    */
-  attempt(answer: string[] | number, filename: string, host?: string, opts?: CodingAttemptOptions): boolean | string;
+  attempt(
```
Snarling, line 3225: Previous documentation was just wrong.
`string` is always an accepted type of input. Other types that would be expected are `number`, `number[]`, `number[][]` `string[]` but that would look ridiculous so any[] covers all the array types.

src/data/codingcontracttypes.ts
```diff
@@ -1422,9 +1422,13 @@ export const codingContractTypesMetadata: ICodingContractTypeMetadata[] = [
...
     solver: (_data: unknown, ans: string): boolean => {
       const data = _data as [number, [number, number][]];
+
+      //Sanitize player input
+      const sanitizedPlayerAns: string = removeBracketsFromArrayString(ans);
+
       //Case where the player believes there is no solution.
       //Attempt to construct one to check if this is correct.
-      if (ans == "[]") {
+      if (sanitizedPlayerAns === "") {
```
Snarling, line 1431: Both `"[]"` and `""` string inputs will be accepted, and both are possible string representations of empty array.

PR #3960, opened 2022-08-06
NETSCRIPT: FIX #3958 Remove old wrapper, fix new wrapper (v2)

### LINKED ISSUE #3958: v2: most functions not checking for env.stopFlag
The only place the stopFlag is being checked is in [netscriptDelay](https://github.com/danielyxie/bitburner/blob/39cf0cb57f5c0c2515ab5352c73e1cdded137882/src/NetscriptEvaluator.ts#L17), so all synchronous ns functions + asleep are currently not checking the stopFlag in v2. They will just continue running forever.

- [x] Minimal scripts to reproduce the issue:
```
export let main=async ns=>{
  ns.tail();
  ns.exit();
  for(;;){
    ns.tprint("I'm still alive.")
    ns.tprint(`INFO\nProcess list: ${ns.ps("home")}\nRam usage: ${ns.getServerUsedRam("home")}`);
    await ns.asleep(3000);
  }
}
```

- [x] Steps to reproduce
Run the provided script. The only way to end script execution is to reload the page.

- [x] Version of the game: Bitburner v2.0.0 (8e859d84)

Snarling:
Fixes #3958 . The previous `wrap`ped functions from NetscriptWorker.ts were not being used in v2. Since there is a new wrapper in APIWrapper.ts, I removed the wrapper from NetscriptWorker.ts. The erroring functionality that previously only existed in the NetscriptWorker.ts wrapper (error on dead script/concurrent functions other than asleep) has been moved to the new wrapper.

`netscriptDelay` is now the function that sets and clears `runningFn`, and `runningFn` is now part of `workerScript.env`. `netscriptDelay` now requires a function name to be passed to it so that it can set runningFn.

### DIFF
--- a/src/Netscript/APIWrapper.ts
+++ b/src/Netscript/APIWrapper.ts
@@ -13,6 +13,8 @@ import { INetscriptHelper, ScriptIdentifier } from "../NetscriptFunctions/INetsc
 import { GangMember } from "../Gang/GangMember";
 import { GangMemberTask } from "../Gang/GangMemberTask";
 import { ScriptArg } from "./ScriptArg";
+import { ScriptDeath } from "./ScriptDeath";
+import { sprintf } from "sprintf-js";
 
 type ExternalFunction = (...args: unknown[]) => unknown;
 export type ExternalAPI = {
@@ -63,6 +65,22 @@ type WrappedNetscriptHelpers = {
   gangTask: (t: unknown) => GangMemberTask;
 };
 
+function checkEnv(ctx: NetscriptContext) {
+  if (ctx.workerScript.env.stopFlag) throw new ScriptDeath(ctx.workerScript);
+  if (ctx.workerScript.env.runningFn && ctx.function !== "asleep") {
+    if (ctx.workerScript.delayReject) ctx.workerScript.delayReject();
+    const msg =
+      "Concurrent calls to Netscript functions are not allowed!\n" +
+      "Did you forget to await hack(), grow(), or some other\n" +
+      "promise-returning function? (Currently running: %s tried to run: %s)";
+    ctx.workerScript.errorMessage = makeRuntimeRejectMsg(
+      ctx.workerScript,
+      sprintf(msg, ctx.workerScript.env.runningFn, ctx.function),
+    );
+    throw new ScriptDeath(ctx.workerScript);
+  }
+}
+
 function wrapFunction(
   helpers: INetscriptHelper,
   wrappedAPI: ExternalAPI,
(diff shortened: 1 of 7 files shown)


Snarling:
@TheMas3212 Mind taking a look at this?

src/Netscript/APIWrapper.ts
```diff
@@ -107,10 +125,11 @@ function wrapFunction(
     },
   };
   function wrappedFunction(...args: unknown[]): unknown {
+    checkEnv(ctx);
     helpers.updateDynamicRam(ctx.function, getRamCost(Player, ...tree, ctx.function));
     return func(ctx)(...args);
   }
-  const parent = getNestedProperty(wrappedAPI, ...tree);
+  const parent = getNestedProperty(wrappedAPI, tree);
```
TheMas3212, line 132: this should still have the argument spread i think (...tree)

src/Netscript/APIWrapper.ts
```diff
@@ -129,7 +148,7 @@ export function wrapAPI(
     if (typeof value === "function") {
       wrapFunction(helpers, wrappedAPI, workerScript, value, ...tree, key);
     } else if (Array.isArray(value)) {
-      setNestedProperty(wrappedAPI, value, key);
+      setNestedProperty(wrappedAPI, value.slice(), key);
```
TheMas3212, line 151: whats your reasoning for doing a slice here?

src/Netscript/APIWrapper.ts
Snarling: With spread on one end and rest on the other, it's taking an array and turning it into individual entries on one end, then converting multiple entries back into an array on the other. I just changed both ends to just pass the `tree` array directly so the end result is the same, there's just less converting back and forth.

src/Netscript/APIWrapper.ts
Snarling: I was trying to address a separate bug where scripts do not end gracefully if the player mutates the args array. This didn't end up being a solution but I think it still makes sense not to give players access directly to an actual array the game uses internally, so the player just gets a copy of it.

src/Netscript/APIWrapper.ts
TheMas3212: i can understand that i suppose

src/Netscript/APIWrapper.ts
TheMas3212: ok, honestly at somepoint i want to put a sanitizer in this wrapper that does a deepCopy on any return values to prevent that completely, but this seems fine for now

src/NetscriptEvaluator.ts
```diff
@@ -4,20 +4,21 @@ import { ScriptDeath } from "./Netscript/ScriptDeath";
 import { WorkerScript } from "./Netscript/WorkerScript";
 import { NetscriptContext } from "./Netscript/APIWrapper";
 
-export function netscriptDelay(time: number, workerScript: WorkerScript): Promise<void> {
+export function netscriptDelay(fnName: string, time: number, workerScript: WorkerScript): Promise<void> {
```
TheMas3212, line 7: if your going to modify netscriptDelay to require the function name you should match the existing functions that do so and put a wrapped version in the context helpers to pass in the function name automatically

Snarling: Standardized in #3967 which includes all these changes. Closing this PR since everything should still go in on the other one.

Snarling:
Replaced by #3967

PR #3967, opened 2022-08-08
NETSCRIPT: Improve real life CPU and memory performance of scripts.

### LINKED ISSUE #3958: v2: most functions not checking for env.stopFlag

Snarling:
Fixes #3958 , since it includes the fixes from #3960 .

Monster PR.

Moved almost all functions that were used by NetscriptFunctions but were not actually part of ns into a separate file, NetscriptHelpers.ts.

NetscriptFunctions.ts now for the most part only contains the actual internal NS API, and it has been moved to the top level since it is put through a wrapper anyway.

Performance with the latest commit was approx 2400ms for 10000 back-to-back script executions (starting the next script as soon as the previous script ends using promises). Image is from the initial commit.

image


Snarling:
Moved ctx.log to `helpers` in NetscriptHelpers.ts, now ctx is only a reference to the WorkerScript, a function name, and a full function path.

Probably the last performance boost this significant. Time for 10000 script executions reduced to 4606ms from 5273ms on previous commit.  Heap now cycles less than 1 full time for these 10000 script executions.

image

phyzical:
daymn 50% performance increase nice

PR #3972, opened 2022-08-10
RFA: New Remote File API addition for transmitting files to the game

Hoekstraa:
# RFA: New Remote File API addition for transmitting files to the game

Implementation of the game-side Websocket protocol, as discussed by Hydro and me.
I'd love it if people could have a look over the code and give feedback where needed!

## Try it out
If you want to try it out, here's links to what's needed:
### A build of the game with RFA
Build of the game with this addition: https://hoekstraa.github.io/bitburner/
### An RFA server
Alpha version of server-side: https://github.com/Hoekstraa/Bitburner-Remote
OR
A Javascript version of server-side: https://github.com/Hoekstraa/bitburner-ws-server

Currently both servers are still WIP, so it might require some finagling to get them working how you want them to.
Feel free to contact me about it on the official Bitburner Discord (Zoë/AltLock#5724)

### DIFF
--- a/src/GameOptions/ui/ConnectionBauble.tsx
+++ b/src/GameOptions/ui/ConnectionBauble.tsx
@@ -1,41 +1,21 @@
-import React from "react";
-
-interface baubleState {
-  connection: boolean;
-  callback: () => boolean;
-}
+import React, {useState, useEffect } from "react";
 
 interface baubleProps {
   callback: () => boolean;
 }
 
-export class ConnectionBauble extends React.Component<baubleProps> {
-  timerID: NodeJS.Timer;
-  state: baubleState;
+export const ConnectionBauble = (props: baubleProps): React.ReactElement  => {
+  const [connection, setConnection] = useState(props.callback())
 
-  constructor(props: baubleProps) {
-    super(props);
-    this.state = {
-      connection: props.callback(),
-      callback: props.callback,
-    };
-  }
+  useEffect(() => {
+    setInterval(() => {
+      setConnection(props.callback());
+    }, 1000);
+  });
 
-  componentDidMount(): void {
-    this.timerID = setInterval(() => this.tick(), 1000);
-  }
-
-  componentWillUnmount(): void {
-    clearInterval(this.timerID);
-  }
-
-  tick(): void {
-    this.setState({
-      connection: this.state.callback(),
-    });
-  }
-
-  render(): string {
-    return this.state.connection ? "Connected" : "Disconnected";
-  }
+  return (
+    <div className="ConnectionBauble">
+      {connection? "Connected" : "Disconnected"}
+    </div>
+  );
 }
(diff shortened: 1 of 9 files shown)


src/GameOptions/ui/ConnectionBauble.tsx
```diff
@@ -0,0 +1,41 @@
...
+interface baubleState {
+  connection: boolean;
+  callback: () => boolean;
+}
+
+interface baubleProps {
+  callback: () => boolean;
+}
+
+export class ConnectionBauble extends React.Component<baubleProps> {
```
Snarling, line 12: Just for consistency, I believe the rest of the project only uses functional React components rather than class components.

src/index.tsx
```diff
@@ -14,6 +17,8 @@ ReactDOM.render(
   document.getElementById("root"),
 );
 
+newRemoteFileApiConnection();
```
Snarling, line 20: My understanding at a glance is that as soon as the player loads in the game will be listening for a remote connection on a particular port. Shouldn't the default be for there to be no remote connection, until/unless the player enables a remote connection setting in the options?

Snarling:
PR fails the build step as well, probably because This branch is [4 commits ahead](https://github.com/Hoekstraa/bitburner/compare/danielyxie:bitburner:dev...rfa), [422 commits behind](https://github.com/Hoekstraa/bitburner/compare/rfa...danielyxie:bitburner:dev) danielyxie:dev.

You need to rebase your branch on the current version of dev branch.

src/index.tsx
Hoekstraa: 2022-08-11-184323_909x458_scrot

src/GameOptions/ui/ConnectionBauble.tsx
Hoekstraa: The component has a timer within. I wasn't able to find a way to do anything with a timer (e.g. internal state) when using functional components. If someone is able to point me to how something like that is possible or an alternative method to have a connectionBauble without internal state, I'd be happy to change it.

Hoekstraa: Update: I figured it out pretty quickly after having another look. New commit is with a functional component instead!

Snarling: Nice! Yeah you can see why functional components are preferred, way less code for the same thing.

A few issues though. You're never clearing the interval you set in useEffect. useEffect should return a callback function which is ran when the component unmounts. Also you're still referencing this.state.connection, when it is just connection now.

There's also an error on gameOptionsPage: `import { ConnectionBaub, ConnectionBauble } from "./ConnectionBauble";`

I believe by convention interfaces/types usually are capitalized so they aren't confused for regular variables.

This is what I got while converting it to a functional component (actually I had left in a tick function, but I just removed it after I saw the way got rid of it/it wasn't needed):

```typescript
import React, { useEffect, useState, ReactElement } from "react";

interface BaubleProps {
  callback: () => boolean;
}

export function ConnectionBauble(props: BaubleProps): ReactElement {
  const [connection, setConnection] = useState(props.callback());
  useEffect(() => {
    const timerID = setInterval(() => setConnection(props.callback()), 1000);
    return () => clearInterval(timerID);
  });

  return <span>connection ? "Connected" : "Disconnected"</span>;
}
```

PR #3978, opened 2022-08-17
NETSCRIPT: FIX #2931 atExit now allows synchronous ns functions

### LINKED ISSUE #2391: `atExit` does not execute properly if script is killed during `sleep`

### LINKED ISSUE #2931: TypeError: Cannot read properties of undefined (reading 'name') (at "Work")
TypeError: Cannot read properties of undefined (reading 'name') (at "Work")

APpears to be a duplicate occurrence of:
https://github.com/danielyxie/bitburner/issues/2804

I was running a larger script which just finished doing faction work:
- looping over ns.workForFaction('Chongqing', 'hacking', true);
And was starting to call into code to purchase augments when the crash occurred.

This has happened several times while I was away, but this is the firsttime I saw it well enough to get an idea of where the script was.

I jumped back into the game and it appears that my script may have completed in calling 
ns.installAugmentations('script.js'); and installed my augments. ...

* Platform: Steam
* UserAgent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) bitburner/1.0.0 Chrome/93.0.4577.82 Electron/14.0.2 Safari/537.3

[RECOVERY_BITBURNER_1644196451.zip](https://github.com/danielyxie/bitburner/files/8011683/RECOVERY_BITBURNER_1644196451.zip)
6

### LINKED ISSUE #4011: Error on atExit when script finishes
## Description

When a script finishes executing and `ns.atExit(() => f)` is provided, following Error occurs:
`Error trying to call atExit for script test.js on home [object Object]`

However, it seems that this only occurs if ns is being used inside of `atExit()`.
Also, if the script is being terminated with `ns.exit()` before "finishing" the script, the error does not occur.

## Steps to reproduce

1. Set ns.atExit() and call a method from ns inside the passed function.
2. Don't terminate the script with ns.exit();

## Example
```javascript
/** @param {NS} ns */
    ns.atExit(() => ns.print('test'));
}
```

## Version
Bitburner v2.0.2 (3067703c)

I noticed this as my main script used to work until (i think) 2 versions ago.
It might also be that i'm just too stupid and not realizing something important. Please let me know if I handled something wrongly.

Snarling:
Fixes #2391
Fixed #4011 

Should also be able to close #2588.

After a script is killed, the stopFlag is briefly set to false for a synchronous run of the atExit function. Once that is complete, the stopFlag is set back to true and any subsequent ns functions ran from the script will fail. This means that synchronous ns functions are allowed in atExit (like tprint) but asynchronous functions are not allowed/will not work.

The error message handling at this point has also been improved. I used this test script to generate an error message:

```javascript
export let main=ns=>ns.atExit(()=>ns.getServer(ns.args[0]));
```

Here is a comparison of old and new error messages with `run atExitTest.js gibson`:

image

Old above, new below:

image

### DIFF
--- a/src/Netscript/killWorkerScript.ts
+++ b/src/Netscript/killWorkerScript.ts
@@ -52,22 +52,24 @@ function killWorkerScriptByPid(pid: number): boolean {
   return false;
 }
 
-function stopAndCleanUpWorkerScript(workerScript: WorkerScript): void {
-  if (typeof workerScript.atExit === "function") {
+function stopAndCleanUpWorkerScript(ws: WorkerScript): void {
+  killNetscriptDelay(ws);
+  if (typeof ws.atExit === "function") {
     try {
-      workerScript.atExit();
+      ws.env.stopFlag = false;
+      ws.atExit();
     } catch (e: unknown) {
+      let message = e instanceof ScriptDeath ? e.errorMessage : String(e);
+      message = message.replace(/.*\|DELIMITER\|/, "");
       dialogBoxCreate(
-        `Error trying to call atExit for script ${workerScript.name} on ${workerScript.hostname} ${
-          workerScript.scriptRef.args
-        } ${String(e)}`,
+        `Error trying to call atExit for script ${ws.name} on ${ws.hostname} ${ws.scriptRef.args}\n` + message,
       );
+      console.log(e);
     }
-    workerScript.atExit = undefined;
+    ws.atExit = undefined;
   }
-  workerScript.env.stopFlag = true;
-  killNetscriptDelay(workerScript);
-  removeWorkerScript(workerScript);
+  ws.env.stopFlag = true;
+  removeWorkerScript(ws);
 }
 
 /**
--- a/src/Netscript/killWorkerScript.ts
+++ b/src/Netscript/killWorkerScript.ts
@@ -62,7 +62,7 @@ function stopAndCleanUpWorkerScript(ws: WorkerScript): void {
       let message = e instanceof ScriptDeath ? e.errorMessage : String(e);
       message = message.replace(/.*\|DELIMITER\|/, "");
       dialogBoxCreate(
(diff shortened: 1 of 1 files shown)


src/Netscript/killWorkerScript.ts
```diff
@@ -52,22 +52,24 @@ function killWorkerScriptByPid(pid: number): boolean {
...
+      ws.env.stopFlag = false;
+      ws.atExit();
     } catch (e: unknown) {
+      let message = e instanceof ScriptDeath ? e.errorMessage : String(e);
+      message = message.replace(/.*\|DELIMITER\|/, "");
       dialogBoxCreate(
-        `Error trying to call atExit for script ${workerScript.name} on ${workerScript.hostname} ${
-          workerScript.scriptRef.args
-        } ${String(e)}`,
+        `Error trying to call atExit for script ${[ws.name, ...ws.args].join(" ")} on ${ws.hostname}\n ${message}`,
```
phyzical, line 65: does ws.args ever contain an arg that is also objects? id assume not due to how the api works but just a thought

phyzical (approved):

src/Netscript/killWorkerScript.ts
Snarling: Nope, they're limited to string/number/boolean, all of which cleanly convert to string.

hydroflame:
I'm merging but don't leave console.log

PR #3980, opened 2022-08-17
CORP FIX: dont take research points for something already researched via api

### LINKED ISSUE #3977: Make ns.corporation.research() check if the research is already researched before charging the research points.
Could it be that the netscript function [ns.corporation.research(divisionName, researchName)](https://github.com/danielyxie/bitburner/blob/dev/markdown/bitburner.officeapi.research.md) doesn't check whether the research is already unlocked before charging the research points from the player (the division of the corporation)?

I wrote a script to automate corporations and it constantly spends research points on "Hi-Tech R&D Laboratory" because of a missing check if it is already researched in my code. So I wonder if it is really nice of the game to show this behaviour. I would expect the function to only charge the points when there's an effect to be expected. But calling the research function on the same target over and over again seems to try to research it again and again - leaving the division poor of points :-(

So... just asking: Is this intended? Could it be fixed?

Thanks in advance!

phyzical:
closes #3977

### DIFF
--- a/src/Corporation/Actions.ts
+++ b/src/Corporation/Actions.ts
@@ -455,6 +455,7 @@ export function Research(division: IIndustry, researchName: string): void {
   if (!allResearch.includes(researchName)) throw new Error(`No research named '${researchName}'`);
   const research = ResearchMap[researchName];
 
+  if (division.researched[researchName]) return;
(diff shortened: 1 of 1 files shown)


Snarling (approved):
👍

PR #3981, opened 2022-08-17
FILES: FIX #3979 Allow characters & and ' in filenames

### LINKED ISSUE #3979: Cannot delete contract files with special characters in filename
When attempting to delete a contract I'd resigned to the "too hard basket" the `rm` command failed, saying the file did not exist.

Output:
```
[zeus-med ~/]> ls
contract-45485-Bachman&Associates.cct 
[zeus-med ~/]> rm contract-45485-Bachman&Associates.cct
No such file exists
```

- **Version**: v1.7.0
- **To reproduce**: find a contract file containing a name like "Joe'sGuns" or "Bachman&Associates" and attempt to delete it with `rm`

I think this is happening because the apostrophe and ampersand characters are failing `isValidFilename` validation in the Terminal/DirectoryHelpers, where the regex only "Allows alphanumerics, hyphens, underscores, and percentage signs".

I'm of two minds about this could be fixed.
1. Expanding the regex with `\'\&` symbols could work, but there could be a side-effect of validating some strings that shouldn't be considered valid, in some part of the system I don't yet understand.
2. Altering the way contract filenames are generated to avoid or remove the invalid characters, e.g. "JoesGuns", "BachmanAssociates" or "BachmanAndAssociates", so that they pass validation. Might be more effort, but a more stable change (at least for filenames generated in future versions of the game).

Snarling:
Fixes #3979 

Those characters are already used in coding contracts.

### DIFF
--- a/src/Terminal/DirectoryHelpers.ts
+++ b/src/Terminal/DirectoryHelpers.ts
@@ -38,7 +38,7 @@ export function removeTrailingSlash(s: string): string {
 export function isValidFilename(filename: string): boolean {
   // Allows alphanumerics, hyphens, underscores, and percentage signs
   // Must have a file extension
-  const regex = /^[.a-zA-Z0-9_-]+[.][a-zA-Z0-9]+(?:-\d+(?:\.\d*)?%-INC)?$/;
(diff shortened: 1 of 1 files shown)


phyzical:
(might be a good one to write a basic test for but no pressure)

PR #3985, opened 2022-08-17
NETSCRIPT: ns.scp and ns.write are now synchronous + fix exec race condition

Snarling:
The only async portion of these was previously the static ram calculation, and only while checking ram usage on imports from http/https imports.

Using ns functions in files imported from http/https was already playing with fire because the ram calculation would not be performed again if the http file was modified. If a player does need to have ns functions that cost ram on remote scripts, they can still work around the dynamic ram calculation by manually paying the ramcost in the main script (just by referencing the function name).

HTTP imports should still work fine, they just will not increase the static ram calculation, so will lead to dynamic ram errors if the player is using ns functions that cost ram in their http imports that they aren't already paying for in the script itself.

Additionally and separate from synchronizing write and scp / disabling static ram checking on http imports, this PR also fixes an error that was present when trying to exec the same uncompiled script multiple times before the first compilation was complete. Previously the compilations would interfere with one another and only the last execution would be successful, now all instances started during compilation will wait for the first compile to finish.

### DIFF
--- a/src/NetscriptFunctions.ts
+++ b/src/NetscriptFunctions.ts
@@ -829,11 +829,7 @@ const base: InternalAPI<NS> = {
   },
   scp:
     (ctx: NetscriptContext) =>
-    async (
-      _scriptname: unknown,
-      _destination: unknown,
-      _source: unknown = ctx.workerScript.hostname,
-    ): Promise<boolean> => {
+    (_scriptname: unknown, _destination: unknown, _source: unknown = ctx.workerScript.hostname): boolean => {
       const destination = helpers.string(ctx, "destination", _destination);
       const source = helpers.string(ctx, "source", _source);
       if (Array.isArray(_scriptname)) {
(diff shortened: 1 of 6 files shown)


Snarling:
http imports don't seem to be working with most files the player would be writing and storing remotely anyway, because there is a strict mime type check. 

e.g. I wrote a script and tried to import it from https://raw.githubusercontent.com/Snarling/bitburner-scripts/main/ramTest.js but that errors out because it identifies itself as text instead of as a js module.

image

Snarling:
The scp function was previously extremely inefficient when an array was passed in - it would invoke NetscriptFunctions once for every file in the array, basically creating 10 new wrapped APIs if there were 10 files in the array. As I was fixing this I also refactored a lot of other code,  e.g. where a for loop and "found" flag were being used to find whether an entry exists in an array, I reworked that into an array.find.

alainbryden:
If you want to get http imports a proper try, check out the solution suggested at https://stackoverflow.com/a/18049842/529618

Even if your change has broken this use case though, I think there's an argument to be made that the pros outweigh the cons

Snarling:
> Even if your change has broken this use case though, I think there's an argument to be made that the pros outweigh the cons

You still would be able to use http imports, they just would not contribute to static ram cost if those imported files were referencing ns functions that cost ram. Even if you did have ram-using http imports, you could still use them, it would just require a workaround where you'd have to manually add enough unused static ram usage to your script to account for the unpayed functions used in the import.

Probably worth adding an extra note in the dynamic ram error message though. I will add that tomorrow morning.

Snarling:
There should no longer be any race conditions surrounding script compilation.

PR #3999, opened 2022-08-18
NETSCRIPT: Correctly log "stock.shortStock" instead of "stock.short".

MageKing17:
This was preventing `disableLog("ALL")` from stopping these log messages, since that's not the function's name now.


Snarling:
Those logs can also be changed to `helpers.log(ctx, ()=>message)` so that you don't have to hardcode the name of the function.

MageKing17:
I think a bigger problem than the lack of import is that `ctx` isn't in scope.

PR #4000, opened 2022-08-19
NETSCRIPT: Correct missing `!` for boolean coercion in `singularity.workForCompany()`.

MageKing17:
`singularity.workForCompany()` was negating its `_focus` argument, unlike similar functions, which used double-negation to coerce to boolean. This was almost certainly a typo, since before PR-#3967 it used `_ctx.helper.boolean()` without negation, just like the other singularity functions.

### DIFF
--- a/src/NetscriptFunctions/Singularity.ts
+++ b/src/NetscriptFunctions/Singularity.ts
@@ -780,7 +780,7 @@ export function NetscriptSingularity(): InternalAPI<ISingularity> {
       function (_companyName: unknown, _focus: unknown = true): boolean {
         helpers.checkSingularityAccess(ctx);
         const companyName = helpers.string(ctx, "companyName", _companyName);
-        const focus = !_focus;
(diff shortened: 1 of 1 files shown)


phyzical:
i wonder why we stopped using the helper to make it a bool though?

MageKing17:
Presumably because the helper was just doing `!!` to the input, unlike other helpers that actually do input validation.

Snarling:
Yeah I just removed the boolean helper because it didn't actually seem useful/needed since it's easier to just `!!` something instead of calling a helper function that does that. I tend to focus my PRs on "trimming the fat" and part of that PR was reorganizing a bunch of helper functions, so I just culled it. My bad on the typo.

Snarling (approved):

phyzical (approved):

hydroflame:
To be fair. The reason the function existed is to avoid situations like these. I know it's a simple mistake but they happen.

PR #4012, opened 2022-08-20
DOCS: Correct documentation for `run()` with 0 threads.

MageKing17:
For consistency, the same phrasing has been applied to the documentation for `exec()` and `spawn()`.


Snarling (changes requested):
I believe the dist/bitburner.d.ts file is autogenerated from the NetscriptDefinitions.d.ts file, only the latter needs to be modified. The manual changes to the dist file should be removed.

MageKing17:
Wasn't 100% sure on that; I've re-done the commit with just changing NetscriptDefinitions.d.ts.

PR #4020, opened 2022-08-22
API FIX: add currentWork to getPlayer result

### LINKED ISSUE #4004: BruteSSH.exe Reacting wierldy to stopping/unfocusing.
After taking a break from creating BruteSSH.exe, the create a program button will try to create BruteSSH.exe from scratch instead of continuing where you left off. If this one is stopped as well, the previously worked on BruteSSH.exe will reappear. If repeated once more, a new, empty BruteSSH.exe can be found and seems to be repeatable indefinitely.

Happened in 2.0.2

[bitburnerSave_1660927018_BN1x1.zip](https://github.com/danielyxie/bitburner/files/9384027/bitburnerSave_1660927018_BN1x1.zip)

### LINKED ISSUE #4009: currentWork is missing from the return value of ns.getPlayer
It looks like the return value of ns.getPlayer() should include the 'currentWork' field, but it's missing in 2.0.2 (3067703)

https://github.com/danielyxie/bitburner/blob/07c7f0641a5dffb0c0139448201876ba6e001b63/src/utils/v2APIBreak.ts#L225

https://github.com/danielyxie/bitburner/blob/2abc5687e14dadcd1fb867e2945fff563b5a7cc5/src/NetscriptFunctions.ts#L1912-L1933

### LINKED ISSUE #4017: ns.singularity.getCurrentWork is not a function
When executing "ns.singularity.getCurrentWork()" an error is thrown stating the function doesn't exist.

phyzical:
closes #4009
closes #4017

### DIFF
--- a/src/NetscriptFunctions.ts
+++ b/src/NetscriptFunctions.ts
@@ -1925,6 +1925,7 @@ const base: InternalAPI<NS> = {
       playtimeSinceLastAug: Player.playtimeSinceLastAug,
       playtimeSinceLastBitnode: Player.playtimeSinceLastBitnode,
       jobs: {},
+      currentWork: Player.currentWork,
       factions: Player.factions.slice(),
       tor: Player.hasTorRouter(),
       inBladeburner: Player.inBladeburner(),


Snarling:
Should be 4009, not 4004.

hydroflame:
The message is wrong. I'd rather not feed too much data to getPlayer, it's correct in singularity.getCurrentWork

phyzical:
👍

PR #4039, opened 2022-08-25
GANG: Added weight to GangMemberTask construction call

ezylot:
Added hackWeight: 100 to GangMember Task construction call to prevent error messages from the constructor checks. 
Since we only compare with roughlyIs() this will have no impact on functionality


ezylot:
grafik

PR #4040, opened 2022-08-25
DOCS, UI: update docs a bit more, amending some BN and SF texts

Mughur:
Doc's bitnode info was lacking/incorrect, now it's less so.

SF3.3 gives access to the full API for free, but this is never mentioned anywhere, added mention of this to the docs and the BitVerse's BN3 text.

Added info about how much SF4.X affects singularity function costs to BitVerse and SF text.

Added mention of what grafting is to BN10 BitVerse text


doc/source/guidesandtips/recommendedbitnodeorder.rst
```diff
@@ -122,7 +124,11 @@ Source-File
     :Max Level: 3
 
     This Source-File lets you access and use the Singularity Functions in other BitNodes.
-    Each level of this Source-File will reduce RAM costs.
+    Each level of this Source-File will reduce RAM costs:
+
+    * Level 1: 16x
+    * Level 2: 6x
```
Kelenius, line 130:
```suggestion
    * Level 2: 4x
```

src/BitNode/BitNode.tsx
```diff
@@ -144,7 +144,13 @@ BitNodes["BitNode4"] = new BitNode(
       <br />
       Destroying this BitNode will give you Source-File 4, or if you already have this Source-File it will upgrade its
       level up to a maximum of 3. This Source-File lets you access and use the Singularity Functions in other BitNodes.
-      Each level of this Source-File reduces the RAM cost of singularity functions.
+      Each level of this Source-File reduces the RAM cost of singularity functions:
+      <br />
+      Level 1: 16x
+      <br />
+      Level 2: 6x
```
Kelenius, line 151:
```suggestion
      Level 2: 4x
```

hydroflame:
This code also needs to be formatted

src/PersonObjects/Sleeve/Work/SleeveBladeburnerWork.ts
```diff
@@ -34,9 +34,26 @@ export class SleeveBladeburnerWork extends Work {
...
-      const actionIdent = player.bladeburner.getActionIdFromTypeAndName(this.actionType, this.actionName);
-      if (!actionIdent) throw new Error(`Error getting ${this.actionName} action`);
+      if (this.actionType === "Contracts") {
+        const action = player.bladeburner.getActionObject(actionIdent);
+        if (!action) throw new Error(`Error getting ${this.actionName} action object`);
+        if (action.count <= 0) {
+          sleeve.stopWork(player);
+          return 0;
+        }
+      }
```
phyzical, line 56: i dont think we need to duplicate this code just having it once in the while loop should do the same thing,

or is there some condition that makes it keep working and the while loop is never entered?

phyzical (approved):

src/PersonObjects/Sleeve/Work/SleeveBladeburnerWork.ts
phyzical: actually looks like youve covered this code in your other pr mind removing for this one?

Mughur: @phyzical 
> i dont think we need to duplicate this code just having it once in the while loop should do the same thing,

I put it there twice because the first check would only happen during the action and second happens after the action. if only the first is there one could get tons of cycles and still go to negative contracts. if only the second is there the game would take to finish the action and then not give any rewards, It's not much time tho. Removing the first check and adding a check to when the action is started would be better, but I couldn't figure out how to add that

> or is there some condition that makes it keep working and the while loop is never entered?

no

> actually looks like youve covered this code in your other pr mind removing for this one?

I'm actually not sure how this commit ended up on this branch -.-'

phyzical (approved):

PR #4044, opened 2022-08-25
RFA: Fix type of RFAMessages with non-String results

Hoekstraa:
non-String result values were incorrectly parsed as JSON twice, instead of being their originally intended types.
This pull fixes that.


phyzical:
hmm i wonder why we ever parsed this as json... 

(also looks like you may have to rebase your branch)

Hoekstraa:
@phyzical answers to both are *my dumb mistakes, sorry!* :sweat_smile:

PR #4049, opened 2022-08-26
NETSCRIPT: FIX #3963 Prevent bladeburner.setActionLevel from setting invalid action levels

### LINKED ISSUE #3963: bladeburner.setActionLevel does not throw an error when an invalid level is passed and allows any level to be set
The title pretty much says it all. Game version is v1.7.0 (cfcdc254). EDIT: Issue still exists in v2.0.2 (`3067703c`).
The `bladeburner.setActionLevel()` function will allow any level to be set, even if it is above the maximum level. This allows the player to cheat in bladeburner, because actions will work at the set level. I have not tried setting negative levels, because I fear it will break the game.

I do not sufficiently understand the source code to suggest an exact fix. However, I did notice that in [this](https://github.com/danielyxie/bitburner/blob/326d9fd7ef4042fb8fc031e6a94eeef091c98521/src/NetscriptFunctions/Bladeburner.ts#L253) file, (line 253) `ctx.helper.MakeRuntimeErrorMsg()` is used, while every other call to this function uses `ctx.MakeRuntimeErrorMsg()`. Perhaps that is where the problem lies.

MPJ-K:
closes #3963

Adds a previously missing `throw` to allow `bladeburner.setActionLevel()` to properly produce errors and prevent invalid action levels from being set.

Bug was thoroughly tested with screenshot proof. See the mentioned issue for details. I am not set up to test this fix in-game, so please forgive me if I end up wasting anyone's time with this pull request.

### DIFF
--- a/src/NetscriptFunctions/Bladeburner.ts
+++ b/src/NetscriptFunctions/Bladeburner.ts
@@ -250,7 +250,7 @@ export function NetscriptBladeburner(): InternalAPI<INetscriptBladeburner> {
         checkBladeburnerAccess(ctx);
         const action = getBladeburnerActionObject(ctx, type, name);
         if (level < 1 || level > action.maxLevel) {
-          helpers.makeRuntimeErrorMsg(ctx, `Level must be between 1 and ${action.maxLevel}, is ${level}`);
+          throw helpers.makeRuntimeErrorMsg(ctx, `Level must be between 1 and ${action.maxLevel}, is ${level}`);
         }
         action.level = level;
       },


MPJ-K:
Tested this in-game today by modifying my `main.bundle.js`.
Changed
```
(c<1||c>f.maxLevel)&&s.a.makeRuntimeErrorMsg(e,`Level must be between 1 and ${f.maxLevel}, is ${c}`),f.level=c
```
(the above code looks strange to me, but that is how the game ships right now)
to
```
if(c<1||c>f.maxLevel)throw s.a.makeRuntimeErrorMsg(e,`Level must be between 1 and ${f.maxLevel}, is ${c}`);f.level=c
```
This fixed the issue, errors are being thrown and invalid levels are not set:
BitburnerSetActionLevelBugFix1
BitburnerSetActionLevelBugFix2

PR #4054, opened 2022-08-28
NETSCRIPT: Unify internal error handling code

### LINKED ISSUE #4037: Not all async functions are wrapped properly by NS1 interpreter.
Discovered via reddit : https://www.reddit.com/r/Bitburner/comments/wx2bg6/n00b_trying_and_failing_bn13_scripting/
Player reported chargeFragment creating an unawaited promise error. Looks like stanek.chargeFragment just is not being wrapped by the NS1 async function handler.

https://github.com/danielyxie/bitburner/blob/ccbca5c38cd75ca4c89cc0c4ad42fe167187da13/src/NetscriptWorker.ts#L126-L127

stanek.chargeFragment is not the only discrepancy here, will need to fully match the list of promise-returning ns functions against the hardcoded list that is being fed to this ns1 async handler.

Snarling:
Errors are handled a bunch of different places, and inconsistently.
* Once in the NS1 wrapper function in startNetscript1Script (NetscriptWorker.ts)
* Once in startNetscript2Script (NetscriptWorker.ts)
* Once in createAndAddWorkerScript (NetscriptWorker.ts)
* Again in the atExit handler in killWorkerScript.ts
* And finally in UncaughtPromiseHandler.ts

All with separate error handling code.

For displaying errors, now there is one handleUnknownError function in NetscriptHelpers.js, which is used in just three locations to catch all errors (createAndAddWorkerScript, atExit handler, and UncaughtPromiseHandler). The first two instances above have been reworked to always properly throw their errors to createAndAddWorkerScript, so the handling for within the main execution of a script can all be handled in that one location.

Modifications were made to startNetscript1Script so that it actually throws an error when an error is encountered, previously it was returning a resolved promise but manually displaying an error dialog box.

Initial version added a new handler to DialogBox.tsx. Since I was modifying this, I also changed dialogBoxCreate to no longer do the dangerouslySetInnerHTML thing it was doing before. Now it just expects regular strings (not strings of HTML). For styled output it will still accept TSX. This is one of the reasons this PR touches so many files, any reference to dialogBoxCreate that used an html string needed to be modified to be a regular string (i.e. `<br>` changed to `\n`, `&nbsp;` changed to `\u00a0`). This also means that errors which contain html content will not be rendered as HTML, they will be rendered as plaintext.

After removing the error handling code from it, startNetscript2Script was basically just a wrapper for executeJSScript. So I  just reworked executeJSScript slightly to take just a workerScript as an input, and moved it to NetscriptWorker.ts as a replacement for startNetscript2Script. Changing the params to work required importing Player directly at RamCostGenerator.ts and removing player from a bunch of parameter chains throughout multiple files, most of which were just for the purpose of delivering Player to RamCostGenerator. During some troubleshooting for these changes, I also removed the prop passing of Terminal/Engine/Player from LoadingScreen to GameRoot. Now GameRoot imports those items.

Errors no longer use the |DELIMITER| format they previously did, but there is still checking to add header information to an error if it isn't already present.

I also renamed makeRuntimeRejectMsg to makeBasicErrorMsg, and now it also adds this header info. It also will optionally accept a ScriptDeath as an input instead of a WorkerScript to produce this header info.

Also, there is a new field for both ErrorMsg functions to specify the type of error. For instance, if an error is thrown due to an invalid type provided for a parameter, the error will say "TYPE ERROR" instead of "RUNTIME ERROR". Other types I have added are:
Dynamic ram check failure: "RAM USAGE ERROR"
netscriptDelay concurrency failure: "CONCURRENCY ERROR"
singularity error for not having SF4/being in BN4: "API ACCESS ERROR"

### DIFF
--- a/src/Augmentation/AugmentationHelpers.tsx
+++ b/src/Augmentation/AugmentationHelpers.tsx
@@ -126,15 +126,15 @@ function installAugmentations(force?: boolean): boolean {
     if (ownedAug.name === AugmentationNames.NeuroFluxGovernor) {
       level = ` - ${ownedAug.level}`;
     }
-    augmentationList += aug.name + level + "<br>";
+    augmentationList += aug.name + level + "\n";
   }
   Player.queuedAugmentations = [];
   if (!force) {
     dialogBoxCreate(
       "You slowly drift to sleep as scientists put you under in order " +
-        "to install the following Augmentations:<br>" +
+        "to install the following Augmentations:\n" +
         augmentationList +
-        "<br>You wake up in your home...you feel different...",
+        "\nYou wake up in your home...you feel different...",
     );
   }
   prestigeAugmentation();
--- a/src/Casino/Game.tsx
+++ b/src/Casino/Game.tsx
@@ -11,7 +11,7 @@ export function win(p: IPlayer, n: number): void {
 export function reachedLimit(p: IPlayer): boolean {
   const reached = p.getCasinoWinnings() > gainLimit;
   if (reached) {
-    dialogBoxCreate(<>Alright cheater get out of here. You're not allowed here anymore.</>);
+    dialogBoxCreate("Alright cheater get out of here. You're not allowed here anymore.");
   }
   return reached;
 }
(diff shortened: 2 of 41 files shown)


Snarling:
Not ready for merge, looks like actual `Error`s also won't have any header information right now.

Snarling:
@hydroflame I should be able to look over and resolve the conflicts for this one in a few hours.

Snarling:
@hydroflame I rebased and fixed conflicts on this

Snarling:
Effectively merged in via #4122

PR #4059, opened 2022-08-28
RFA: NetscriptDefinitions retains export strings

Hoekstraa:
Fixes the missing `export`s in NetscriptDefinitions sent through RFA. Should resolve the issues regarding importing the file after downloading it through filesync.


Hoekstraa:
Don't ask me how this happened...

PR #4069, opened 2022-08-31
API: Corp functions now return copy of constant arrays instead of the original

Mughur:
Corp functions `getMaterialNames`, `getIndustryTypes`, `getUnlockables`, `getUpgradeNames` and `getResearchNames` returned the original arrays, which if modified, modified the original array instead of just the returned array. This array could be used as a global variable and resets on prestige. Changed the functions to return copies of the original so that the arrays can be manipulated without affecting the original.


phyzical (approved):
at this point we should consider making a wrapper for all api responses 😆

PR #4073, opened 2022-09-01
API: More ports (previously max 20, now practically unlimited)

Hoekstraa:
# API: More ports (previously max 20, now practically unlimited)

Lets ports be created on-demand instead of always initializing all ports.
Also bump the maximum number up to the safe maximum.

- Formatted
- Linted
- Tested with:

```js
/** @param {NS} ns */
//works
	await ns.writePort(1.5, "Hello!");
	ns.tprint(ns.readPort(1.5));
//error, over the max limit
	await ns.writePort(Infinity, "Hello again!");
	ns.tprint(ns.getPortHandle(Infinity).peek())
//works
	await ns.writePort(1e103, "Big number");
	ns.tprint(ns.readPort(1e103));
//error, under limit
	await ns.writePort(0, "Hello again!");
	ns.tprint(ns.getPortHandle(0).peek())
}
```

### DIFF
--- a/src/Constants.ts
+++ b/src/Constants.ts
@@ -124,7 +124,7 @@ export const CONSTANTS: {
   // NeuroFlux Governor Augmentation cost multiplier
   NeuroFluxGovernorLevelMult: 1.14,
 
-  NumNetscriptPorts: Infinity,
+  NumNetscriptPorts: Number.MAX_SAFE_INTEGER,
 
   // Server-related constants
   HomeComputerMaxRam: 1073741824, // 2 ^ 30
--- a/src/Constants.ts
+++ b/src/Constants.ts
@@ -124,7 +124,7 @@ export const CONSTANTS: {
   // NeuroFlux Governor Augmentation cost multiplier
   NeuroFluxGovernorLevelMult: 1.14,
 
-  NumNetscriptPorts: 20,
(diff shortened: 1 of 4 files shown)


phyzical:
im not 100% but i think this was done due to performance issues with too many ports?

Hoekstraa:
@phyzical 
2022-09-02-115532_1536x975_scrot

Here's a quick benchmark with 10000 ports on my laptop (i7 7700HQ, battery saving mode, no turboboost, Firefox, Linux).
Initial numbers (include initialization of the Arrays) end up with records between 737 and 845 ms.
Running the script again (meaning on initialisation) end up with numbers between 395 and 478.

I plugged it in as well, and now it's roughly 480ms, regardless of initializing the ports or not (Yes, I ctrl-F5'ed the page to clean).
So on a hampered i7 7700HQ, it'd take about half a millisecond to do the instructions in the script * 10000.

Doesn't seem to be too bad to me.

hydroflame:
the get function should return already created ports or create one if it doesn't exists. That'll simplify the duplicated code.

Hoekstraa:
@hydroflame thanks for the feedback!
I believe you meant that tryWritePort wasn't using the helper, which meant duplicated code.
That's fixed now. :+1:

PR #4074, opened 2022-09-02
CORP API: Fix up param order for limitProductProduction to match docs

### LINKED ISSUE #4058: Wrong documentation of `ns.corporation.limitProductProduction`
The parameters cityName and productName of `limitProductProduction` do not match the parameter order in the documentation.

## [Netscript Function Definition](https://github.com/danielyxie/bitburner/blob/3ee55b291829854cd97cf7c929041ebcdeba6b0c/src/NetscriptFunctions/Corporation.ts#L535)
```ts
    limitProductProduction:
      (ctx: NetscriptContext) =>
      (_divisionName: unknown, _productName: unknown, _cityName: unknown, _qty: unknown): void => {
```

## [Documentation](https://github.com/danielyxie/bitburner/blob/dev/markdown/bitburner.warehouseapi.limitproductproduction.md)

```ts
limitProductProduction(divisionName: string, cityName: string, productName: string, qty: number): void;
```

## Version

I've found the issue in the `dev` branch and in (through playing) in v2.0.2 (3067703c)

phyzical:
closes #4058

### DIFF
--- a/src/NetscriptFunctions/Corporation.ts
+++ b/src/NetscriptFunctions/Corporation.ts
@@ -534,7 +534,7 @@ export function NetscriptCorporation(): InternalAPI<NSCorporation> {
       },
     limitProductProduction:
       (ctx: NetscriptContext) =>
-      (_divisionName: unknown, _productName: unknown, _cityName: unknown, _qty: unknown): void => {
+      (_divisionName: unknown, _cityName: unknown, _productName: unknown, _qty: unknown): void => {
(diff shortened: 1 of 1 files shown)


Snarling (changes requested):
Wouldn't it be better to change the docs to match the existing parameter order in the function? That way it doesn't break people's currently-working scripts

phyzical:
@Snarling i said the same thing but hydro said meh as only one person has raised the issue

PR #4080, opened 2022-09-05
DOCUMENTATION: Fixed Argument order for scp()

njalooo:
NETSCRIPT: Fixed Argument order for scp()


Snarling (commented):
Title should probably be DOCUMENTATION, since you are fixing the docs, not changing the actual internals of anything. Other than that, looks good to me.

PR #4082, opened 2022-09-05
SCRIPTS: FIX #4081 Rerunning a script from tail window recalculates ram usage

### LINKED ISSUE #4081: Rerunning a script from a tail window does not recalculate ram usage for RunningScript
See https://discord.com/channels/415207508303544321/928919860891750410/1016270565826367489

Snarling:
Fixes #4081 

Previously the run button on a tail window was always using the cached ram calculation for the script from the first time the log box was spawned, due to the runningScript being the same object and this line from RunningScriptHelpers.ts:

https://github.com/danielyxie/bitburner/blob/2592c6ccd89d5559c9cc3cdf99416eb1c57edca2/src/Script/RunningScriptHelpers.ts#L4-L7

Now rerunning a script using the run button will force a cost recalculation by resetting the ramUsage to 0 before relaunching, which disables this caching behavior.

Here is a test on current dev branch:

BrokenAttemptToRunScriptFromTail

And here is the fixed result from this PR:

FixedAttemptToRunScriptFromTail

### DIFF
--- a/src/NetscriptWorker.ts
+++ b/src/NetscriptWorker.ts
@@ -393,10 +393,9 @@ function createAndAddWorkerScript(runningScriptObj: RunningScript, server: BaseS
   const ramAvailable = server.maxRam - server.ramUsed;
   if (ramUsage > ramAvailable + 0.001) {
     dialogBoxCreate(
-      `Not enough RAM to run script ${runningScriptObj.filename} with args ` +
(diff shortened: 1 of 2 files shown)


phyzical (approved):
should we consider moving this logic into startWorkerScript/ disable to caching to avoid this issue in other places or is it an expensive task?

PR #4083, opened 2022-09-05
UI: Split money earned into income and expenses

Kelenius:
Additionally: money the player starts with is also listed in that window; this should make the total always equal player's current money. "Hacknet expenses" category has been removed and merged into "hacknet nodes".

изображение

Any stats gathered before this change will be converted; any positive number will be treated as income and any negative number will be treated as expenses. It's not going to be completely accurate until augment install/next bitnode, but won't cause any errors.

### DIFF
--- a/src/Achievements/Achievements.ts
+++ b/src/Achievements/Achievements.ts
@@ -635,10 +635,7 @@ export const achievements: IMap<Achievement> = {
     ...achievementData["CHALLENGE_BN9"],
     Icon: "BN9+",
     Visible: () => hasAccessToSF(Player, 9),
-    Condition: () =>
-      Player.bitNodeN === 9 &&
-      bitNodeFinishedState() &&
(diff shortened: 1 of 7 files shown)


phyzical (commented):
changes look sound and clean good job :)

my only concern is that this may open it up to dev error in the future if we went with this approach, as there is no more intellisense on the sources provided as they become strings.

the only issue i can see is migrating users before their next prestige/reset or will the fromJSON fallback do a good enough job to preload the user with 0 expenses and then it just works?

Kelenius:
> the only issue i can see is migrating users before their next prestige/reset or will the fromJSON fallback do a good enough job to preload the user with 0 expenses and then it just works?

Yes. I tested importing a game from the current version.

> Any stats gathered before this change will be converted; any positive number will be treated as income and any negative number will be treated as expenses. It's not going to be completely accurate until augment install/next bitnode, but won't cause any errors.

phyzical (approved):

PR #4086, opened 2022-09-06
CORPORATION: Removed employees as objects from corporations

Kelenius:
Changes:

Employees-as-objects have been removed, and only exist as numbers of people assigned to each position and in total.
Automatic/manual employee management switch has been removed. New system is most similar to automatic management.
Employees are treated as a blob with 75 in every stat (average of what they'd have previously).
Experience is treated as a shared pool divided by the employee amount. Hiring a new employee adds 75, nudging the average to that number. Employees gain experience as previously, contributing it to the pool.
Employees in training create 5 times as much experience instead of the previous effect.

**Morale/happiness/energy change speed is reduced, because the changes made it change faster. Longer explanation:**
> previously if you had 9 employees with 100 morale and 1 with 90, the time to reach an average of 100 would be slower than if you had 10 employees with 99 morale. With the PR there's no individual employees, so it works as if every employee has the same morale.

Netscript:

ns.corporation.getEmployee: removed.
ns.corporation.assignJob: removed.
ns.corporation.hireEmployee:
- now returns a boolean: true if an employee was hired, false otherwise.
- **now accepts an optional third parameter: name of the position to place into. If omitted, the employee is unassigned.**

ns.corporation.getOffice:
- `employees` is no longer an array of strings with employee names; now a number with employee amount.
- New members: `avgEne`, `avgHap`, `avgMor`.

Compatability:

Old employees are convered into the new office; total experience, average happiness, morale, and energy are preserved; everything else is discarded.

Testing:

I would like to test this a little more before I'm confident that this won't break anything. Currently I have tested that:

- My own script can run the corp from 0 into Tobacco in BN3 with minimal changes (to accommodate NS changes above)
- Corporations before this change load and nothing important is lost

I would especially appreciate if someone good with corporations would take a look to check that balance remains the same.


Mughur:
I'll run some tests. My corp script takes bit over 2 hours (creation to 1e100/s gains) and to account for randomness, I'll have to run it a few times, with and without your changes. I'll be back with results tomorrow

Kelenius:
New changes:

Morale/happiness/energy change speed is reduced, because the changes made it change faster.

`ns.corporation.hireEmployee` now accepts an optional third parameter: name of the position to place into. If omitted, the employee is unassigned.

PR #4093, opened 2022-09-11
SLEEVES: FIX #3819 Allow using the regeneration chamber with sleeves to heal them.

### LINKED ISSUE #3819: Sleeve HP Loss Persists Across Bitnodes
Now that sleeves can take on Bladeburner contracts, they can lose HP if they fail contracts, which makes sense. The problem is, that lost HP seems to persist, not just across augment cycles (which makes sense), but across bitnodes, as well. Given how literally everything else about sleeves resets, I can't imagine starting the bitnode with 1 HP was intended... especially since the only way to heal them is to send them out to fail more contracts and take the financial hit when they inevitably get hospitalized.

coderanger:
This also allows using sleeves to generate stamina faster even if at full HP and resets them to full HP when starting a new bitnode.

Fixes #3819.

### DIFF
--- a/src/PersonObjects/Sleeve/Sleeve.ts
+++ b/src/PersonObjects/Sleeve/Sleeve.ts
@@ -440,6 +440,9 @@ export class Sleeve extends Person {
       case "Diplomacy":
         this.startWork(p, new SleeveBladeburnerWork({ type: "General", name: "Diplomacy" }));
         return true;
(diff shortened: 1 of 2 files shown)


coderanger:
Screen Shot 2022-09-10 at 6 02 58 PM

Happy little sleeve.

PR #4095, opened 2022-09-11
MISC: Adds info regarding augments and focus

Lagicrus:
Adds info, though tries to avoid spoilers regarding the existence of an augment that gets around the focus debuff

### DIFF
--- a/doc/source/basicgameplay/companies.rst
+++ b/doc/source/basicgameplay/companies.rst
@@ -11,3 +11,7 @@ While working for a company, you can click "Do something else simultaneously" to
 to do things while you continue to work in the background. There is a 20% penalty to the 
 related gains. Clicking the "Focus" button under the overview will return you to the 
 current work.   
+
+.. warning:: This section contains spoilers regarding augments.
+
+It is possible to get an augment to remove the penalty for not focusing on a task such as working on a job or working for a faction
--- a/doc/source/basicgameplay/companies.rst
+++ b/doc/source/basicgameplay/companies.rst
@@ -12,6 +12,6 @@ to do things while you continue to work in the background. There is a 20% penalt
 related gains. Clicking the "Focus" button under the overview will return you to the 
 current work.   
 
-.. warning:: This section contains spoilers regarding augments.
+.. warning:: This section contains spoilers regarding Augmentations.
 
 It is possible to get an augment to remove the penalty for not focusing on a task such as working on a job or working for a faction
--- a/doc/source/basicgameplay/crimes.rst
+++ b/doc/source/basicgameplay/crimes.rst
@@ -13,12 +13,16 @@ Basic Mechanics
 When you visit the 'Slums' you will see a list of buttons that show all of the
 available crimes. Simply select one of the options to begin attempting that
 crime. Attempting to commit a crime takes a certain amount of time. This time
-varies between crimes. 
+varies between crimes.
 
-While doing criemes, you can click “Do something else simultaneously” 
-to be able to do things while you continue to do crimes in the background. There is a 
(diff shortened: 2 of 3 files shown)


quacksouls:
Nice addition.  I see that the documentation uses "Augmentation" or "Augmentations", although at various points you might see "augment".  Perhaps for the sake of consistency, the documentation should start using only "Augmentation" or "Augmentations" as the case arises.  Ignore if this is useless quibble.

Lagicrus:
> Nice addition.  I see that the documentation uses "Augmentation" or "Augmentations", although at various points you might see "augment".  Perhaps for the sake of consistency, the documentation should start using only "Augmentation" or "Augmentations" as the case arises.  Ignore if this is useless quibble.

Tweaked, thanks for the suggestion @quacksouls 
Tempted to add more info regarding the augmentation itself, perhaps even link to it.
But unsure if it already exists in doc form anywhere? Any reason not to add it you think?

quacksouls:
> Tempted to add more info regarding the augmentation itself, perhaps even link to it. But unsure if it already exists in doc form anywhere? Any reason not to add it you think?

I don't remember the online tutorial having anything specific about any particular Augmentation.  It tries to strike a balance between basic introductory information and having as few spoilers as possible.  But if anyone don't mind spoilers, they could search online and find something like this:

https://docs.google.com/spreadsheets/d/1jAmijErKvqwffbu9nBF9SnP905WzZ8gGr-Fe3UbUAOQ/edit#gid=358538604

Within a document on basic game play, you have provided enough information to entice new players to explore and find that specific Augmentation you hint at.

Kelenius:
This feels a bit too specific, and not very fitting to the page. It applies just as much to factions or crimes, why put it on the page with companies?

Lagicrus:
> This feels a bit too specific, and not very fitting to the page. It applies just as much to factions or crimes, why put it on the page with companies?

Factions doesn't seem to talk about the negative effects, and I missed the mention on Crimes
Could easily add it to both honestly?
I added it here as that is where I spotted the issue and lacking text

quacksouls:
> > This feels a bit too specific, and not very fitting to the page. It applies just as much to factions or crimes, why put it on the page with companies?
> 
> Factions doesn't seem to talk about the negative effects, and I missed the mention on Crimes Could easily add it to both honestly? I added it here as that is where I spotted the issue and lacking text

Nice catch to both of you.  The PR by [Lagicrus](https://github.com/Lagicrus) encompasses penalty related to company so naturally the documentation at 

https://bitburner.readthedocs.io/en/latest/basicgameplay/companies.html

should be changed as per the PR.  It also does make sense to have similar addition for crimes:

https://bitburner.readthedocs.io/en/latest/basicgameplay/crimes.html

as well as for faction work:

https://bitburner.readthedocs.io/en/latest/basicgameplay/factions.html

I see the logic in [Kelenius](https://github.com/Kelenius)'s suggestion.  It's a good to have the change apply to all three sections, all in one PR.

Lagicrus:
Updated the additional pages, @quacksouls @Kelenius

Kelenius:
I should have probably explained what I mean by too specific in the previous message, and the thing is I don't see why this information is important enough to be mentioned at all, let alone on three separate pages. Its not very helpful because it doesn't give any hints about where to find that augment, and once you do find that augment, it's useless, because you can immediately see the description once you join the faction that has it. It also awkwardly inserts a spoiler into the middle of a spoiler-free page.

If you want to put information about this augment somewhere in the documentation, what would make sense to me is to make a spoiler page that contains information about all augments in the game. Another, lighter option would be to put it on the augmentations page, to say what augmentations can do, with a mention that some augments have special effects, like removing the focus penalty.

PR #4101, opened 2022-09-12
UI: Bitnode stats now show if BB/Corporation are disabled

Kelenius:
BN8 used to be hardcoded to disable bladeburner; this is now tied to `BitNodeMultipliers.BladeburnerRank` being set to 0. No other bitnode uses it at the moment.

If a bitnode has bladeburner disabled (because of the check above), it will be shown in the stats as such. Similarly, if corporations are disabled (which happens when CorporationSoftcap is less than 0.15), it will also be shown.

BN8 stats:

изображение


hydroflame:
Next time make sure you use triple equal

PR #4102, opened 2022-09-13
Fix for #2442, allowing going forward in tutorial

### LINKED ISSUE #2442: Tutorial gets stuck after the first script if you hit back
I created and saved the first script but then I wanted to go back a tutorial step to read it again but got stuck as there is no longer a forward button on the tutorial prompts and it doesn't count the `nano` command as part of the tutorial anymore.

G4mingJon4s:
Fixes #2442.
The code now shows the forward arrow inside the interactive tutorial, if the step was completed before. This prevents being hard stuck inside the tutorial, if the tutorial prevents any terminal input.


G4mingJon4s:
i did a baddy

PR #4104, opened 2022-09-14
MISC: multiple hasAugmentation checks didn't check if the augment was installed

Mughur:
TRP's message to find the cave was being sent if you had high enough hacking level and had bought the red pill, even if you hadn't installed it yet. Made it so that the `hasAugmentation()` checks whether or not it's installed, not just owned.


Mughur:
Added more hasAugmentation fixes

phyzical (approved):

PR #4105, opened 2022-09-14
Fix Jest CI Error

geggleto:
Fixes the import statement that has been causing the CI suite to fail.

Also adds jetbrains IDE exclusions to the list.


geggleto:
Screen Shot 2022-09-14 at 11 23 49 AM

PR #4122, opened 2022-09-20
CODEBASE: Remove redundant code

Snarling:
So yeah this is an overall -2500 line PR that touches 6500 existing lines. Builds off of #4054 so that's where 250 of the lines were removed. There are probably some things I'm forgetting here, but I'll go through some of the major redundancies I addressed in this PR:

### 1. Excessive prop passing
* A lot of unique, top-level game objects (Player, Terminal, Router, Engine) were being needlessly passed around as either React props or as function parameters. Instead these are now imported where they are directly needed. I also removed the context providers for Player and Router, since they were also unnecessary.
* This also inadvertently fixes one of the major exploits, accessing Player/Terminal/Router through React due to prop leaking. Fixing this exploit wasn't the goal of the PR, just a side effect.

### 2. Overuse of interfaces for classes
* The best example here are the various Person objects. There was an IPerson interface which defines the structure of a person object, then a Person class which also redefines that entire structure. Same with Player, and for the PlayerObject itself, not only was the full structure defined in the IPlayer and IPerson interfaces, but then in the PlayerObject class body, and then again in the constructor when actually assigning values to the specific instance of PlayerObject. Since PlayerObject wasn't actually a Person, it even had to redefine all the functions which are common to every Person. In some cases the signature for a function was being defined 5 different times. There were notes on the IPlayer interface that it only existed because "the player object can't be converted to typescript" so this pattern may have been a remnant from early on in the conversion of Bitburner to typescript.
* Since classes are already strictly defined types, in many cases the associated interfaces could just be removed (like IPlayer, IPerson) and I replaced references to those interfaces with references to the class. I also made Player properly inherit from Person.
* The only functions that reference the PlayerObject type are the PlayerObject functions, and only to assign a type to `this`. Any functions that are strictly for _The Player_ now just import Player. so no type definition is needed. Functions that take a player-like object that isn't necessarily _The Player_ (e.g. formulas stuff which could take in a fake-player, or stuff that works with sleeves or the player) take in a Person instead.

### 3. Explicitly defined types for an object that already has that exact same implicit type.
* I removed many of the type definitions in instances like this example, where no ambiguous types were present and all types just followed from what values were being assigned to the object.  e.g. the entire type definition can be cleanly removed from this example object and the type of `obj` does not change:
```typescript
const obj: {
  key1: number,
  key2: string,
  key3: CityName[],
} = {
  key1: 1,
  key2: "Hello there",
  key3: [CityName.Aevum, CityName.Sector12]
}
```
In some cases, the assignment portion was so bulky that it was nice having the separate type section as an overview, like with the main game constants. In these cases I did not remove the unnecessary types section.


Snarling:
@hydroflame This is also rebased and conflicts fixed now. Might want to wait to merge this one to dev until after the next time we merge dev to master.

PR #4132, opened 2022-09-23
UI: FIX #2962 Add a setting to display middle time unit in Time Elapsed String

### LINKED ISSUE #2962: Don't omit middle components in long-form times

hydroflame:
Copy of https://github.com/danielyxie/bitburner/pull/3695 because git wouldn't let me fix MC in github. All credit to borisflagell

Fix https://github.com/danielyxie/bitburner/issues/2962

### DIFF
--- a/src/GameOptions/ui/InterfacePage.tsx
+++ b/src/GameOptions/ui/InterfacePage.tsx
@@ -50,6 +50,12 @@ export const InterfacePage = (): React.ReactElement => {
           <>If this is set all references to memory will use GiB instead of GB, in accordance with IEC 60027-2.</>
         }
       />
+      <OptionSwitch
+        checked={Settings.ShowMiddleNullTimeUnit}
+        onChange={(newValue) => (Settings.ShowMiddleNullTimeUnit = newValue)}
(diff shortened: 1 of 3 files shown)


hydroflame:
Fix https://github.com/danielyxie/bitburner/issues/2962

PR #4134, opened 2022-09-23
MISC: FIX #3649 Preventing server starting security level from going above 100

### LINKED ISSUE #3649: Server initial security levels are outside valid limits
The security level of a server is clamped by the hacking mechanics to the range 1 - 100 via `server.capDifficulty`:
https://github.com/danielyxie/bitburner/blob/4993a024211714c5af9a79fd73f8d9a37ed4025f/src/Server/Server.ts#L91
However, when a server is created (e.g. after installing augmentations), a server's initial security level can start outside this range. For example, several servers spawn with 0 security on all bitnodes, such as CSEC; this results in `weaken` actually *increasing* the server security:

bb1

This case is easy to fix: just have them spawn at a security of 1 instead. The other end requires more thought. For example, on BN13 where servers spawn with 3x their normal starting security level, `hack` or `grow` can cause a massive drop in security level:

bb2

One solution would be to just cap the initial security at 100, but I think the problem really lies in the 1-100 cap itself. For example, `fulcrumassets` and other servers with a base security of 99 would normally have a minimum security of 33, but on BN13 that minimum security becomes 99: the security level only takes values from 99-100, which seems clearly undesirable.

My proposed solution would instead be to cap security levels in the range 1-(100 * bitnode.ServerStartingSecurity), i.e. the maximum security level would depend on the bitnode. This would affect any hacking mechanic that assumes the security is at most 100, e.g.:
https://github.com/danielyxie/bitburner/blob/4993a024211714c5af9a79fd73f8d9a37ed4025f/src/Hacking.ts#L50
Those formulas could instead depend on `Math.min(server.hackDifficulty, 100)` - i.e. at security levels above 100, the mechanics would act the same as at security level 100, but you now have to actually weaken the server down from whatever larger value it starts at. This seems to me to be what is intended by starting the servers off at a higher security level: to slow down the hacking mechanics and make the player have to chew through a long initial weaken before they can actually do anything with the servers.

At least, that's my proposed solution: I'm not going to start poking around with the fundamental hacking mechanics without some discussion first.

However, if we do fiddle with server's initial security levels, this might also be a good opportunity to either fully deprecate the concept of 'base security level' altogether and just work everything in terms of the minimum security level, or make the concepts of 'initial security level' and 'minimum security level' independent - at the moment they're basically just related by a factor of 3.

Shiiyu:
fixes #3649 

Testing:
Minimum starting security level on server creation seems to have been fixes since this issue was made as it properly shows as 1 on both branches. The changes appear to have fixed the new starting security of each server on creation, though that also affects the minimum security calculations, which I am not sure myself whether the new or old values are intended, and if they were I can revise the pr as so. **Will note that in the screenshots "Maximum" was a typo on my script as the value after the word is actually the minimum security and github also did not allow me to upload screenshots so I have left links.**

Current (BN13):
https://prnt.sc/4gpx7Am1zaCq
PR( BN13):
https://prnt.sc/1EYiHWcqOeyu

### DIFF
--- a/src/Server/Server.ts
+++ b/src/Server/Server.ts
@@ -76,7 +76,9 @@ export class Server extends BaseServer {
 
     //Hack Difficulty is synonymous with server security. Base Difficulty = Starting difficulty
     this.hackDifficulty =
-      params.hackDifficulty != null ? Math.min(params.hackDifficulty * BitNodeMultipliers.ServerStartingSecurity, 100) : 1;
+      params.hackDifficulty != null
+        ? Math.min(params.hackDifficulty * BitNodeMultipliers.ServerStartingSecurity, 100)
+        : 1;
     this.baseDifficulty = this.hackDifficulty;
     this.minDifficulty = Math.max(1, Math.round(this.hackDifficulty / 3));
     this.serverGrowth = params.serverGrowth != null ? params.serverGrowth : 1; //Integer from 0 to 100. Affects money increase from grow()
--- a/src/Server/Server.ts
+++ b/src/Server/Server.ts
@@ -75,12 +75,11 @@ export class Server extends BaseServer {
     this.moneyMax = 25 * this.moneyAvailable * BitNodeMultipliers.ServerMaxMoney;
 
     //Hack Difficulty is synonymous with server security. Base Difficulty = Starting difficulty
-    this.hackDifficulty =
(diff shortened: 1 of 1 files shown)


Shiiyu:
Screenshot of minimum security fix:
https://prnt.sc/5g9lMA1j-gdx

PR #4135, opened 2022-09-23
CORPORATION: FIX #3880, #3876, #3322 and #3138 Bunch of corporation fixes

### LINKED ISSUE #3138: CorportionInfo's issuedShares property is explained incorrectly
Version 1.5.0 (b265d7ba)

**issuedShares** is indicated to represent the 'Amount of shares issued'
**totalShares** also seems to represent the same data: 'Total number of shares issues by this corporation'

In reality, it seems that **issuedShares** provides the number of shares available for buy back. I think it should be given a name that more accurately reflects what it is and that also doesn't seem to conflict with the **totalShares** property.

### LINKED ISSUE #3322: Corp share price tanks too dramatically after 1e6 shares sold.
See the gif below from selling an increasing number of shares in a corporation.

bitburner_2022-04-01_11-44-54

Notice that the "total sale price" barely increases after 1e6.

This is due to an over-zealous mechanic in `calculateShareSale` that repeatedly divides the share price by a large value (possibly also incorrectly calculated) for every 1e6 shares sold.

I believe the current constant `SHARESPERPRICEUPDATE` should actually be a relative amount (relative to the total number of shares) rather than a fixed amount, and the calculation for computing the new share price should be based tested to scale more smoothly.

There are a few other oddities about this routine that should maybe be addressed while fixing it:
- `this.shareSalesUntilPriceUpdate` appears to have been added to keep track of the "remainder" of shares sold in between increments of `SHARESPERPRICEUPDATE` - but the value is never updated, so one could in theory cheese the system by repeatedly selling `SHARESPERPRICEUPDATE - 1` shares and never affect the stock price.
- We may also wish to add the Corp Softcap modifier as a factor here so that the overall sale value is nerfed idendically to dividends.

### LINKED ISSUE #3876: Corporation export/import not working
Exporting plants from agriculture seems to work. At least they are listed as "exported".

However, the receiving end of the shipment does not list anything as "imported".
The resources seem to disappear.

- [X] Steps to reproduce
Add an export for plants to some agriculture warehouse.
Check that the expected import for plants does not show up at (in my case) the chemical warehouse.

- [X] Version of the game
v1.7.0 (b60f3da8)

### LINKED ISSUE #3880: Corporations - "Drones - Transport" not working without "Smart Storage"
Warehouse storage does not seem to receive the 1.5x from Drones - Transport until Smart Storage has at least 1 level.

- Steps to reproduce
1. Start Corporation
2. Expand Industry
3. Expand Cities/Warehouses (Had all cities, 300 storage Sector-12)
4. Research Drones, Drones - Transport

Tried restarting Bitburner with no change.

- Version: 1.7.0 (00adc2ff)

Mughur:
Various small fixes for corporation.
(TerminalRoot.tsx somehow got caught in prettify ¯\\\_(ツ)\_/¯ )

- Fixes #3880
- Fixes #3876 
- Fixes #3322
- Fixes #3138 
- Fixes some other bugs that were reported in discord

### DIFF
--- a/src/Corporation/Actions.ts
+++ b/src/Corporation/Actions.ts
@@ -1,6 +1,8 @@
+import { Player } from "../Player";
 import { IPlayer } from "src/PersonObjects/IPlayer";
 import { MaterialSizes } from "./MaterialSizes";
 import { ICorporation } from "./ICorporation";
+import { Corporation } from "./Corporation";
 import { IIndustry } from "./IIndustry";
 import { IndustryStartingCosts, IndustryResearchTrees } from "./IndustryData";
 import { Industry } from "./Industry";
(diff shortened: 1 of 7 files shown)


Mughur:
Fixes  #3880, #3876, #3322, #3138 and some other bugs that were reported in discord

PR #4141, opened 2022-09-23
API: Remove incorrectly placed 's' in ns.tFormat()

LJNeon:
As the title states, this PR corrects results like "1 minutes" to "1 minute".


Snarling:
This looks good to me 👍

PR #4153, opened 2022-09-24
API: Add amount parameter to ns.clearLog()

LJNeon:
### Test Code:
```js
/** @param {NS} ns */
	ns.print("A silent beginning...");
	ns.print("As old as time itself...");
	ns.clearLog(Number.MAX_SAFE_INTEGER);
	ns.print("You should see this :)");
	ns.print("But not this one :(");
	ns.print("Magic doesn't exist y'know.");
	ns.clearLog(2);
	ns.print("And this! :o");
	ns.clearLog(-1);
}
```

### Before:
image

### After:
image

### DIFF
--- a/dist/bitburner.d.ts
+++ b/dist/bitburner.d.ts
@@ -3137,8 +3137,10 @@ export declare interface NS {
      * Clears the script’s logs.
      * @remarks
      * RAM cost: 0 GB
+     *
+     * @param amount - amount of lines to clear
      */
-    clearLog(): void;
+    clearLog(amount?: number): void;
 
     /**
      * Disables logging for the given function.
--- a/src/NetscriptFunctions.ts
+++ b/src/NetscriptFunctions.ts
@@ -470,8 +470,16 @@ const base: InternalAPI<NS> = {
       }
       Terminal.print(`${str}`);
     },
-  clearLog: (ctx: NetscriptContext) => (): void => {
-    ctx.workerScript.scriptRef.clearLog();
+  clearLog: (ctx: NetscriptContext) => (_amount?: unknown): void => {
+    if (_amount == null) {
+      ctx.workerScript.scriptRef.clearLog();
+    } else {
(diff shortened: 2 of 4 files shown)


G4mingJon4s:
IMO:
`==`  should be `===` in multiple lines to stop `ns.clearLog("null")`. ([here](https://github.com/danielyxie/bitburner/pull/4153/commits/f2e3822f42b3d6ff09100bae19ca8557b8d93d42#diff-101b74ef7e6805b98066f084ea20134631789cef5a1ce6dfe83d84e59a8a5033R98) and [here](https://github.com/danielyxie/bitburner/pull/4153/commits/f2e3822f42b3d6ff09100bae19ca8557b8d93d42#diff-851fc049fdd04e0ee245ab28fd29c8b014069e5bbddeada2c012c3c251e9a0e4R474))
[this](https://github.com/danielyxie/bitburner/pull/4153/commits/f2e3822f42b3d6ff09100bae19ca8557b8d93d42#diff-101b74ef7e6805b98066f084ea20134631789cef5a1ce6dfe83d84e59a8a5033R98) could also check for the amount being equal to the length, so it saves looping if it could just clear all.
The loop could be changed to just slicing the reversed array.
`this.log = this.log.reverse().slice(0, amount).reverse();`

Nice change tbh, will be useful...

LJNeon:
1. It doesn't seem like `"null"` will be a problem?
image
2. Good point, thanks for the optimization suggestion.
3. [`.reverse()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reverse) works in-place, which is why you have to call it again at the end. I much prefer the simple and quite functional `for` loop.

LJNeon:
```js
const oldLog = ns.getScriptLogs();
ns.clearLog();
for (const line of oldLog) {
  if (line.includes('DELETE ME') continue;
  ns.print(line);
}
``` As it's actually possible to do this already, and after feedback on discord, I'm closing this.

PR #4155, opened 2022-09-25
CORP: fixed wrongly named division type

Mughur:
"Computer" division type was erroneously named "Hardware" in the constants.
Division types list is only called by the getDivisionTypes ns function, so the bug didn't cause issues


phyzical:
it is called software though isnt it? 
image
image

phyzical:
oh maybe it should be Computer is missing?

Mughur:
> oh maybe it should be Computer is missing?

computer <-> hardware was the issue

phyzical:
Ah updated the wrong constant :D

Mughur:
oh ffs

phyzical (approved):

PR #4158, opened 2022-09-26
NETSCRIPT: atExit no longer fails to run ns functions when an async function was in progress

### LINKED ISSUE #4143: AtExit clashes with ns-await
`atExit` is a delegate method that gets called when the script terminates. However, this method can't do anything ***useful*** (such as saving its state data to a file!) if the script being terminated is currently `await`ing for any reason.

For example, this script can only ever terminate with a runtime error when killed with `kill` and will fail to print anything to the console:

		ns.atExit(() => {
			ns.tprint('Critical Error'); //this will never print
		});
		while(true) {
			await ns.sleep(200);
		}
	}

Yet there is no way to wait for the `await` to complete before running the atExit portion of the code as (1) the script is already awaiting which would throw an error and (2) there is no control over *when* a script is terminated as it is (usually) an external process.

Suggestions:

1. Ignore any existing await errors when calling atExit (they're going to be interrupted anyway)
2. Interrupt await calls *before* calling atExit
3. Add a `isAwaiting()` method/property (although if this is true and the script still wants to do something on exit, it would have to be awaitable itself, and (1) comes back into play again, although it could be more of a localized "if await was called on isAwaiting, don't error" type of situation).
4. If the await that is active is `sleep`, let it complete first, then exit

- [x] Save file [bitburnerSave_1663868241_BN1x1.zip](https://github.com/danielyxie/bitburner/files/9636331/bitburnerSave_1663868241_BN1x1.zip)
- [x] Minimal scripts to reproduce the issue
  - Run `kill invest.js` upon loading the save
- [x] Version of the game, e.g. Bitburner v2.0.2

Snarling:
FIX #4143

### DIFF
--- a/src/Netscript/killWorkerScript.ts
+++ b/src/Netscript/killWorkerScript.ts
@@ -53,7 +53,9 @@ function killWorkerScriptByPid(pid: number): boolean {
 }
 
 function stopAndCleanUpWorkerScript(ws: WorkerScript): void {
-  killNetscriptDelay(ws);
+  if (ws.delay) clearTimeout(ws.delay);
+  ws.delayReject?.(new ScriptDeath(ws));
+  ws.env.runningFn = "";
   if (typeof ws.atExit === "function") {
     try {
       ws.env.stopFlag = false;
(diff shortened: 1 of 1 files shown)


Snarling:
Changes were merged in with #4122

PR #4165, opened 2022-09-27
SLEEVE: Can no longer use API to assign sleeves to bladeburner tasks before player is member

### LINKED ISSUE #4089: Assigning a sleeve to Bladeburner tasks before joining Bladeburner triggers recovery mode
Version: 2.1.0 (https://github.com/danielyxie/bitburner/commit/87054f610410911c6cd15b2a921d55e47d32a83e) 

I specifically got an error about Diplomacy. My main script started assigning sleeves to Bladeburner work items but I hadn't actually joined yet. It continued as normal until I clicked the Sleeves UI button in the menu bar and then crash.

Snarling:
FIX #4089 

Additionally I made changes in Sleeve.ts to remove player as a parameter from main sleeve functions. E.g. Sleeve.stopWork(Player) is now Sleeve.stopWork().

Other changes are just follow-on changes for these parameter changes in Sleeve.ts, and a lint fix from a different merged PR that wasn't lint checked properly (AugmentationCreator.tsx)

### DIFF
--- a/src/PersonObjects/Sleeve/Sleeve.ts
+++ b/src/PersonObjects/Sleeve/Sleeve.ts
@@ -424,6 +424,7 @@ export class Sleeve extends Person {
    * Begin a bladeburner task
    */
   bladeburner(action: string, contract: string): boolean {
+    if (!Player.bladeburner) return false;
     switch (action) {
       case "Field analysis":
         this.startWork(new SleeveBladeburnerWork({ type: "General", name: "Field Analysis" }));
@@ -448,7 +449,6 @@ export class Sleeve extends Person {
         this.startWork(new SleeveBladeburnerWork({ type: "Contracts", name: contract }));
         return true;
     }
-
     return false;
   }
 


src/PersonObjects/Sleeve/Sleeve.ts
```diff
@@ -427,41 +422,41 @@ export class Sleeve extends Person {
   /**
    * Begin a bladeburner task
    */
-  bladeburner(p: IPlayer, action: string, contract: string): boolean {
+  bladeburner(action: string, contract: string): boolean {
+    if (!Player.bladeburner) return false;
```
Snarling, line 426: This is the main fix. The rest is parameter changes.

phyzical:
i thought the reason the player object was thrown around is that it actually was the sleeve in question? (may be misremembering)

Snarling:
> i thought the reason the player object was thrown around is that it actually was the sleeve in question? (may be misremembering)

Any time it's the IPlayer interface, it's talking about the PlayerObject, since that's the only thing that satisfies the IPlayer interface (there are also a few instances in Formulas type stuff where it might refer to a fake netscript player as well, with the netscript call just pretending that the fakePlayer is an IPlayer).

Inside the Sleeve class (where most of the changes are on this particular PR), the sleeve in question is just `this`. In functions outside the sleeve class, it's usually called something like `sleeve` (when it can only be a sleeve) or `person`/`actor` (when it can be a player or sleeve). e.g.
https://github.com/danielyxie/bitburner/blob/69eda4340e4fc63716fd1f8c7b039a32a7fcbf2f/src/Bladeburner/Bladeburner.tsx#L1259

Values that can be players or sleeves always use the IPerson interface because that's the only type overlap for Player and Sleeves. Things that are always sleeves use the Sleeve class as a type directly.

src/PersonObjects/Sleeve/Sleeve.ts
```diff
@@ -196,24 +194,24 @@ export class Sleeve extends Person {
     cyclesUsed = Math.min(cyclesUsed, 15);
     this.shock = Math.min(100, this.shock + 0.0001 * cyclesUsed);
     if (!this.currentWork) return;
-    this.currentWork.process(p, this, cyclesUsed);
+    this.currentWork.process(Player, this, cyclesUsed);
```
Snarling, line 197: Player can also be removed as a parameter from sleeve work functions, but those are defined in different files. I kept the parameter changes in this PR basically to just Sleeve.ts to increase chances this gets merged.

phyzical (approved):

PR #4172, opened 2022-09-27
CCT: Generate IPs contract type accepts properly formatted js array

### LINKED ISSUE #4171: "Generate IP Addresses" contract did not accept solution via API nor quotation marks in manual solution
This is in Bitburner 2.1.0 (a3c92dd0), and I _believe_ is new in v2, since I had not encountered this issue previously and my script hasn't changed in a while.

I was presented with a "Generate IP Addresses"-type coding contract which resulted in 1 solution. The NS2 script I normally use to solve the contract resulted in a "failure" response. The simplified code is basically...

```
let solutionArray = solveThisContract(...); // The return value was a 1-element string array
let attemptResponse = ns.codingcontract.attempt(solutionArray, contractFilename, contractHostname, {returnReward: true});
// attemptResponse was an empty string
```

I dumped the output to terminal and ran the contract manually. I typed `["84.74.100.1"]` into the solution text box which again failed to close the contract. I then typed `[84.74.100.1]` into the solution text box, and the contract closed successfully (meaning the script's original solution array was correct as far as the contract was concerned).

The no-quotation-marks format of typing in the solution matches the examples in the contract, so I kinda-sorta get that. But I think that it's a bug that the API-submitted string array failed.

### LINKED ISSUE #4197: Type check broke "Generate IP Addresses" solvers
Version: 2.1.0 (8f4636cb)

Some recent update broke all (?) solvers for "Generate IP Addresses" contracts. The new version introduced this check:

https://github.com/danielyxie/bitburner/blob/fa099c22ab9bbc1412291ae0fb07c80ddf63e0f0/src/data/codingcontracttypes.ts#L511

In the past the game was happy to accept a JSArray as valid solution.

There was no real reason for this breaking change and the original behaviour should be restored.

Snarling:
fix #4171
fix #4197 

Probably should do a larger rewrite that doesn't compare all contract solutions internally as strings, but this . .

Prior to this, generate IPs was only accepting IPs that were not contained in quotation marks, e.g. something like 
`[192.168.0.1, 192.16.80.1, 19.216.80.1]`. 
This will still be accepted (it probably shouldn't be) but the more correct 
`["192.168.0.1", "192.16.80.1", "19.216.80.1"]`
which is how an array of strings is denoted in js, should also be accepted after this PR.

### DIFF
--- a/src/data/codingcontracttypes.ts
+++ b/src/data/codingcontracttypes.ts
@@ -490,8 +490,8 @@ export const codingContractTypesMetadata: ICodingContractTypeMetadata[] = [
         "Note that an octet cannot begin with a '0' unless the number",
         "itself is actually 0. For example, '192.168.010.1' is not a valid IP.\n\n",
         "Examples:\n\n",
-        "25525511135 -> [\"255.255.11.135\", \"255.255.111.35\"]\n",
-        "1938718066 -> [\"193.87.180.66\"]",
+        '25525511135 -> ["255.255.11.135", "255.255.111.35"]\n',
+        '1938718066 -> ["193.87.180.66"]',
       ].join(" ");
     },
     difficulty: 3,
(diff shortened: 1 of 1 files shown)


quacksouls:
Related issue #4197.

PR #4201, opened 2022-10-04
MISC: Fix bug in person.hasAugmentation

stalefishies:
#4122 introduced a bug where `person.hasAugmentation` did not correctly return if a player had a queued augmentation.

Screenshot of the bug in question: I should be able to purchase CSP Gen III since I own CSP Gen II (it is queued, but not installed)
image

For reference, the commit showing the regression: https://github.com/danielyxie/bitburner/commit/b364cdf7343c365f4edd3e39e2cca8fe41011cb8?diff=split#diff-1a340806bb6720dd5be22ba71979a6b26940f77dcc4b72d029360b6d2e85b3fdL10

### DIFF
--- a/src/PersonObjects/PersonMethods.ts
+++ b/src/PersonObjects/PersonMethods.ts
@@ -199,5 +199,11 @@ export function updateSkillLevels(this: Person): void {
 }
 
 export function hasAugmentation(this: Person, augName: string, ignoreQueued = false) {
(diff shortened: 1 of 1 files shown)


Snarling (approved):
Yeah I don't know what I was thinking with that one. For some reason I thought that `augmentations` contained both owned and queued augmentations.

PR #4224, opened 2022-10-09
DOC: `ns.scan()`: typo fixes and add 2 examples

quacksouls:
This PR changes the documentation of the function `ns.scan()` as follows:

1. Commit 67f5d601162f3600e7cde6abbae2b2106e7b3ea9.  Some typographical fixes in the parameter and return value of the function.
2. Commit b03288a7003404814b7d9e21ce84defdc375bbf4.  Two examples on using `ns.scan()`.  The first example uses `ns.scan()` without an argument.  The second example uses `ns.scan()` by passing the hostname of a target server.  These examples can be written more concisely by using the `forEach()` method of array, but for the purposes of illustration we use a `for` loop to make it easier for beginners to read.
3. ~Commit e3c466e2ecf8c8bedb3e75fd6a03d9355151c0a2.  Add an example on how to use the function to navigate all servers in the game.~

### DIFF
--- a/src/ScriptEditor/NetscriptDefinitions.d.ts
+++ b/src/ScriptEditor/NetscriptDefinitions.d.ts
@@ -5156,8 +5156,8 @@ export interface NS {
    * node way from the specified target server. The hostnames in the returned
    * array are strings.
    *
-   * @param host - Optional, Hostname of the server to scan, default to current server.
-   * @returns Returns an string of hostnames.
+   * @param host - Optional. Hostname of the server to scan, default to current server.
+   * @returns Returns an array of hostnames.
    */
   scan(host?: string): string[];
 
--- a/src/ScriptEditor/NetscriptDefinitions.d.ts
+++ b/src/ScriptEditor/NetscriptDefinitions.d.ts
@@ -5156,6 +5156,41 @@ export interface NS {
    * node way from the specified target server. The hostnames in the returned
    * array are strings.
    *
+   * @example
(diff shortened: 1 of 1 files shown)


Snarling (commented):
I don't think including solutions to game-problems (like "how do I scan the whole network") should be included in function documentation. The player should need to combine their knowledge of loops / arrays / whatever, as well as basic documentation on ingame functions like scan, in order to solve these problems themselves.

quacksouls:
> I don't think including solutions to game-problems (like "how do I scan the whole network") should be included in function documentation. The player should need to combine their knowledge of loops / arrays / whatever, as well as basic documentation on ingame functions like scan, in order to solve these problems themselves.

Can't argue against this logic.  Seems like I was giving away a solution to a homework problem.  The updated examples show the use of `ns.scan()` with and without an argument.  See commit 3b631fe33e23dc904b510110e587984a2b853866.

hydroflame:
I think using forEach is an unecessary concept to introduce via examples. Specially for newbies using ns1.

Can you use ` for(let i = 0` loop instead?

quacksouls:
> I think using forEach is an unecessary concept to introduce via examples. Specially for newbies using ns1.
> 
> Can you use ` for(let i = 0` loop instead?

Nice catch!  The examples have been updated to use the `for` loop.  Please see commit b03288a7003404814b7d9e21ce84defdc375bbf4.

PR #4227, opened 2022-10-11
API: Added ns.getMoneySources() to get the money since install/start

xxxsinx:
Test code:

/** @param {NS} ns */
	const money= ns.getMoneySources();
	ns.tprint(money);
}


Snarling (approved):
Looks good to me. The number of times the hardcoded structure is duplicated is gross, but MoneySourceTracker being a class forces that to some extent.

xxxsinx:
Changed the comment for the function to remove the word "node", unnecessary spoiler, even though the returned structure is a spoiler in itself :D

hydroflame:
Adde manually because merge conflitcs

PR #4237, opened 2022-10-12
MISC: message formatting

Jakob002:
When the player successfully solved a coding contract and the reward was reputation for multiple factions, the method PlayerObjectGeneralMethods.ts#gainCodingContractReward() returned a string like
`Gained 1250 reputation for each of the following factions: BitRunners,The Black Hand,NiteSec`. 
Note the missing white space between faction names. I changed that behaviour so it now instead says:
`Gained 1250 reputation for each of the following factions: BitRunners, The Black Hand, NiteSec`


quacksouls:
Looks good.  Works as advertised.  Here's what it looks like before the patch:
before

Here's what it looks like after the patch:
after
Also, the patch does not include extraneous commas even when you have joined only 1 faction.

PR #4255, opened 2022-10-16
DOC: `gang.canRecruitMember()`: typo fix and elaboration

quacksouls:
This PR changes the documentation of `ns.gang.canRecruitMember()` as follows:

1. Commit ac0e7c959558e3daf715a2788cdc084460ab5b45.  Some typographical fixes.
2. Commit 6f1822bf4d94d96a56e4b9f0f146d0e60463aacb.  Elaborate (vaguely) on various conditions under which the function would return true or false.


Snarling (commented):
I feel like this is documentation on how the gang mechanic works, not documentation on how gang.canRecruitMember works. IMO Netscript documentation should not be the resource that explains how other game mechanics work.

PR #4265, opened 2022-10-19
REPO MOVED TO github.com/bitburner-official/bitburner-src

hydroflame:
Do not merge this PR

PR #4274, opened 2022-11-05
Update bitburner.basichgwoptions.md

profess:
Updating to include examples on how to use the interface in a call using NS2.


Snarling:
1. See #4265 the active development repo is at github.com/bitburner-official/bitburner-src
2. Markdown files do not get modified directly. They are autogenerated from NetscriptDefinitions.d.ts (this is stated in the contributing readme)

PR #4298, opened 2023-01-17
Update Master to v2.2.1 stable

Snarling:
Commit 6e29317: Changed README, ISSUE_TEMPLATE, and PULL_REQUEST_TEMPLATE to make it clear this is the wrong repo. Moved entire folder structure other than the game build into archive folder.

Commit bec33d0: Built stable version 2.2.1 from https://github.com/bitburner-official/bitburner-src/tree/stable using github action https://github.com/Snarling/bitburner/blob/dev/.github/workflows/update.yml

Commit fc6029b: Moved .github template files back to a root .github folder (they were in archive/.github)


Snarling:
@hydroflame 

We'll also need to merge this into dev since that's the default branch (so that's where github pulls the template files and where it shows the readme when someone just visits https://github.com/danielyxie/bitburner. If you want I can just resubmit this as a merge to dev, and then you can merge it to master after. In the future it should only be master that needs updating since that's where pages builds from, so I was planning on just submitting PRs straight to master.

